PR #9, opened 2016-08-19
Git server role

sabre1041:
#### What does this PR do?
Adds a role to create a httpd backed git server

Create a git host group containing target machines and a playbook with the following contents

```
- name: Creates Git Server
  hosts: git
  roles:
  - role: git-server
```

None

/cc @detiber @etsauer @oybed

### DIFF
--- a/roles/git-server/tasks/main.yaml
+++ b/roles/git-server/tasks/main.yaml
@@ -1,0 +1,45 @@
+---
+- name: Install Packages
+  action: "{{ ansible_pkg_mgr }} name={{ item }} state=present"
+  with_items:
+    - git
+    - httpd
+    - firewalld
+    - libsemanage-python
+    
+- name: Create Git User
+  user: name="{{ git_user }}"
+
+- name: Create Git User Authorized Keys
+  authorized_key:
+    user: "{{ git_user}}"
+    key: "{{ item }}"
+  with_items:
+    - "{{ git_user_authorized_keys | default('') }}"
+  
+- name: Configure Git HTTP Configuration
+  template:
+    src: "{{ role_path }}/templates/git.conf.j2"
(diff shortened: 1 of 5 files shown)


detiber:
Only comment I have is that the README should probably mention that this role should be deployed on a system that is not part of the OpenShift cluster, since it requires firewalld.

sabre1041:
@detiber added content to README as requested

oybed:
@sabre1041 LGTM

sabre1041:
@oybed @detiber After a few more validations with ansible 2.0, added additional filter to prevent errors with undefined authorized keys

roles/git-server/tasks/main.yaml
```diff
@@ -0,0 +1,45 @@
+---
+- name: Install Packages
+  action: "{{ ansible_pkg_mgr }} name={{ item }} state=present"
```
detiber: Unless 1.9 support is needed here, I would suggest using the `package` module here.

If 1.9 support is needed, should probably define ansible_pkg_mgr in vars/main.yaml for the role.

Otherwise, lgtm.

oybed:
@detiber @sabre1041 good point about which versions are supported. Maybe we should just note that in the README for which version of Ansible the implementation has been tested with and/or intended for use with. Hence leaving it up to the author for what they'd like to support. Thoughts?

sabre1041:
@detiber @oybed I added a compatibility section describing the ansible version the role was tested against

detiber:
@sabre1041 lgtm unless @oybed has any outstanding items, I'm ready to merge.

oybed:
@sabre1041 thank you for adding the compatibility section - LGTM. I'll go ahead and merge.

PR #10, opened 2016-08-22
Role for seeding a git server with OpenShift example repositories

sabre1041:
#### What does this PR do?
Role for seeding git server introduced by #9 with example repositories

Create a git host group containing target machines and a playbook with the following contents

```
- name: Creates Git Server
  hosts: git
  roles:
  - role: seed-git-server
```

None

/cc @detiber @etsauer @oybed

### DIFF
--- a/roles/seed-git-server/README.md
+++ b/roles/seed-git-server/README.md
@@ -1,0 +1,46 @@
+Seed Git Server
+====================
+
+Seeds a git server with example repositories 
+
+Requirements
+------------
+
+This role has dependencies that are not suitable for systems that are part of an OpenShift Cluster as it requires *firewalld* to be installed
+
+
+Role Variables
+--------------
+
+From this role:
+
+| Name                     | Default value |  Description                                 |
+|--------------------------|---------------|-------------------------------------------------------------------------------------|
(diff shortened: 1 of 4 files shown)


roles/seed-git-server/README.md
```diff
@@ -0,0 +1,41 @@
+Seed Git Server
+====================
+
+Seeds a git server with example repositories 
+
+Requirements
+------------
+
+This role has dependencies that are not suitable for systems that are part of an OpenShift Cluster as it requires *firewalld* to be installed
```
oybed: Maybe update this "requirement" to remove the mention of firewalld as that's for the git-server role rather than this role. However, maybe just list a requirement that a git server needs to be operational (in addition to the `Dependencies` below).

oybed:
In addition to comment inline, just a nitpick - should `ose` be changed to `openshift` as I don't think anything in these roles really depends on OSE (or OCP rather) and hence it could be applicable to other OpenShift types (or even non-OpenShift deployments)

detiber:
Agree on the comments made by @oybed, lgtm otherwise.

sabre1041:
@oybed renamed ose -> openshift

oybed:
@sabre1041 LGTM - will merge

PR #12, opened 2016-08-26
incorrect path discovered while ci testing

cooktheryan:


cooktheryan:
@detiber not sure if you have time to give this a plus 1 today. Collier is currently on travel so rather low on sign offs

detiber:
:+1:

PR #13, opened 2016-08-31
inclusion of ansible.cfg to facilitate installation

cooktheryan:
The ansible.cfg allows for certain parameters to be set during playbook run time.

### DIFF
--- a/reference-architecture/aws-ansible/ansible.cfg
+++ b/reference-architecture/aws-ansible/ansible.cfg
@@ -7,9 +7,6 @@ host_key_checking = False
 hostfile = inventory/aws/hosts/ec2.py
 remote_user = ec2-user
 gathering = smart
-fact_caching = jsonfile
-fact_caching_connection = .ansible/cached_facts
-fact_caching_timeout = 900
 #lookup_plugins = ./playbooks/lookup_plugins
 #log_path = /tmp/ansible.log
 
--- a/reference-architecture/aws-ansible/ansible.cfg
+++ b/reference-architecture/aws-ansible/ansible.cfg
@@ -7,6 +7,8 @@ host_key_checking = False
 hostfile = inventory/aws/hosts/ec2.py
 remote_user = ec2-user
 gathering = smart
+retry_files_enabled = false
+nocows = true
 #lookup_plugins = ./playbooks/lookup_plugins
 #log_path = /tmp/ansible.log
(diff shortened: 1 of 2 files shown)


cooktheryan:
@scollier forwarded you an email talking with @detiber the other day about including ansible.cfg in the repo.

reference-architecture/aws-ansible/ansible.cfg
```diff
@@ -0,0 +1,22 @@
...
+[defaults]
+#callback_plugins = ../openshift-ansible/ansible-profile/callback_plugins
+forks = 50
+host_key_checking = False
+hostfile = inventory/aws/hosts/ec2.py
+remote_user = ec2-user
+gathering = smart
+fact_caching = jsonfile
+fact_caching_connection = .ansible/cached_facts
+fact_caching_timeout = 900
```
detiber: I would suggest against using fact caching, unless it will only be used through the wrapper script and the wrapper clears the fact cache between runs.

Only reason I say that is because the fact caching will cause unintuitive behavior for users that aren't familiar with the feature in advance.

cooktheryan: I'm good with removing that.

Ryan

cooktheryan:
@detiber pushed update of file with fact lines removed

reference-architecture/aws-ansible/ansible.cfg
```diff
@@ -0,0 +1,19 @@
+# config file for ansible -- http://ansible.com/
+# ==============================================
+[defaults]
+#callback_plugins = ../openshift-ansible/ansible-profile/callback_plugins
+forks = 50
+host_key_checking = False
+hostfile = inventory/aws/hosts/ec2.py
+remote_user = ec2-user
+gathering = smart
```
detiber: This line should be removed as well.

You'll probably want to set the following:
```
retry_files_enabled = false
nocows = true
```

Otherwise, lgtm.
@abutcher any thoughts?

cooktheryan:
Definitely cool with removing the retry. Feel bad for the admin who will be bummed out that there is no cowsay :)

detiber:
Heh, those that care can override. Better than a CTO wondering why our refarch script is displaying ASCII cows :)

abutcher:
lgtm

PR #14, opened 2016-09-01
Ansible 2 and moving to cloudformations

cooktheryan:
This will be the exact same setup as using pure ansible but this time with cloudformations. Using previous cloud formation templates we were able to trim down ports and role privileges.  This merge will be what we use with CI.


cooktheryan:
I would probably suggest squashing this when the time comes...its a bit long

cooktheryan:
@detiber if you get time this week to check out this PR it would be much appreciated. Collier is presenting this week.

This PR is the work to migrate to cloudformations with the exact infrastructure diagram we used before but rather than using pure ansible I migrated to CF. Closed up a bunch of the extra ports in SGs after the review you sent. No rush for review today or tomorrow.

reference-architecture/aws-ansible/playbooks/openshift-install.yaml
```diff
@@ -7,6 +7,14 @@
   roles:
       # Group systems
       - instance-groups
+- hosts: bastion
```
detiber: Not directly related to this PR, but I would add some blank lines between the separate plays. It gets hard on the eyes to try to follow based on the indentation level alone.

reference-architecture/aws-ansible/playbooks/openshift-install.yaml
```diff
@@ -34,6 +41,14 @@
       - vars/main.yaml
   roles:
       - openshift-volume-quota
+- hosts: masters
+  gather_facts: no
```
detiber: Is gather_facts causing a big issue? If so, it might be better to leverage fact caching, at least with the python wrapper (where you can wipe the cache between invocations to ensure no bad data being pulled in between runs).

reference-architecture/aws-ansible/playbooks/openshift-setup.yaml
```diff
@@ -0,0 +1,34 @@
...
+    osm_cluster_network_cidr: 172.16.0.0/16
+    openshift_registry_selector: "role=infra"
+    openshift_router_selector: "role=infra"
+    openshift_master_cluster_method: native
+    openshift_cloudprovider_kind: aws
+    openshift_master_cluster_hostname: "internal-openshift-master.{{ public_hosted_zone }}"
+    openshift_master_cluster_public_hostname: "openshift-master.{{ public_hosted_zone }}"
+    osm_default_subdomain: "{{ wildcard_zone }}"
+    deployment_type: openshift-enterprise
+    openshift_master_identity_providers: [ {'name': 'google', 'kind': 'GoogleIdentityProvider', 'login': 'true', 'challenge': 'false', 'mapping_method': 'claim', 'client_id': '246358064255-5ic2e4b1b9ipfa7hddfkhuf8s6eq2rfj.apps.googleusercontent.com', 'client_secret': 'Za3PWZg7gQxM26HBljgBMBBF', 'hosted_domain': 'redhat.com'} ]
```
detiber: probably want to pass some of these values in as variables rather than hard-coded in the template.

cooktheryan:
@detiber opened containerized story https://trello.com/c/mgbKejSC/377-install-containerized-ose-on-rhel and cockpit story https://trello.com/c/FGehBUy9/378-install-cockpit-during-ose-deployment as they will require tool changes and update of documentation.

Also, added spaces to openshift-install.yaml I agree needed line spaces

reference-architecture/aws-ansible/playbooks/openshift-install.yaml
cooktheryan: No issues...was just trying to make sure fact gathering was only done on the 1 play that needs it. Any suggestions appreciated

reference-architecture/aws-ansible/playbooks/openshift-setup.yaml
```diff
@@ -0,0 +1,34 @@
...
+    openshift_node_debug_level: "{{ node_debug_level | default(debug_level, true) }}"
+    osm_controller_args:
+      cloud-provider:
+      - "aws"
+    osm_api_server_args:
+      cloud-provider:
+      - "aws"
+    openshift_node_kubelet_args:
+      max-pods:
+      - "100"
```
detiber: I'm not sure you want to override the max-pods value. We did for the demo environments, but we also knew what the workload running was going to be.

detiber: Also, this is going to limit 3.3 installs to lower density than the default config.

reference-architecture/aws-ansible/playbooks/roles/cloudformation-infra/files/brownfield-byo-bastion.json
```diff
@@ -0,0 +1,1029 @@
...
+            "Ebs": {
+              "DeleteOnTermination": "true",
+              "VolumeSize": {"Ref": "MasterDockerVolSize"},
+              "VolumeType": {"Ref": "MasterDockerVolType"}
+            }
+          },
+          {
+            "DeviceName": "/dev/xvdc",
+            "Ebs": {
+              "DeleteOnTermination": "true",
```
detiber: Since this volume houses the etcd data, it may make sense to set DeleteOnTermination to false to prevent the data volume from disappearing accidentally..  Cleanup would require a second step to remove the volumes, though.

reference-architecture/aws-ansible/playbooks/roles/cloudformation-infra/files/brownfield-byo-bastion.json
```diff
@@ -0,0 +1,1029 @@
...
+                  "Resource": "*"
+                }
+              ]
+            }
+          },
+          {
+            "PolicyName": "node-detach",
+            "PolicyDocument": {
+              "Version" : "2012-10-17",
+              "Statement": [
```
detiber: I think all three of these statements can be combined under a single PolicyDocument. I think I copy and pasted from the upstream kube example and never consolidated it in demo-env-ansible.

reference-architecture/aws-ansible/playbooks/roles/cloudformation-infra/files/brownfield-byo-bastion.json
```diff
@@ -0,0 +1,1029 @@
...
+          {
+            "Name": { "Ref": "AppWildcardDomain" },
+            "Type": "A",
+            "AliasTarget": {
+                "HostedZoneId": { "Fn::GetAtt" : ["InfraElb", "CanonicalHostedZoneNameID"] },
+                "DNSName": { "Fn::GetAtt" : ["InfraElb","CanonicalHostedZoneName"] }
+            }
+          },
+          {
+            "Name": {"Fn::Join": [".", ["ose-master01",{"Ref": "Route53HostedZone"}]]},
```
detiber: I would try to avoid managing private dns entries. Since the native cloud integration already requires the host have a resolvable private dns name from the VPC config, I would just use those hostnames for internal cluster use.

reference-architecture/aws-ansible/playbooks/roles/cloudformation-infra/files/brownfield-byo-bastion.json
```diff
@@ -0,0 +1,1029 @@
...
+	      "InstanceType": {"Ref": "MasterInstanceType"},
+          "SecurityGroupIds": [{ "Fn::GetAtt" : ["NodeSG", "GroupId"] }, { "Fn::GetAtt" : ["MasterSG", "GroupId"] }, { "Fn::GetAtt" : ["EtcdSG", "GroupId"] }],
+          "SubnetId" : {"Ref": "PrivateSubnet1"},
+          "IamInstanceProfile": { "Ref": "MasterInstanceProfile" },
+          "Tags": [
+            { "Key": "Name",
+              "Value": {"Fn::Join": [".", ["ose-master01",{"Ref": "PublicHostedZone"}]]}
+            },
+            { "Key": "openshift-role",
+              "Value": "master"
```
detiber: Not related, but at some point we need to figure out a better way to tag hosts based on function...  a single tag with separate values isn't nearly flexible enough for future needs...  One example is that a master is also a node, but if allowing users to choose whether they want to have things like metrics or logging on their own dedicated hosts vs co-located on "infra" hosts, etc starts to get more difficult with a single 'role' setting.

reference-architecture/aws-ansible/playbooks/roles/cloudformation-infra/tasks/main.yaml
```diff
@@ -0,0 +1,135 @@
...
+  cloudformation:
+    stack_name: "OpenShift-Infra-Brownfield"
+    state: "present"
+    region: "{{ region }}"
+    template: "roles/cloudformation-infra/files/brownfield.json"
+    template_parameters:
+      Vpc: "{{ vpc_id }}"
+      PrivateSubnet1: "{{ private_subnet_id1 }}"
+      PrivateSubnet2: "{{ private_subnet_id2 }}"
+      PrivateSubnet3: "{{ private_subnet_id3 }}"
```
detiber: One issue I see here is that there are some regions that only have 2 AZs and us-east-1 has 4. Using ASGs and allowing for a list of subnets would help abstract this out a bit.

reference-architecture/aws-ansible/playbooks/roles/etcd-volume/tasks/main.yaml
```diff
@@ -0,0 +1,25 @@
...
+      src: '/dev/xvdc'
+      fstype: xfs
+      state: present
+
+- name: Mount fstab entry
+  mount:
+      name: /var/lib/etcd
+      src: '/dev/xvdc'
+      fstype: xfs
+      state: mounted
```
detiber: Any reason to no use cloud-init for configuring the etcd filesystem and directory?

cooktheryan: I believe the long term plan is for the block storage device to become a variable so that in the future when we move to common roles for all providers then providers like vmware that can't support user-data will be covered as well.

reference-architecture/aws-ansible/playbooks/openshift-setup.yaml
cooktheryan: resolved

reference-architecture/aws-ansible/playbooks/roles/cloudformation-infra/tasks/main.yaml
cooktheryan: In the reference arch document I called out that a region must have at least 3 or more AZs.  I haven't put forth the time yet to figure out a long term plan for zones with only 2 AZs. Most likely it would be spawn all masters in 1 AZ because if an AZ would go down we would have split brain. I Just don't think we are ready to facilitate this yet.  

In the future, we would like to move to ASGs but this work mainly just migrated the ansible deployed infrastructure as is to support the refarch document as is.  For the 3.3/3.4 release we will be diving more into that work.

reference-architecture/aws-ansible/playbooks/roles/cloudformation-infra/files/brownfield-byo-bastion.json
```diff
@@ -0,0 +1,1029 @@
...
+            "Ebs": {
+              "DeleteOnTermination": "true",
+              "VolumeSize": {"Ref": "MasterEtcdVolSize"},
+              "VolumeType": {"Ref": "MasterEtcdVolType"}
+            }
+          }
+         ]
+     }
+   },
+    "InfraNode01" : {
```
cooktheryan: Right now with ASGs we don't have a solid private DNS story in place to be able to support deploying using ASGs.  Right now we are still kind of using the pet methodology rather than cow in regards to the instances.

reference-architecture/aws-ansible/playbooks/roles/cloudformation-infra/files/brownfield-byo-bastion.json
cooktheryan: This is probably some fallout of migrating away from pure ansible deployment rather than cloudformation.  Without the private entries though I don't know how when using a bastion it would handle names without configuring ansible or the .ssh/config to support passing any ip-10* traffic through the bastion. Which wouldn't be too tricky. I can add a story for 3.3 to move towards a more cow approach to the instance.

reference-architecture/aws-ansible/playbooks/roles/cloudformation-infra/files/brownfield-byo-bastion.json
cooktheryan: This has been combined in a commit below

cooktheryan:
Tagged for 3.3. Use ASGs rather than individual instances. This will cover the ELB concerns as well as private DNS concerns. 

https://trello.com/c/3aXXXEDa/379-use-asgs-to-deploy-instances

reference-architecture/aws-ansible/playbooks/roles/cloudformation-infra/files/brownfield-byo-bastion.json
detiber: If you require dns assignment for the vpc configuration, then the existing dns naming should be sufficient...  we have a playbook now for re-configuring certificates, which could be used if instances are shutdown/restarted.

reference-architecture/aws-ansible/playbooks/roles/cloudformation-infra/files/brownfield-byo-bastion.json
detiber: Since you are on ansible 2.2, you can always use: http://docs.ansible.com/ansible/faq.html#how-do-i-configure-a-jump-host-to-access-servers-that-i-have-no-direct-access-to

reference-architecture/aws-ansible/playbooks/roles/cloudformation-infra/tasks/main.yaml
detiber: I think you could still leverage 2 AZs (one AZ housing 2 masters), that way instead of the AZ taking out the entire cluster, you could still have a read-only cluster if the AZ housing 2 masters goes down.

reference-architecture/aws-ansible/playbooks/roles/etcd-volume/tasks/main.yaml
detiber: VMWare is definitely the outlier...  Of course, GCP and Azure don't natively support cloud-init, but they do have alternatives for doing the same type of actions.

For VM-based deployments, it might be better to do something one-off just for those, or possibly not even support a separate etcd directory by default (unless they provide an image that has it pre-baked).

That said, we could require a VM image that uses LVM and has all the LVs created for these types of things out of the box.

detiber: Bigger issue will be device ordering, naming, etc. won't necessarily be identical between those environments.

reference-architecture/aws-ansible/playbooks/roles/cloudformation-infra/files/brownfield-byo-bastion.json
cooktheryan: Resolved with commit and new trigger used for ci cleanup

reference-architecture/aws-ansible/playbooks/roles/etcd-volume/tasks/main.yaml
cooktheryan: I think we might skip cloud_init until a future release of cloud-init. I've spent a chunk of the morning comparing our code with demo-ansible and even testing the code from demo-ansible. When running cloud-init I'm consistently hitting the following error when trying to partition the filesystem.  

`Sep  9 11:05:58 localhost cloud-init: Cloud-init v. 0.7.6 running 'modules:config' at Fri, 09 Sep 2016 15:05:58 +0000. Up 65.76 seconds.
Sep  9 11:05:58 localhost cloud-init: 2016-09-09 11:05:58,949 - stages.py[WARNING]: Could not find module named cc_fs_setup (searched ['cc_fs_setup', 'cloudinit.config.cc_fs_setup'])`

I'm not sure if you have any ideas on this. Right now to get cloud-init to do the needful its a series of runcmd's which seems clunky.

detiber: Hmm, odd it looks like the 'disk_setup' module handles it...  I'm not sure why/how it was working previously :)

cooktheryan: resolved

PR #15, opened 2016-09-14
Update of README as more providers are coming

cooktheryan:


cooktheryan:
@detiber the external load balancer has the lime green lines which only go to the masters

detiber:
@cooktheryan oops, I meant the internal load balancer.

cooktheryan:
ahh ok I got you.. Will update momentarily

PR #16, opened 2016-09-16
Fix of missing var and prep for containerized install

cooktheryan:


reference-architecture/aws-ansible/playbooks/roles/cloudformation-infra/tasks/main.yaml
```diff
@@ -22,23 +22,23 @@
       MasterInstanceType: "{{ master_instance_type }}"
       AmiId: "{{ ami }}"
       BastionInstanceType: "{{ node_instance_type }}"
-      BastionUserData: "{{ lookup('template', 'files/user_data_bastion.yml') | b64encode }}" 
+      BastionUserData: "{{ lookup('file', 'roles/cloudformation-infra/files/user_data_bastion.yml') | b64encode }}"
```
detiber: @abutcher do you remember the file lookup precedence? I would expect `user_data_bastion.yml` to work here, but I think there might be a version-based dependency where you might have to use `files/user_data_bastion.yml` in some cases. Either way, `roles/cloudformation-infra/files/user_data_bastion.yml` would limit things to where ansible being run outside of the directory that contained `roles` would no longer work.

abutcher: `lookup('template', 'user_data_bastion.yml')` works in testing with 2.2.

cooktheryan: It does look like template is more forgiving in regards to location.

PR #17, opened 2016-09-21
Containerized installation of Ref Arch.

cooktheryan:
Commit uses the generic {{openshift.common.admin_binary }} rather than oadm and {{openshift.common.client_binary }}. Also, skips certain steps if installation is containerized.

Ready for review but won't be merged until documentation changes are made


reference-architecture/aws-ansible/playbooks/roles/prerequisites/tasks/main.yaml
```diff
@@ -11,3 +11,4 @@
       - bash-completion
       - docker
       - python-httplib2
+  when: ansible_pkg_mgr != "unknown"
```
detiber: If this requires ansible 2.x+, you can use the `package` module rather than yum.

reference-architecture/aws-ansible/playbooks/roles/rhsm-repos/tasks/main.yaml
```diff
@@ -1,11 +1,12 @@
 ---
 - name: disable unneeded repos
   command: subscription-manager repos --disable='*'
-  when: register_result | changed
+  when: register_result | changed and (ansible_os_family == "RedHat" and ansible_distribution != "Fedora")
```
detiber: Doesn't take CentOS into account.

reference-architecture/aws-ansible/playbooks/roles/validate-app/library/openshift_facts.py
```diff
@@ -0,0 +1,2162 @@
+#!/usr/bin/python
```
detiber: I'd try to avoid vendoring openshift_facts if at all possible, I would symlink to the version of openshift-ansible being used rather than vendor.

reference-architecture/aws-ansible/playbooks/roles/prerequisites/tasks/main.yaml
detiber: shouldn't need ansible_pkg_mgr here at all with the package module, it should handle it on it's own on the backend.

reference-architecture/aws-ansible/playbooks/roles/rhsm-repos/tasks/main.yaml
```diff
@@ -1,11 +1,12 @@
 ---
 - name: disable unneeded repos
   command: subscription-manager repos --disable='*'
-  when: register_result | changed
+  when: register_result | changed and ansible_os_family == "RedHat"
```
detiber: I think ansible_os_family includes CentOS, I think you want to test for `ansible_distribution not in ['Fedora', 'CentOS']` as well

reference-architecture/aws-ansible/playbooks/roles/prerequisites/tasks/main.yaml
cooktheryan: Without it doing an install on Atomic "Could not find a module for unknown" so this way we can bypass if host is Atomic

detiber: Ah, we probably need to make sure that there is a default ansible fact that will expose an Atomic Host.

Until that is available, you can test for `/run/ostree-booted` or use the openshift.common.is_atomic fact from openshift_facts.

reference-architecture/aws-ansible/playbooks/roles/validate-app/library/openshift_facts.py
cooktheryan: resolved..Excellent suggestion as the fact file is subject to change

cooktheryan:
Trying this again in new PR. Too much outdated info

Issue #18, opened 2016-09-24
Registration Broken


mrobson:
It looks like a 'Subscription Name' update leads to the pool RegEx[1] matching nothing.

Subscription Name:   Red Hat OpenShift Container Platform, Premium, 2-Core
Subscription Name:   Red Hat OpenShift Container Platform, Standard, 2-Core
Subscription Name:   Red Hat OpenShift Container Platform Broker/Master Infrastructure

[1] https://github.com/openshift/openshift-ansible-contrib/blob/master/reference-architecture/aws-ansible/playbooks/roles/rhsm-subscription/tasks/main.yaml#L7

cooktheryan:
This should be resolved in merge #20

PR #20, opened 2016-09-26
subscription manager fix

cooktheryan:
Pool ID can be used or Subscription Name can be provided. Potential fix for #18


cooktheryan:
Document update will be pushed tomorrow then this will be merged

PR #24, opened 2016-09-29
OCP Version 3.3

cooktheryan:
Removal of unneeded items and change from repo 3.2 to 3.3


cooktheryan:
Validated install. Reference Arch documents include directions to use ansible-2.2.0-0.5.prerelease.el7.noarch due to an Ansible cloudformation error.  Error has been made aware to productization team

reference-architecture/aws-ansible/playbooks/openshift-setup.yaml
```diff
@@ -21,20 +21,23 @@
...
     deployment_type: openshift-enterprise
     openshift_master_identity_providers:
     - name: github
       kind: GitHubIdentityProvider
       login: true
       challenge: false
       mapping_method: claim
-      clientID: e72863337b0417387b53
-      clientSecret: a2230462295d6b579d25f46dd51eb25a170e7e60
+      clientID: e76865557b0417387b35
```
detiber: I just noticed that clientID and clientSecret are here. Doesn't seem like something that should be in repo.

cooktheryan: Garbage creds. In the document it calls out to swap with credentials created in previous steps. The credentials still allow for a deployment in CI using these fake creds just no GUI access.

PR #26, opened 2016-10-01
Add check for ansible version

Mangirdaz:
Add check for Ansible version


cooktheryan:
Verified code works as expected

reference-architecture/aws-ansible/playbooks/roles/pre-install-check/tasks/main.yaml
```diff
@@ -0,0 +1,9 @@
+---
+
+- name: Get current Ansible version on local host
+  command: command ansible --version
```
detiber: The ansible_version variable can be used rather than using the command module: https://github.com/openshift/openshift-ansible/blob/master/playbooks/common/openshift-cluster/verify_ansible_version.yml

PR #30, opened 2016-10-05
Per env vars

stuartbfox:


cooktheryan:
Very intrigued to give this a try. I'm OOO until Monday but will catch up with this then.

stuartbfox:
Im going to withdraw this for about a week, there are some more changes needed before it's usable

PR #34, opened 2016-10-12
Add node playbook and tasks

cooktheryan:
Plays to add either an infrastructure node or app node to the reference architecture deployment.

Note: A bulk of the code in the PR is due to updating the ec2.py script with the newest from upstream. The new ec2.py supports pulling instances ids better.


reference-architecture/aws-ansible/playbooks/roles/cloudformation-infra/files/add-infra-node.json
```diff
@@ -0,0 +1,131 @@
+{
```
detiber: Does this create an additional stack each iteration? If so, then this will start causing issues related to stack limits for long running environments. It might be better to just use the native ansible modules for scale-up until ASGs are in use.

cooktheryan: Ya I was back and forth about creating a new stack or using pure ansible. By using a new stack we have a bit of consistency because we launch the everything that way but then you may have to call aws to increase your limits in the future.  I prefer the stack method because I have the assurance that the host and dns are ready by the time the next play starts without having to write a readiness check. Also, I can reuse some of the cloudformation role that is already established.

detiber: @cooktheryan I just think about how to manage the environment ongoing...  when scaling/tearing down the environment how do you know what stacks to destroy? When ASGs are in use, it is just a matter of scaling up/down the ASG through the existing stack. For multiple stacks, management is going to be wonky.

cooktheryan: Completely agree. This is just a stop gap until ASG. When cleaning up both scenarios kind of stink in regardless of pure ansible and a new stack because there is really no workflow in destroying the environment. I'm definitely not looking forward to going back to potential API errors when launching with pure ansible. Ive been spoiled by the 100% consistency when launching with cloudformations. But going back this is just a stop gap of the ASG work that I really need to visit.

reference-architecture/aws-ansible/playbooks/roles/rhsm-repos/tasks/main.yaml
```diff
@@ -1,7 +1,6 @@
 ---
 - name: disable unneeded repos
   command: subscription-manager repos --disable='*'
-  when: register_result | changed
```
cooktheryan: Yep already queued up to remove

reference-architecture/aws-ansible/playbooks/roles/rhsm-repos/tasks/main.yaml
```diff
@@ -8,3 +8,12 @@
...
       - rhel-7-server-ose-3.3-rpms
+
+- name: remove rhui repos
+  file:
+    path: "{{ item }}"
+    state: absent
+  with_items:
+      - /etc/yum.repos.d/redhat-rhui-client-config.repo
+      - /etc/yum.repos.d/redhat-rhui.repo
+      - /etc/yum.repos.d/rhui-load-balancers.conf
```
detiber: Would it be better to disable these using yum-config-manger instead?

You could also leverage the rhui repos for base/extra packages (would need to enable rhui extras), and only use rhsm repos for the ose packages.

cooktheryan: rhui killed all of my playbook runs this morning due to a missing CDS. I could always do a regex or replace to disable the repo in the file. It doesn't look like yum-config-manager bites in disabling these repos sadly.

```
[root@ip-10-20-4-228 ~]# yum-config-manager --disable rhui-REGION-rhel-server-releases/7Server/x86_64
Loaded plugins: amazon-id, rhui-lb
[root@ip-10-20-4-228 ~]# yum-config-manager --disable rhui-REGION-rhel-server-rh-common/7Server/x86_64
Loaded plugins: amazon-id, rhui-lb
[root@ip-10-20-4-228 ~]# yum repolist
Loaded plugins: amazon-id, rhui-lb, search-disabled-repos, subscription-manager
Could not contact CDS load balancer rhui2-cds01.us-east-1.aws.ce.redhat.com, trying others.
repo id                                                                                        repo name                                                                                                     status
rhel-7-server-extras-rpms/x86_64                                                               Red Hat Enterprise Linux 7 Server - Extras (RPMs)                                                                297
rhel-7-server-ose-3.3-rpms/x86_64                                                              Red Hat OpenShift Container Platform 3.3 (RPMs)                                                                  542
rhel-7-server-rpms/7Server/x86_64                                                              Red Hat Enterprise Linux 7 Server (RPMs)                                                                      11,396
rhui-REGION-client-config-server-7/x86_64                                                      Red Hat Update Infrastructure 2.0 Client Configuration Server 7                                                    6
rhui-REGION-rhel-server-releases/7Server/x86_64                                                Red Hat Enterprise Linux Server 7 (RPMs)                                                                      11,396
rhui-REGION-rhel-server-rh-common/7Server/x86_64                                               Red Hat Enterprise Linux Server 7 RH Common (RPMs)                                                               209
repolist: 23,846
[root@ip-10-20-4-228 ~]# yum install zsh
Loaded plugins: amazon-id, rhui-lb, search-disabled-repos, subscription-manager
Could not contact CDS load balancer rhui2-cds01.us-east-1.aws.ce.redhat.com, trying others.
```

cooktheryan: +1 It may have been a rage commit :dagger:  fix-up coming soon

cooktheryan: Disabling is now more sane

reference-architecture/aws-ansible/playbooks/roles/rhsm-repos/tasks/main.yaml
```diff
@@ -1,11 +1,20 @@
...
+
+- name: remove rhui repos
+  replace:
+    dest: "{{ item }}"
+    regexp: 'enabled=1'
+    replace: 'enabled=0'
+  with_items:
+      - /etc/yum.repos.d/redhat-rhui-client-config.repo
+      - /etc/yum.repos.d/redhat-rhui.repo
+      - /etc/yum.repos.d/rhui-load-balancers.conf
```
detiber: I think you can drop rhui-load-balancers.conf from the list, otherwise lgtm

PR #35, opened 2016-10-12
GCE: Replace HTTPS LB for master console with SSL Proxy

pschiffe:
Master console requires WebSockets, and HTTPS LB doesn't support that, so replace it with SSL Proxy.

### DIFF
--- a/reference-architecture/gce-cli/gcloud.sh
+++ b/reference-architecture/gce-cli/gcloud.sh
@@ -628,6 +628,30 @@ else
     echo "Bucket '${REGISTRY_BUCKET}' already exists"
 fi
 
+# Configure local SSH so we can connect directly to all instances
+ssh_config_file=~/.ssh/config
+if ! grep -q '# OpenShift on GCE Section' "$ssh_config_file"; then
+    echo -e '\n# OpenShift on GCE Section\n' >> "$ssh_config_file"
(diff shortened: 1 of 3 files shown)


detiber:
This LB is used just for external access, correct? I would have concerns about using the ssl proxy for internal api load balancing, since we rely on client certs for internal cluster communication with the api.

pschiffe:
Yes, that's correct, only for external access. For internal LB I'm using just network LB for that reason you described. But for external access, I just wanted to have control over the public facing HTTPS cert, so user can replace it with valid one, etc..

reference-architecture/gce-cli/gcloud.sh
```diff
@@ -628,6 +628,30 @@ else
...
+    instances=$(gcloud --project "$GCLOUD_PROJECT" compute instances list --filter='tags.items:ocp' --format='value(name)')
+    for i in $instances; do
+        echo "Host ${i}
+    User cloud-user
+    proxycommand ssh bastion -W %h:%p
+" >> "$ssh_config_file"
+    done
+    echo -e '# End of OpenShift on GCE Section\n' >> "$ssh_config_file"
+fi
+
```
themurph: This should probably be it's own script, or turned into a function since there is a very good chance it of it needing to be run a second or third time as hosts are moved, migrated, reprovisioned, etc. Other than that, LGTM

themurph:
LGTM

themurph:
@detiber if you're satisfied with the External SSL proxy question. Can you merge this branch when you get the chance?

PR #36, opened 2016-10-13
Advanced Openshift Install (Ansible) using Vagrant (multi machine)

petenorth:
As suggested by Jason DeTiberus  creating this pull request.

This project demonstrates an Openshift Enterprise 3.3 / Openshift Container Platform 3.3 advanced installation using ansible.

On my machine the entire process (excluding the creation of the initial RHEL 7.2 vagrant box) takes around 40 mins.

This is to create a single master / two node set up.

My laptop has 16 GiB, c. 7 GiB used after installation.

Intel® Core™ i7-3520M CPU @ 2.90GHz × 4

Note: this isn't targeting the 'oc cluster up' use case or the CDK use case. It is targeting users who need to become familiar with the Openshift advanced installation process using Ansible.


detiber:
@abutcher @tbielawa ptal, I would like to see if we can provide some suggestions for consolidating the two Vagrantfiles and also providing support for libvirt.

detiber:
@sdodson ptal as well

petenorth:
@detiber I'll look at consolidating and providing support for libvirt.

detiber:
@petenorth thanks, the existing Vagrantfile in openshift-ansible should help out with libvirt support.

vagrant/Vagrantfile
```diff
@@ -0,0 +1,139 @@
...
+
+hostname = Socket.gethostname
+localmachineip = IPSocket.getaddress(Socket.gethostname)
+puts %Q{ This machine has the IP '#{localmachineip} and host name '#{hostname}'}
+
+# Vagrantfile API/syntax version. Don't touch unless you know what you're doing!
+VAGRANTFILE_API_VERSION = '2'
+
+# Validate required plugins
+REQUIRED_PLUGINS = %w(vagrant-registration vagrant-hostmanager landrush)
```
detiber: probably want to limit vagrant-registration to just enterprise deployments, assuming origin deployments would be CentOS/Fedora only.

vagrant/Vagrantfile
```diff
@@ -0,0 +1,139 @@
...
+unless errors.empty?
+  msg = errors.size > 1 ? "Errors: \n* #{errors.join("\n* ")}" : "Error: #{errors.first}"
+  fail Vagrant::Errors::VagrantError.new, msg
+end
+
+if ENV['DEPLOYMENT_TYPE'] == 'enterprise'
+  box_name = 'rhel/7.2'
+else
+  box_name = 'centos/7'
+end
```
detiber: Should probably have an option for Fedora here as well, maybe a combination of DEPLOYMENT_TYPE and HOST_OS, where 'enterprise' is only valid for RHEL and origin could be any of the above.

vagrant/Vagrantfile
```diff
@@ -0,0 +1,139 @@
...
+#sudo yum clean all ; yum repolist
+#sudo yum -y update
+sudo yum -y install wget git net-tools bind-utils iptables-services bridge-utils bash-completion
+
+# Vagrant's "change host name" capability for Fedora
+# maps hostname to loopback, conflicting with hostmanager.
+# We must repair /etc/hosts
+
+sudo sed -ri 's/127\.0\.0\.1\s.*/127.0.0.1 localhost localhost.localdomain/' /etc/hosts
+SCRIPT
```
detiber: This looks static, any reason not to have this as just a file alongside the Vagrantfile?

Also, is there an issue with the ssh key injection with Vagrant? We shouldn't need the root password.

vagrant/Vagrantfile
```diff
@@ -0,0 +1,139 @@
...
+      admin1.vm.provision "shell", inline: "echo \"updating deployment type\""
+      admin1.vm.provision "shell", inline: "sudo cp /vagrant/update_deployment_type.sh /root"
+      admin1.vm.provision "shell", inline: "sudo /root/update_deployment_type.sh"
+    end
+
+  end
+
+  if ENV['DEPLOYMENT_TYPE'] == 'enterprise'
+    config.vm.provision "shell", inline: "echo \"setting up Red Hat subscriptions\""
+    config.vm.provision "shell", path: "repos.sh"
```
detiber: I'm not sure there is a guarantee that this step is run prior to the admin1 provisioning.

vagrant/Vagrantfile
petenorth: Yes agreed (was on my list!)

petenorth:
Jason,

I'm fairly close to getting the ansible local provisioner working but there seems to be some differences in behaviour between the origin install and enterprise with respect to the --limit option.

The ansible provisioner seems to use the --limit option regardless of whether it is specified.

For Origin the only value which works for me is --limit="*", --limit="all" doesn't work.

For Enterprise neither --limit="*" or --limit="all" doesn't work.

The ansible version that gets installed is different:

Origin:

ansible 2.1.1.0
  config file = /etc/ansible/ansible.cfg
  configured module search path = Default w/o overrides

Enterprise

ansible 2.2.0
  config file = /etc/ansible/ansible.cfg
  configured module search path = Default w/o overrides

Any ideas?

Peter.

petenorth:
Update on the --limit question, think I've got it, need to specify an empty string i.e. --limit=""

detiber:
@petenorth are you specifying the limit as a command line flag, or as an option within the vagrant config?

petenorth:
@detiber the ansible local provisioner sets it to --limit="all" if you don't specify it,

petenorth:
Moved to ansible local provisioner.
Using vagrant user and vagrant private keys (as suggested in ansible local vagrant documentation).

Outstanding:

1) required plugin checks.
2) libvirt provider.
3) clean up Vagrantfile (set deployment type as a variable once rather rather then endless ENV[]).

petenorth:
Some cleanup of Vagrantfile complete.
Required plugins fixed.

Outstanding:

libvirt provider.

petenorth:
@detiber I maybe seeing some instability with the ansible local install.  If confirmed the main difference between what I had originally and the current state of this pull request would be the use of the vagrant user for the install and/or the use of --limit="".

detiber:
@petenorth can you elaborate on the instability you are seeing?

petenorth:
I have only seen it once and the messages I was getting were that the client etcd was not configured correctly (this is from memory). I only saw it once and I noticed that the router and registry had not finished deploying .

A fresh run and making sure that the registry and router were running correctly seemed to fix the issue.

detiber:
@petenorth working up a few suggestions, PR incoming after I finish some testing (with libvirt support as well).

petenorth:
@detiber OK Thanks.

detiber:
@petenorth Just started some more testing. I'm not seeing the same issue you are with the hosts file not being updated properly. It seems to be working correctly (at least for CentOS7/Ansible 2.1).

One thing that I am noticing is that if I do not set VAGRANT_LOG=debug, them I am hitting weird deadlock issues on provisioning with libvirt. The issue seems to go away once I set logging to debug, though.

detiber:
@petenorth My previous test run has completed and I see a deployed router and registry. I haven't tested any further, though.

detiber:
@petenorth I'm going to go ahead and merge this, and any further issues can be sorted out through additional issues/PRs.

PR #38, opened 2016-10-13
Idempotent version of local SSH config

pschiffe:
If the script is rerun after the deployment was reverted, local SSH configuration needs to be updated to match the current set up.

For now I'm not putting this into separate function because it's used only on one place, but later it will be probably refactored once we start adding features for adding nodes, etc.


themurph:
LGTM

PR #39, opened 2016-10-14
Large structural change to make this repo more usable in production

stuartbfox:
I just pull from your master and notice you've added scaling by calling a playbook. I took a different approach by adding cli arguments (--master-instance-count --infra-instance-count --node-instance count) and letting ose-on-aws.py take care of it. I've tested many times and it works.
Other major features :-
# Stack naming
Propagate stack_name everywhere so you can have multiple stacks per region

# Per environment vars file
Separate vars_file per env, maybe you want smaller instance types for a dev cluster

# Scaling
See above

# Jinjifying the cfn templates
Required by the scaling work

# Ability to merge additional cfn features into the main cfn template.
My use case is to deploy rds instances into the same stack.


detiber:
@cooktheryan ptal

cooktheryan:
This is beyond awesome. The one concern I have though is that we are launching everything in one AZ(PrivateSubnet1).

stuartbfox:
Hmm good point, I missed that. Let me have a think on how to solve it.
Quick question, is the scheduler AZ aware now?

stuartbfox:
Turned out to be an easy fix :)

cooktheryan:
@stuartbfox awesome turn around. I ran the first set of validation tests with the following.

```
./ose-on-aws.py --rhsm-user=rhit_rcook --rhsm-password=nopenopenope --public-hosted-zone=rcook-aws.sysdeseng.com --keypair=OSE-key --rhsm-pool="Employee SKU" --master-instance-type=t2.medium --node-instance-count=3 --infra-instance-count=3
```

but encountered the following

```
TASK [instance-groups : Add bastion to group] **********************************
[DEPRECATION WARNING]: Skipping task due to undefined Error, in the future this will be a fatal error.: 'dict object' has no attribute 'tag_openshift-role_bastion'.
This feature will be removed in a future 
release. Deprecation warnings can be disabled by setting deprecation_warnings=False in ansible.cfg.
fatal: [localhost]: FAILED! => {"failed": true, "msg": "The conditional check 'hostvars[item]['ec2_tag_aws_cloudformation_stack-name'] == \"{{ stack_name }}\"' failed. The error was: error while evaluating conditional (hostvars[item]['ec2_tag_aws_cloudformation_stack-name'] == \"{{ stack_name }}\"): 'item' is undefined\n\nThe error appears to have been in '/home/rcook/openshift-ansible-contrib/reference-architecture/aws-ansible/playbooks/roles/instance-groups/tasks/main.yaml': line 2, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n---\n- name: Add bastion to group\n  ^ here\n"}
```

stuartbfox:
Fixing this now, I suspect it happened when I merged your master. brb :)

stuartbfox:
Fixed

cooktheryan:
testing this now

cooktheryan:
@stuartbfox looks like S3 complains if the stack name has Capital letters.

cooktheryan:
Also in the teardown make a quick change to greenfield.json to template: roles/cloudformation-infra/files/greenfield.json.j2

stuartbfox:
Interesting on the caps stack name, I didnt know that. I could put a .to_lower filter on the stack name.

I've also just noticed another limitation of using jinja to generate the cfn templates as is. Above 7 app nodes causes the file to become too large, 50k+. It work in #41 to use nested stacks with ASG for app and infra nodes might be the better route. Stack naming and per env vars files are the bits I'd really need

cooktheryan:
@stuartbfox haven't forgotten about you but a general question in regards to standing up multiple stacks. Would you use the same bastion for all the stacks (multiple openshift deployments) or a 1:1 relationship where each stack has its own bastion.

stuartbfox:
I would have 1 bastion per stack, unless you are going to  force the use of a byo bastion for all stacks.
Could get tricky using the bastion deployed in stack1 cfn to jump thru for install a byo stack2. What happens if you decommission stack1? for the small cost of a t2.small, I dont think its worth relying on an external bastion

stuartbfox:
@cooktheryan  Thanks for making the s3 and stack names stuff working. Do you want me to close this PR? I have at least 2 other features in here I'd like reviewed; per stack vars and custom template injection

cooktheryan:
Ya let's close it and start a new one with those two items.

PR #41, opened 2016-10-17
start of nested stacks

detiber:


detiber:
@cooktheryan working branch for nested stack development.

cooktheryan:
@detiber 

I know this is super MVP but is there a way that we can provide this as an input in the template or at run-time to be bypassed.

````
[rcook@localhost cloudformation]$ grep -i "s3" ./* -R
./files/nested/three_master_infra_asg_node_asg_no_bastion.yaml:    Default: 'https://s3.amazonaws.com/openshift-cloudformation-templates/vpc/default.yaml'
./files/nested/three_master_infra_asg_node_asg_no_bastion.yaml:    Default: 'https://s3.amazonaws.com/openshift-cloudformation-templates/security-groups/default.yaml'
./files/nested/three_master_infra_asg_node_asg_no_bastion.yaml:    Default: 'https://s3.amazonaws.com/openshift-cloudformation-templates/iam-profiles/default.yaml'
./files/nested/three_master_infra_asg_node_asg_no_bastion.yaml:    Default: 'https://s3.amazonaws.com/openshift-cloudformation-templates/control-plane/default.yaml'
````

**Comment by detiber** [comment]
@cooktheryan indeed, they are already exposed as parameters in the template, just need to expose them in the template_parameters for the cloudformation task.

**Comment by stuartbfox** [comment]
Any movement on this? Not being able to specify the amount of nodes at deploy time via either ASG or jinja rendering is a bit of a blocker

**Comment by cooktheryan** [comment]
@stuartbfox @detiber has been running this mainly but other items have popped up in his schedule.

From a high level though i believe the first iteration will not include a bastion if that makes any difference

**Comment by detiber** [comment]
@stuartbfox @cooktheryan I hope to pick this back up later this week.

**Comment by cooktheryan** [comment]
@detiber you may want to swap in the ec2_zones_by_region.py you patched up for me. I was failing super early on in the playbook run during vpc creation due to ec2 classic

**Comment by cooktheryan** [comment]
All items launched as expected

**Comment by detiber** [comment]
@cooktheryan I think the latest commit addresses the ec2_zones_by_region issue.

**Comment by stevekuznetsov** [comment]
@detiber Do you think it would be wise to use the k8s `master` and `minion` roles [here](https://github.com/kubernetes/kubernetes/tree/master/cluster/aws/templates/iam) as the base for our `master` and `node` roles, then add a whitelist of permissions we see OpenShift needing that the Kubernetes roles don't give us on top?

File: playbooks/provisioning/aws/cloudformation/files/nested/control-plane/default.yaml [hunk 1]
```diff
@@ -0,0 +1,229 @@
...
+        Timeout: 2
+        UnhealthyThreshold: 2
+        Target:
+          'Fn::Join':
+          - ''
+          - - 'Fn::Join':
+              - ':'
+              - - 'HTTPS'
+                - Ref: MasterApiPort
+            - /healthz/ready
```
stevekuznetsov: What's going on here? Why is the target `https:443/healthz/ready` ? Don't we need a hostname?

detiber:
@stevekuznetsov We actually need fewer permissions than what the kubernetes iam permissions require. The ones that we are using now are a bit overly broad and can be further restricted.

playbooks/provisioning/aws/cloudformation/files/nested/control-plane/default.yaml
detiber: No, no need to specify the hostname, This specifies an https health check using port 443 and the path /healthz/ready. While it looks odd, it is the correct format for an ELB health check target.

tomassedovic:
Hey folks! I'd like to do something similar for OpenStack and I've a few questions, if you don't mind. 

1. How is this exepected to be used? Running `ansible-playbook nested_setup.yml`, waiting for it to finish, get/write the inventory of the created nodes and then running the openshift-ansible installer against the environment created by the playbook here? Or is there something that integrates it all into a single command?

2. What's the relationship between `cloudformation_setup.yml` and `nested_setup.yml`? As far as I can see, you'd use one or the other, but not together. Is that correct?

3. This playbook creates the master nodes, but not the infra or app ones. Are they going to be created elsewhere or have they not just made it into this patch yet?

4. What's the root volume for? The instance's root partition?

5. As far as I can tell, the master01, master02 and master03 nodes are identical. Why not create an autoscaling/resource group (not sure what the CloudFormation name for this is) and set the count to 3? Does it have to do with the subnets associated with each node?

6. What is the `ec2_zones_by_region` lookup for? Getting the availability zones so we have each master node in a different AZ to make the deployment more resilient?

Apologies for the stream of questions. Thanks!

stevekuznetsov:
I don't have all the answers for you, but I have some:

>  get/write the inventory of the created nodes and then running the openshift-ansible installer against the environment created by the playbook here?

The appropriate way to interface a provisioner script with other playbooks is through a dynamic inventory. I'm working on some updates to this patch that will use the default Ansible dynamic EC2 inventory and add support for host groups that OpenShift-Ansible needs.

> is there something that integrates it all into a single command?

There may be in the future, but there is not today. I assume the packaged and disseminated installer will have a nice flow for this, but that's a ways off.

> What's the relationship between `cloudformation_setup.yml` and `nested_setup.yml`? 

The former has some hard-coded values for faster startup, but yes they achieve similar end results. The latter is the one meant for general consumption.

> This playbook creates the master nodes, but not the infra or app ones. Are they going to be created elsewhere or have they not just made it into this patch yet?

This is still WIP. I have a patch for this patch that adds things like this, but it's not quite ready yet.

> What's the root volume for? The instance's root partition?

Yes, where the OS + friends live.

> Why not create an autoscaling/resource group (not sure what the CloudFormation name for this is) and set the count to 3?

@detiber and I had talked about this ... I don't know what is more common for AWS but an auto-scaling group that is never intended to scale seems weird. In my patch we generate the config for each of these nodes using a template so there isn't a chance that there will be drift between hard-coded master node instances, anyway.

tomassedovic:
Thanks for your reply @stevekuznetsov. I understand this is WIP, I just wondered whether I was missing regarding app nodes.

PR #50, opened 2016-10-19
Detect invalid RH username, password or pool id

### LINKED ISSUE #44: gce-cli: Doesn't handle subscription failures due to bad pool IDs gracefully (should terminate)
Running in my setup, using the qcow downloaded from server.  Is this because i'm not in the right pool for my subscription?

```
+ gcloud --project openshift-gce-devel compute copy-files /Users/clayton/projects/origin/src/github.com/openshift/openshift-ansible-contrib/reference-architecture/gce-cli/ansible-config.yml cloud-user@bastion: --zone us-central1-a
Warning: Permanently added 'compute.5845721122299694843' (ECDSA) to the list of known hosts.
ansible-config.yml                                                                   100%  779     0.8KB/s   00:00
+ gcloud --project openshift-gce-devel compute ssh cloud-user@bastion --zone us-central1-a --ssh-flag=-t --command 'sudo sh -c '\''
    yum install -y python-libcloud atomic-openshift-utils;

    if ! grep -q "export GCE_PROJECT=openshift-gce-devel" /etc/profile.d/ocp.sh 2>/dev/null; then
        echo "export GCE_PROJECT=openshift-gce-devel" >> /etc/profile.d/ocp.sh;
    fi
    if ! grep -q "export INVENTORY_IP_TYPE=internal" /etc/profile.d/ocp.sh 2>/dev/null; then
        echo "export INVENTORY_IP_TYPE=internal" >> /etc/profile.d/ocp.sh;
    fi
'\'''
Loaded plugins: search-disabled-repos
google-cloud-compute/signature                                                                  |  454 B  00:00:00
google-cloud-compute/signature                                                                  | 1.4 kB  00:00:00 !!!
google-cloud-compute/primary                                                                    | 1.8 kB  00:00:00
google-cloud-compute                                                                                               4/4
No package python-libcloud available.
No package atomic-openshift-utils available.
Error: Nothing to do
Connection to 104.155.172.153 closed.
```

pschiffe:
When user provides invalid username, password or pool id without openshift
rpms, the script will detect it early and terminate.

Fixes #44

### DIFF
--- a/reference-architecture/gce-cli/gcloud.sh
+++ b/reference-architecture/gce-cli/gcloud.sh
@@ -190,8 +190,8 @@ function revert {
     fi
 
     # Master backend service
-    if gcloud --project "$GCLOUD_PROJECT" beta compute backend-services describe "$MASTER_SSL_LB_BACKEND" &>/dev/null; then
-        gcloud -q --project "$GCLOUD_PROJECT" beta compute backend-services delete "$MASTER_SSL_LB_BACKEND"
+    if gcloud --project "$GCLOUD_PROJECT" beta compute backend-services describe "$MASTER_SSL_LB_BACKEND" --global &>/dev/null; then
+        gcloud -q --project "$GCLOUD_PROJECT" beta compute backend-services delete "$MASTER_SSL_LB_BACKEND" --global
     fi
 
     # Master health check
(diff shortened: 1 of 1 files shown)


pschiffe:
@smarterclayton what do you think?

pschiffe:
@detiber that and `--command "sudo sh -c '` vs `--command "sudo bash -euc '`. With `bash -eu` the "inner" script will terminate and fail instantly when any of the commands fails. So if you provide wrong username or password, it will fail directly at `subscription-manager register` command. And the `yum -q list` fails if the package doesn't exist, so that is a last safeguard.

detiber:
@pschiffe ah, overlooked that when scanning from my phone, thanks!

themurph:
LGTM

Issue #56, opened 2016-10-20
Having problems with AWS deployment - ERROR: "Forbidden", while: getting RDS instances


alberttwong:
command
```
./ose-on-aws.py --create-key=yes --rhsm-user=alwong@redhat.com --rhsm-password=XXXX --public-hosted-zone=ocp.alberttwong.com --key-path=/root/.ssh/id_rsa.pub --keypair=us-east-1 --rhsm-pool=Employee
```
error message
```
ERROR: "Forbidden", while: getting RDS instancesERROR! Inventory script (inventory/aws/hosts/ec2.py) had an execution error: ERROR: "Forbidden", while: getting RDS instances
```

modified my IAM policy to below with S3FullAccess
```
{
   "Version": "2012-10-17",
   "Statement": [
       {
           "Sid": "Stmt1459269951000",
           "Effect": "Allow",
           "Action": [
               "cloudformation:*",
               "iam:*",
               "route53:*",
               "elasticloadbalancing:*",
               "ec2:*",
               "cloudwatch:*",
               "autoscaling:*"
           ],
           "Resource": [
               "*"
           ]
       }
   ]
}
```

alberttwong:
also tried in ec2.ini

```
rds = False
```

but get

```
ERROR: "Forbidden", while: getting ElastiCache clustersERROR! Inventory script (inventory/aws/hosts/ec2.py) had an execution error: ERROR: "Forbidden", while: getting ElastiCache clusters
```

cooktheryan:
I have this fix in a PR that's already approved give me 5 minutes.

Ryan

alberttwong:
just did a pull on the newest code from git... trying again.

alberttwong:
new error in cloudformation

```
Value (us-east-1a) for parameter availabilityZone is invalid. Subnets can currently only be created in the following availability zones: us-east-1c, us-east-1e, us-east-1d, us-east-1b.
```

cooktheryan:
Ill need to look into this as I have not seen it before.

alberttwong:
I had this problem with demo-ansible too.... switching to us-west-2 (Oregon).

alberttwong:
Just for fun... also attached the following policies.

AmazonRDSFullAccess, AmazonEC2FullAccess, AmazonElastiCacheFullAccess, AmazonS3FullAccess, AmazonRoute53FullAccess 

with the custom policy I posted above.

alberttwong:
[...]

PR #59, opened 2016-10-20
handle ec2 classic accounts better

detiber:
Resolves: #56

### DIFF
--- a/reference-architecture/aws-ansible/inventory/aws/hosts/ec2.ini
+++ b/reference-architecture/aws-ansible/inventory/aws/hosts/ec2.ini
@@ -61,3 +61,9 @@ cache_max_age = 900
 # destination_variable and vpc_destination_variable.
 # destination_format = {0}.{1}.rhcloud.com
 # destination_format_tags = Name,environment
+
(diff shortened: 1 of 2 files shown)


cooktheryan:
@detiber i tried to launch with the new script and hit 

```
fatal: [localhost]: FAILED! => {"failed": true, "msg": "{{ lookup('ec2_zones_by_region', region) }}: Could not lookup zones for region: us-east-1\nexception: 'VPC' object has no attribute 'isDefault'"}
```
same for west

````
fatal: [localhost]: FAILED! => {"failed": true, "msg": "{{ lookup('ec2_zones_by_region', region) }}: Could not lookup zones for region: us-west-2\nexception: 'VPC' object has no attribute 'isDefault'"}
````

detiber:
@cooktheryan that's what I get for testing in a repl rather than the actual filter...  fix incoming

detiber:
@cooktheryan mind trying again?

cooktheryan:
@detiber testing now

cooktheryan:
Tested against us-east-1 with success! Excellent work @detiber :+1:

PR #63, opened 2016-10-20
More restrictive firewall

pschiffe:
Allow only what's really necessary


themurph:
Confirmed (offline) that the newly named FW Rules will still get picked up by the provisioning script.

LGTM

PR #65, opened 2016-10-20
DO NOT MERGE - Example of customizations done for OpenShift CI system

smarterclayton:
@themurph this is a lot of changes and I don't think they're necessary to merge, but wanted to document what I was doing w.r.t. the CI install in a way you could see.
1. Move bastion out of the critical path and offer a containerized option that can be run outside of GCE
2. Split the image initialization to be separate, since we will likely be building a custom image from another CI path
3. Add support for startup-script and downloading instance data to the instances so we can put our custom RPMs for openshift into that and have the installer use that
4. Run some steps in parallel to dramatically speed them up
5. Fixes for the deployment_type and ansible_host issues

Depending on what you guys want to do with deployment manager / changes some of these are not relevant, but this is likely to be our custom base for deploying the openshift CI system onto GCE and we'll continue to iterate and try to push features back.


smarterclayton:
@detiber in case you're interested in what I'm doing here.

detiber:
@smarterclayton think the script to spin up the actual instances is missing, and running the gcloud.sh script seemed to not be the answer.

smarterclayton:
That's gcloud.sh

pschiffe:
@smarterclayton nice work, thanks. I'll be looking into this and probably I'll try to incorporate the changes eventually (or accept PRs), but not before our paper is released.

BTW, it seems that deployment_type is deprecated in favour of openshift_deployment_type:
https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_facts/tasks/main.yml#L26

smarterclayton:
So in the immediate near term OpenShift CI wants to call provisioners to stand up clusters against cloud.  We need something to do that.  I'd prefer the code to be standard, be shared, and be part of this contrib.  We will likely have more sophisticated requirements than customers at first, but I don't want duplicate code.  We're going to start with GCE and work backwards to AWS.  We *must* be able to run in a container (I'm not going to mess with a second system to deploy ansible versions on machines) - we are going to do that by putting openshift-ansible + openshift-ansible-contrib in a container build and having that be reusable.  I'd like that to be here, and make it so that when you want to deploy OpenShift repeatably we test it here first.

So, my proposal is:

1. I'm going to create a docker image (similar to here) that contains the full set of supported provisioners, but start with GCE / AWS.  I'd like that to overlap with ref arch as much as possible
  1. Primary reason for that is getting the correct dependencies set up is too hard, and there is no reason I should have to install my own gcloud when I could just be running an image
2. The interface for the image will be bind mounting in secrets (for the cloud providers) and config
3. We will set up a CI chain that tests that this image can deploy, and gate merges that break deployment (either breaks to openshift-ansible or breaks to openshift-ansible-contrib)
4. We'll tie this into the CI deployment system for origin and have reuse for other CI systems

I will start with a fork of this repo (so you guys aren't gated by that) but will slowly drip feed pieces in as we agree on them.  So smarterclayton/openshift-ansible-contrib probably to start.

smarterclayton:
Further updates - I'm managing a branch right now in this repo `origin-ci` which will contain the changes and is built to `openshift/gce-cloud-installer:latest` (I'm probably going to rename it to `openshift/origin-install-gce` or similar at some point).

smarterclayton:
Also, I disabled requiretty in ./gcloud-image because it blocks pipelining in Ansible.

PR #71, opened 2016-10-24
Replace kube proxy with haproxy for HTTP health check

### LINKED ISSUE #37: Google Cloud refarch http proxy shouldn't just forward 8080 to the api service
Instead of leveraging kubeproxy, it would probably be better to stand up either apache or haproxy to forward only the healthz/ready endpoint rather than proxy the entire api behind 8080.

pschiffe:
Fixes #37


pschiffe:
@detiber what do you think?

PR #72, opened 2016-10-24
Exit with nonzero when ansible-playbook fails.

jaryn:
Let's just propagate the exit status of the last executed ansible-playbook command.


cooktheryan:
Good find @jaryn

Issue #73, opened 2016-10-24
s3-registry-user peeks nonexisting `s3user.user_meta`


jaryn:
```
./ose-on-aws.py --rhsm-user=$USERNAME --rhsm-password=$PASS --public-hosted-zone=jhenner.mooo.com --keypair=jenkins --ami=ami-775e4f16 --rhsm-pool=ES0113909 --no-confirm --region=us-west-2
Configured values:
	ami: ami-775e4f16
	region: us-west-2
	master_instance_type: m4.large
	node_instance_type: t2.medium
	bastion_instance_type: t2.micro
	keypair: jenkins
	create_key: no
	key_path: /dev/null
	create_vpc: yes
	vpc_id: None
	private_subnet_id1: None
	private_subnet_id2: None
	private_subnet_id3: None
	public_subnet_id1: None
	public_subnet_id2: None
	public_subnet_id3: None
	byo_bastion: no
	bastion_sg: /dev/null
	console port: 443
	deployment_type: openshift-enterprise
	public_hosted_zone: jhenner.mooo.com
	app_dns_prefix: apps
	apps_dns: apps.jhenner.mooo.com
	rhsm_user: ******
	rhsm_password: *******
	rhsm_pool: ES0113909
	containerized: False
```

```
TASK [s3-registry-user : Create S3 OpenShift registry user] ********************

ok: [localhost]

TASK [s3-registry-user : Set fact] *********************************************
fatal: [localhost]: FAILED! => {
    "failed": true
}

MSG:

the field 'args' has an invalid value, which appears to include a variable that is undefined. The error was: 'dict object' has no attribute 'user_meta'

The error appears to have been in '/home/jenkins/workspace/ose-on-aws-deploy/openshift-ansible-contrib/reference-architecture/aws-ansible/playbooks/roles/s3-registry-user/tasks/main.yaml': line 9, column 3, but may
be elsewhere in the file depending on the exact syntax problem.

The offending line appears to be:

  register: s3user
- name: Set fact
  ^ here

```

This is with the commit 8c52fb6e989dd0c3ec44a6c3168813ea8e44c9b1

cooktheryan:
@jaryn This usually occurs when the s3_username=openshift-s3-docker-registry already exists. Currently there is WIP to fix this. This can be modified manually by changing the var s3_username in vars/main.yaml.

Issue #43, opened 2016-10-18
AWS - S3 task cannot be run more than once


stuartbfox:
Hi

Im having an issue I dont know how to solve. Basically you cannot run the s3 user tasks more than once and retrieve the required data structure.

If a user exists and this task runs, you do not get the s3user.user_meta data needed to run the subsequent tasks. ie

```
TASK [s3-registry-user : debug s3] *********************************************
task path: playbooks/roles/s3-registry-user/tasks/main.yaml:9
ok: [localhost] => {
    "s3user": {
        "changed": false,
        "created_keys": [],
        "groups": null,
        "keys": {
            "CHANGED": "Active"
        },
        "user_name": "apim"
    }
}
```

Which results in 

```
TASK [s3-registry-user : Set fact] *********************************************
task path: playbooks/roles/s3-registry-user/tasks/main.yaml:11
fatal: [localhost]: FAILED! => {
    "failed": true, 
    "msg": "the field 'args' has an invalid value, which appears to include a variable that is undefined. The error was: 'dict object' has no attribute 'user_meta'\n\nThe error appears to have been in 'playbooks/roles/s3-registry-user/tasks/main.yaml': line 11, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n  debug: var=s3user\n- name: Set fact\n  ^ here\n"
}
```

This means that any method to add an infra node will fail and any run further run of `ose-on-aws.py` will not complete due to missing data structures. To get round this could mean pre-creating the s3 user and storing the aws secret and passing in as a param but that seems clunky.

cooktheryan:
I wonder if we can go the route of  when: s3_user_access is defined and same for the secret

stuartbfox:
I was just looking at moving this to the cfn template http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-iam-accesskey.html

Generate a secret key in code thats consistent, hash a known value so it doesnt change between runs and doesnt need to be stored. It would simplify playbooks a little by having less to do. Also the user would be removed at stack deletion time.

Im currently trying 

```
---
- name: Create S3 OpenShift registry user
  iam:
    iam_type: user
    name: "{{ s3_username }}"
    state: present
    access_key_state: create
  register: s3user
- name: debug s3
  debug: var=s3user
- name: Set fact
[...]

PR #75, opened 2016-10-24
s3 and multistack attempt based of stuartbfox work

cooktheryan:


reference-architecture/aws-ansible/playbooks/roles/cloudformation-infra/files/brownfield-byo-bastion.json
```diff
@@ -708,6 +708,9 @@
             },
             { "Key": "openshift-role",
               "Value": "master"
+            },
+            { "Key": "stack-name",
+              "Value": {"Ref": "AWS::StackName"}
```
detiber: Why not define this at using the tags param on the cloudformation stack? Any tags set at the stack level should filter down to all stack-created objects.

detiber: That said, I think they already get assigned a dedicated tag associated with the stack: `aws:cloudformation:stack-name`

cooktheryan:
The last piece is making s3 bucket names unique. Some reason its not picking up the epoch time anymore. Will look into later tonight or tomorrow.

reference-architecture/aws-ansible/playbooks/roles/s3-registry-bucket/tasks/main.yaml
```diff
@@ -1,6 +1,10 @@
 ---
 - name: set the name of the bucket
-  set_fact: s3_bucket="{{ ansible_date_time.epoch }}-openshift-docker-registry"
+  set_fact: s3_bucket="{{ ansible_date_time.epoch }}-openshift-docker-registry-{{ stack_name}}"
```
detiber: why the epoch for the s3 bucket here? Isn't that going to make it so that if this role is re-applied that we are going to create a new s3 bucket each time?

stuartbfox: :+1:  good point

cooktheryan: @detiber I added the epoch time here due to the S3 buckets must be unique regardless of region. So for example if an org would use prod-openshift-infra and then another org blindly uses the same thing then the installation will fail at the s3 point.  

Also, the s3 bucket creation is so far down in the deployment that I would think the role wouldn't be re-applied.

cooktheryan:
resolves #73 #43

cooktheryan:
The version of ansible compiled did not have the module. I believe it will
be in the official rc.

reference-architecture/aws-ansible/playbooks/roles/s3-registry-bucket/tasks/main.yaml
detiber: @cooktheryan seems a bit fragile, that said if you do take this approach, I would suggest adding tags to the s3 bucket and adding some checks for a s3 bucket with the proper tags to conditionalize the creation.

For the nested_stacks, I'm planning on supporting env_id and cluster_id and requiring that combinations of those be globally unique, could probably leverage something similar for the s3 bucket...  might want an early "pre-req" check to validate that the bucket doesn't exist or exists in the right region, though.

cooktheryan: hrm. @detiber it feels like if I move s3 bucket creation up to the cloudformation play and output the bucket name it will allow for us to use the cloudformation lookup and not repeat the creation step. So even if we use epoch at time of creation it will always be defined in the cfn output?

Thoughts? 

cc @stuartbfox

detiber: @cooktheryan agree on creating it as part of the cloud formation template. With that in mind, I would probably leverage the stack_id rather than epoch, though.

PR #84, opened 2016-10-26
Make the cloud-formation name configurable

jaryn:
I will need this to test with different versions of Openshift in parallel. I hope it will be possible.


cooktheryan:
@jaryn actually am finishing up testing on #75 that should accomplish this but you were definitely on the right path

jaryn:
Good to hear (:

PR #86, opened 2016-10-28
ssh, variable and ansible fixes

dcritch:
A couple things in this PR.

On a fresh environment, some of the assumptions about the environment re: ssh are not correct. I put in some logic to account for that.

During testing, the openshift-ansible-playbooks rpm was updated, and seemed to break the openshift-install.yaml playbook. A variable was changed/added? And the cakephp template is no longer deployed. I tweaked the ansible playbook a bit to address those changes.

On my tests, the bucket for the RAW image was already present but created by a different user? So I could not delete or use it. I variable'ized it as IMAGE_BUCKET and updated the sample config accordingly.

I also screwed up a couple of installs by not updating my FQDN everywhere, so I flipped the hostnames to dervice the domain from the variable already present, $DNS_DOMAIN.


pschiffe:
Thanks for the PR!

PR #104, opened 2016-11-11
Atomic sshfs etc

cgwalters:
This is a bit of a work in progress.  If you agree with the direction though I can split up the first two patches into a separate PR.

### DIFF
--- a/vagrant/Vagrantfile
+++ b/vagrant/Vagrantfile
@@ -13,11 +13,16 @@ VAGRANTFILE_API_VERSION = '2'
 deployment_type = ENV['DEPLOYMENT_TYPE'] || 'origin'
 origin_os = ENV['ORIGIN_OS'] || 'centos'
 rhsm_pool = ENV['RHSM_POOL'] || 'Employee SKU'
+atomic = ENV['ATOMIC'] || 'no'
+landrush = ENV['LANDRUSH'] || 'yes'
+sync_type = ENV['SYNC_TYPE'] || ''
 
+REQUIRED_PLUGINS = %w(vagrant-hostmanager)
 if deployment_type == 'openshift-enterprise'
-  REQUIRED_PLUGINS = %w(vagrant-registration vagrant-hostmanager landrush)
-else
-  REQUIRED_PLUGINS = %w(vagrant-hostmanager landrush)
+  REQUIRED_PLUGINS.push('vagrant-registration')
+end
+if landrush != 'no'
+  REQUIRED_PLUGINS.push('landrush')
 end
 
 errors = []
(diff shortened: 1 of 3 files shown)


vagrant/Vagrantfile
```diff
@@ -146,29 +175,38 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
...
     }
 
+    if deployment_type == 'origin'
+      ansible_groups[:'OSEv3:vars'][:openshift_release] ='latest'
+    end
+    print("deployment_type = %s" % (deployment_type))
+
     ansible_host_vars = {
       master1:  {
+        ansible_ssh_common_args: '"-o StrictHostKeyChecking=no"',
```
detiber: This shouldn't be needed, the ansible.cfg has host_key_checking=false and should be copied over to the ansible.provisioning_path.

cgwalters:
I'm going to split this into separate PRs.  First one: https://github.com/openshift/openshift-ansible-contrib/pull/122

vagrant/Vagrantfile
cgwalters: Ah, so it turns out I broke this by moving the sync path to `..` instead of `.`.  Not entirely sure now why I did that.  Looking :eyes:

cgwalters: Okay, trying to go back to syncing `.` instead of `..`, I get:

```
[vagrant@admin1 ~]$ cd /home/vagrant/sync && PYTHONUNBUFFERED=1 ANSIBLE_FORCE_COLOR=true ansible-playbook --limit="OSEv3:localhost" --inventory-file=/tmp/vagrant-ansible/inventory -v /home/vagrant/sync/install.yaml
Using /home/vagrant/sync/ansible.cfg as config file

ERROR! the role 'rhsm-subscription' was not found in /home/vagrant/sync/roles:/etc/ansible/roles:/home/vagrant/sync

The error appears to have been in '/home/vagrant/sync/install.yaml': line 18, column 5, but may
be elsewhere in the file depending on the exact syntax problem.

The offending line appears to be:

  roles:
  - role: rhsm-subscription
    ^ here

[vagrant@admin1 sync]$ 
[vagrant@admin1 sync]$ ls
ansible.cfg  install.retry  install.yaml  README.md  roles  tasks  Vagrantfile
[vagrant@admin1 sync]$ ls -al roles/
total 16
drwxrwxr-x. 1 vagrant vagrant  49 Nov  8 21:53 .
drwxrwxr-x. 1 vagrant vagrant 144 Nov 21 16:23 ..
lrwxrwxrwx. 1 vagrant vagrant  67 Nov  8 21:53 rhsm-repos -> ../../reference-architecture/aws-ansible/playbooks/roles/rhsm-repos
lrwxrwxrwx. 1 vagrant vagrant  74 Nov  8 21:53 rhsm-subscription -> ../../reference-architecture/aws-ansible/playbooks/roles/rhsm-subscription
[vagrant@admin1 sync]$ 
```

Did this work before?

detiber: I want to say that it did work because the rsync plugin follows symlinks by default. I wonder if this is a behavior difference between rsync and sshfs plugins.

cgwalters:
Obsoleted by https://github.com/openshift/openshift-ansible-contrib/pull/132

PR #105, opened 2016-11-11
print variable correctly

cooktheryan:
I discovered that the variable was not getting printed correctly in the Ansible plays. Only the first word was being used.

For example 

"{{ rhsm_pool }}" would print Employee

Now corrected
"{{ rhsm_pool }}" Employee SKU


cooktheryan:
[BUG: Subscription manager play is not expanding the variable correctly.](https://trello.com/c/PLpe8khg/492-bug-subscription-manager-play-is-not-expanding-the-variable-correctly)

Issue #83, opened 2016-10-26
Deploy stucks on rhsm-subscription


jaryn:
Hi
My deployments are sometimes getting stuck quite a while on the RHN registration. After two hours being "stuck" it failed:

```
PLAY [localhost] ***************************************************************

TASK [instance-groups : Add bastion to group] **********************************
changed: [localhost]

TASK [instance-groups : Add masters to requisite groups] ***********************
changed: [localhost] => (item=ip-10-20-4-101.us-west-2.compute.internal)
changed: [localhost] => (item=ip-10-20-5-157.us-west-2.compute.internal)
changed: [localhost] => (item=ip-10-20-6-26.us-west-2.compute.internal)

TASK [instance-groups : Add a master to the primary masters group] *************
changed: [localhost] => (item=ip-10-20-4-101.us-west-2.compute.internal)

TASK [instance-groups : Add infra instances to host group] *********************
changed: [localhost] => (item=ip-10-20-5-61.us-west-2.compute.internal)
changed: [localhost] => (item=ip-10-20-4-22.us-west-2.compute.internal)

TASK [instance-groups : Add app instances to host group] ***********************
changed: [localhost] => (item=ip-10-20-5-170.us-west-2.compute.internal)
changed: [localhost] => (item=ip-10-20-4-177.us-west-2.compute.internal)

TASK [instance-groups : Add app instances to host group] ***********************

PLAY [localhost] ***************************************************************

TASK [host-up : check to see if host is available] *****************************
ok: [localhost]

PLAY [cluster_hosts] ***********************************************************

TASK [setup] *******************************************************************
ok: [ose-master01....com]

TASK [rhsm-subscription : Register host] ***************************************
changed: [ose-master01....com]

PLAY [cluster_hosts] ***********************************************************

TASK [setup] *******************************************************************
ok: [ose-master02....com]

TASK [rhsm-subscription : Register host] ***************************************

fatal: [ose-master02....com]: UNREACHABLE! => {
    "changed": false, 
    "unreachable": true
}

MSG:

Failed to connect to the host via ssh: Shared connection to ose-master02....com closed.


PLAY RECAP *********************************************************************
localhost                  : ok=6    changed=5    unreachable=0    failed=0   
ose-master01....com : ok=2    changed=1    unreachable=0    failed=0   
ose-master02....com : ok=1    changed=0    unreachable=1    failed=0   
```

[...]

PR #106, opened 2016-11-13
added timeout configuration variable to help assist installation

cooktheryan:
This will allow the host-up delay check to be configurable. This will help alleviate any slow dns propagation issues

This resolved a slow DNS issue we had with the CI system


cooktheryan:
resolves #83

PR #107, opened 2016-11-14
Bugfix due to an rpm dependency

cooktheryan:
Error occurs due to https://bugzilla.redhat.com/show_bug.cgi?id=1394491


detiber:
@cooktheryan has that bug shipped with a released version of OpenShift?

cooktheryan:
@detiber upstream issue but for origin need to have this in place to support

detiber:
@cooktheryan origin should probably be using the specific branches related to the latest release, so for 1.3, it should probably be using the release-1.3 branch for openshift-ansible.

Master should really only be used for 1.4/3.4 testing.

cooktheryan:
@detiber hmm. I wonder if we should cancel this then. Only reason why this is in place I ran into an issue with 
Destination /etc/origin/cloudprovider/aws.conf does not exist ! and using the fix from upstream resolved the issue but introduced this. I know the aws.conf error will go away once the updated rpms get pushed today.

Leaning towards close now

detiber:
@cooktheryan I would agree on closing as well. I'd like to avoid shipping workarounds if at all possible.

PR #108, opened 2016-11-14
Rollback of features that didn't make the release yet

cooktheryan:


dav1x:
LGTM offload the variables from setup into the jinja template.

PR #109, opened 2016-11-14
merge when version of playbooks > .41

cooktheryan:
Waiting for new rpm playbooks to GA. Previously accepted but had to role back as rpm code wasn't ready yet


cooktheryan:
.50 appears to have been released today of openshift-ansible-playbooks/roles/etc

PR #112, opened 2016-11-16
Minor tweaks and fixes for GCE

pschiffe:
This PR contains couple of minor tweaks and fixes for GCE code, being still backward compatible with published ref. arch.


pschiffe:
Well, depends on when you will have time for the paper, because as it's now, it's not possible to deploy ocp on gce with the current code (because of the ansible 2.2).

themurph:
If the install is broken then we should merge now, the paper should be updated early next week.

PR #113, opened 2016-11-16
Add script fragments for manual install of OCE on OSP

markllama:
This PR adds a set of scripts to demonstrate manual preparation and installation of OpenShift on OpenStack.  These scripts mirror the fragments in the OCP on OSP reference architecture document.

https://access.redhat.com/articles/2743631

### DIFF
--- a/reference-architecture/osp-cli/README.adoc
+++ b/reference-architecture/osp-cli/README.adoc
@@ -1,0 +1,153 @@
+== Scripted Installation of OpenShift on OpenStack.
+
+The scripts presented here match those in https://access.redhat.com/articles/2743631[Red Hat OpenShift Container Platform 3 on Red Hat OpenStack Platform 8] reference architecture document.  These script fragments are intended to demonstrate the steps needed to deploy OCP on OSP, but they are not the recommended method for doing so.  Production deployment should be done using the https://github.com/openshift/openshift-on-openstack[OpenShift on OpenStack] Heat templates and the https://github.com/openshift/openshift-ansible[OpenShift Ansible] playbooks.
+
+The documentation here does not provide comprehensive usage
+instructions for these script fragments.  The scripts are meant to be
+used in conjunction with the reference architecture document above.
+
+These scripts are offered as-is, and no warranty is implied or offered.
+
+=== Prerequisites
+
+To use these scripts require a DNS server running BIND.
+It must have a delegated sub-domain for the OpenShift service.  In
+these examples that subdomain is `osp3.example.com`. Adjust this for
+the target environment.  The subdomain must allow dynamic updates using
+`nsupdate`. The process requires a copy of the `rndc.key`
+from the BIND server to allow updates.
+
+The scripts expect a pre-existing `haproxy` proxy server. The user
+must have a key-based SSH passwordless access to the proxy server.
+The user account on the proxy server must have the ability to execute
+commands as root via `sudo`
+
+The `MASTER_DNS_NAME` and `APPS_DNS_SUFFIX` values below must each be
+a domain name that refers to the IP address of the haproxy server.
+
+The `APP_DNS_SUFFIX` domain name is a _wildcard record_.  The hostname
+portion is an asterisk (*).  This will match any name in that
+subdomain.
+
+.Proxy DNS Records 
+----
+proxy.ocp3.example.com.	300 IN	A	10.19.XX.XX
+devs.ocp3.example.com. 300 IN	A	10.19.XX.XX
+*.apps.ocp3.example.com.	300 IN	A	10.19.XX.XX
+----
+
+A RHEL 7 image must be provisioned into Glance.  Note the
+`GLANCE_IMAGE` name below.
+
+There must be an LDAP or Active Directory server available.  There
+must be a user and password that can bind to the LDAP server to enable
+authentication.
+
+=== Deployment Parameters
+
+These scripts make use of a set of enviroment variables to control the
+deployment process. The values listed below must be set in the
+caller's environment before invoking the first script.  You can write
+them into a file and then `source` them into the environment.
+
+.Define Installation Variables
+----
+# Software Version Override
+export GLANCE_IMAGE=rhel73
+export OCP_VERSION=3.3
+export OSP_VERSION=8
+
+# Infrastructure
+export NAMESERVER=10.19.X.X
+export OCP3_DOMAIN=ocp3.example.com
+
+# Generate DNS
+export OCP3_DNS_SERVER=10.19.X.X
+export OCP3_DNS_UPDATE_KEY=/home/testuser/keys/ns1.ose3.key
+
+# Generate haproxy
+export HAPROXY_HOST=ocp3-proxy.example.com
+export HAPROXY_KEY=/home/testuser/keys/services_rsa
+
+# Bastion access
+export BASTION_HOST=bastion.ocp3
+export BASTION_KEY=/home/testuser/keys/ocp3_rsa
+
+export MASTER_DNS_NAME=devs.ocp3.example.com
+export APPS_DNS_SUFFIX=apps.ocp3.example.com
+
+export LDAP_BIND_DN=cn=openshift,cn=users,dc=example,dc=com
+export LDAP_BIND_PASSWORD=password
+export LDAP_URL=ldap://ad1.example.com:389/cn=users,dc=example,dc=com?sAMAccountName
+----
+
+
+=== Red Hat Software Credentials
+
+This procedure requires a valid set of Red Hat Access credentials and a
+subscription pool which contains the OCP and OSP repositories for the
+selected versions.
+
+Each host must be registered to receive software from the Red Hat RPM
+repositories. 
+
+* `rhel-7-server-ose-${OCP_VERSION}-rpms`
+* `rhel-7-server-openstack-${OSP_VERSION}-rpms`
+* `rhel-7-server-openstack-${OSP_VERSION}-director-rpms`
+
+.rnh_credentials
+----
+export RHN_USERNAME=<username>
+export RHN_PASSWORD=<password>
+export RHN_POOL_ID=<pool id>
+export OCP_VERSION=3.3
+# Uncomment to use development version (unsupported)
+#export OCP_ANSIBLE_GIT_REPO=http://github.com/openshift/openshift-ansible.git
+#export OCP_ANSIBLE_ROOT=/home/cloud-user
+----
+
+=== Creating the Infrastructure
+
+The first script creates the network and VMs that will make up the
+OpenShift cluster.
+
+.Create OSP Infrastructure
+----
+sh ./infrastructure.sh
+----
+
+This creates the bastion host, master, infrastructure node and app
+node VMs. This includes updating the DNS and haproxy configurations.
+
+//.Update External Infrastructure Configuration
+//sh ./generate_dns_updates.sh
+//sh ./generate_haproxy_conf.sh
+//----
+
+=== Prepare the host instances
+
+This step initializes and customizes the instances for OpenShift
+installation via Ansible.
+
+First it configures the bastion host; registering for software repos,
+installing Ansible and other tools.  Then it repeats the process with
+all of the OpenShift VMs from the bastion host.
+
+.Stage Bastion and VM Configuration Scripts
+----
+sh ./prepare_bastion.sh
+sh ./prepare_instances.sh
+----
+
+The final step is to run the OpenShift Ansible installer.  This script
+generates the Ansible input files, pushes them to the bastion host and
+executes the installation process.
+
+.Install and Configure OpenShift
+----
+sh ./install_openshift.sh
+----
+
+
+When this is complete and sucessful users will be able to browse to
+the OpenShift service masters through the load balancer.
--- a/reference-architecture/osp-cli/ansible.cfg
+++ b/reference-architecture/osp-cli/ansible.cfg
@@ -1,0 +1,19 @@
+# config file for ansible -- http://ansible.com/
+# ==============================================
+[defaults]
+forks = 50
+host_key_checking = False
+remote_user = cloud-user
+gathering = smart
+retry_files_enabled = false
+nocows = true
+#lookup_plugins = ./playbooks/lookup_plugins
+#log_path = /tmp/ansible.log
+
+[privilege_escalation]
+become = True
+
+[ssh_connection]
+ssh_args = -o ControlMaster=auto -o ControlPersist=900s -o GSSAPIAuthentication=no
+control_path = /var/tmp/%%h-%%r
+pipelining = True
--- a/reference-architecture/osp-cli/ch4.8_copy_ssh_key.sh
+++ b/reference-architecture/osp-cli/ch4.8_copy_ssh_key.sh
@@ -1,0 +1,6 @@
+#!/bin/sh
+OCP3_KEY_FILE=${OCP3_KEY_FILE:-keys/ocp3_rsa}
+BASTION_HOST=${BASTION_HOST:-bastion.${OCP3_BASE_DOMAIN}}
+scp -i ${OCP3_KEY_FILE} ${OCP3_KEY_FILE} cloud-user@${BASTION_HOST}:.ssh/id_rsa
+
+ssh -i ${OCP3_KEY_FILE} cloud-user@${BASTION_HOST} chmod 600 .ssh/id_rsa
--- a/reference-architecture/osp-cli/ch4.9_allow_docker_flannel.sh
+++ b/reference-architecture/osp-cli/ch4.9_allow_docker_flannel.sh
@@ -1,0 +1,4 @@
+#ansible masters -i inventory -m shell \
+#  -a 'iptables -A DOCKER -p tcp -j ACCEPT'
+ansible nodes -i inventory -m shell \
+  -a 'iptables -A DOCKER -p tcp -j ACCEPT'
--- a/reference-architecture/osp-cli/haproxy.cfg.template
+++ b/reference-architecture/osp-cli/haproxy.cfg.template
@@ -1,0 +1,63 @@
+global
+    log         127.0.0.1 local2
+
+    chroot      /var/lib/haproxy
+    pidfile     /var/run/haproxy.pid
+    maxconn     4000
+    user        haproxy
+    group       haproxy
+    daemon
+    # turn on stats unix socket
+    stats socket /var/lib/haproxy/stats
+
+defaults
+    log                     global
+    option                  httplog
+    option                  dontlognull
+    option http-server-close
+    option                  redispatch
+    retries                 3
+    timeout http-request    10s
+    timeout queue           1m
+    timeout connect         10s
+    timeout client          1m
+    timeout server          1m
+    timeout http-keep-alive 10s
+    timeout check           10s
+    maxconn                 3000
+listen  stats :9000
+    	stats enable
+    	stats realm Haproxy\ Statistics
+        stats uri /haproxy_stats
+        stats auth admin:password 
+		stats refresh 30
+		mode http
+
+frontend  main *:80
+    default_backend             router80
+
+backend router80
+    balance source
+    mode tcp
+    # INFRA_80
+    #server infra-node-0.%OCP3_DOMAIN% %INFRA_0_IP%:80 check
+
+
+frontend  main *:443
+    default_backend             router443
+
+backend router443
+    balance source
+    mode tcp
+    # INFRA_443
+    #server infra-node-0.%OCP3_DOMAIN% %INFRA_0_IP%:443 check
+
+
+frontend  main *:8443
+    default_backend             mgmt8443
+
+backend mgmt8443
+    balance source
+    mode tcp
+    # MASTERS
+    #server master-0.%OCP3_DOMAIN% %MASTER_0_IP%:8443 check
--- a/reference-architecture/osp-cli/inventory.template
+++ b/reference-architecture/osp-cli/inventory.template
@@ -1,0 +1,36 @@
+[OSEv3:children]
+infra
+masters
+nodes
+etcd
+
+[infra]
+localhost
+
+[masters]
+master-0.control.ocp3.example.com openshift_hostname=master-0.control.ocp3.example.com openshift_public_hostname=master-0.control.ocp3.example.com
+master-1.control.ocp3.example.com openshift_hostname=master-1.control.ocp3.example.com openshift_public_hostname=master-1.control.ocp3.example.com
+master-2.control.ocp3.example.com openshift_hostname=master-2.control.ocp3.example.com openshift_public_hostname=master-2.control.ocp3.example.com
+
+[masters:vars]
+openshift_schedulable=true
+openshift_router_selector="region=infra"
+openshift_registry_selector="region=infra"
+
+[etcd]
+master-0.control.ocp3.example.com openshift_hostname=master-0.control.ocp3.example.com openshift_public_hostname=master-0.control.ocp3.example.com
+master-1.control.ocp3.example.com openshift_hostname=master-1.control.ocp3.example.com openshift_public_hostname=master-1.control.ocp3.example.com
+master-2.control.ocp3.example.com openshift_hostname=master-2.control.ocp3.example.com openshift_public_hostname=master-2.control.ocp3.example.com
+
+[nodes]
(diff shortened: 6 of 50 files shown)


reference-architecture/osp-cli/ch4.3_infra_node_security_group.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/sh
+neutron security-group-create infra-node-sg
+neutron security-group-rule-create infra-node-sg --protocol icmp
+neutron security-group-rule-create infra-node-sg \
+        --protocol tcp --port-range-min 22 --port-range-max 22 \
+        --remote-group-id bastion-sg
+
+for PORT in 80 443 5000 10250
```
cooktheryan: Is external 5000 required to be open to security group?

reference-architecture/osp-cli/ch4.4_boot_bastion.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+OCP3_DOMAIN=${OCP3_DOMAIN:-ocp3.example.com}
+nova boot --flavor m1.small --image rhel72 --key-name ocp3 \
```
cooktheryan: It looks like above you are creating the glance image of rhel73

reference-architecture/osp-cli/ch4.4_boot_infra_nodes.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+OCP3_DOMAIN=${OCP3_DOMAIN:-ocp3.example.com}
+for HOSTNAME in infra-node-0 infra-node-1
+do
+  VOLUMEID=$(cinder show ${HOSTNAME}-docker | grep ' id ' | awk '{print $4}')
+
+  nova boot --flavor m1.medium --image rhel72 --key-name ocp3 \
```
cooktheryan: Same as above with image name

reference-architecture/osp-cli/ch4.4_boot_masters.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+OCP3_DOMAIN=${OCP3_DOMAIN:-ocp3.example.com}
+for HOSTNUM in 0 1 2 ; do
+  nova boot --flavor m1.small --image rhel72 --key-name ocp3 \
```
cooktheryan: Same as above with image name

reference-architecture/osp-cli/ch4.4_security_groups.sh
```diff
@@ -0,0 +1,129 @@
...
+#   Port/Proto  From      Reason
+#
+#  22/TCP      bastion   Secure Shell
+#  8443/TCP    lb        WebUI/REST
+#  8443/TCP    master    WebUI/REST     ??
+#  8443/TCP    node      WebUI/REST     ??
+#  8053/TCP    node      Name Service   ??
+#  8053/UDP    node      Name Service   ??
+neutron security-group-rule-create master-sg --protocol icmp
+for PORT in 22 53 2379 4001 4789 8053 8443 10250 24224 ; do
```
cooktheryan: As long as its multi-master only for the ref arch 4001 can be removed

reference-architecture/osp-cli/ch4.3_master_security_group.sh
```diff
@@ -0,0 +1,25 @@
...
+neutron security-group-rule-create master-sg --protocol icmp
+neutron security-group-rule-create master-sg \
+     --protocol tcp --port-range-min 22 --port-range-max 22 \
+     --remote-group-id bastion-sg
+
+neutron security-group-rule-create master-sg \
+     --protocol tcp --port-range-min 2380 --port-range-max 2380 \
+     --remote-group-id master-sg
+
+for PORT in 53 2379 2380 4001 8053 8443 10250 24224
```
cooktheryan: As long as ref arch is only supporting multi master 4001 can be removed

reference-architecture/osp-cli/ch4.4_security_groups.sh
```diff
@@ -0,0 +1,129 @@
...
+#     --protocol tcp --port-range-min 22 --port-range-max 22 \
+#     --remote-group-id bastion-sg
+#neutron security-group-rule-create node-sg \
+#     --protocol tcp --port-range-min 10250 --port-range-max 10250 \
+#     --remote-group-id master-sg
+#neutron security-group-rule-create node-sg \
+#     --protocol tcp --port-range-min 4789 --port-range-max 4789 \
+#     --remote-group-id node-sg
+neutron security-group-rule-create node-sg --protocol icmp
+for PORT in 22 10250 ; do
```
cooktheryan: Do we all 22 from 0.0.0.0 or only from the bastion SG? If from 0.0.0.0/0 we should look to restrict to bastion only in the future to gain the full advantages of using a bastion

reference-architecture/osp-cli/inventory.template
```diff
@@ -0,0 +1,35 @@
...
+localhost
+
+[masters]
+master-0.control.ocp3.example.com openshift_hostname=master-0.control.ocp3.example.com openshift_public_hostname=master-0.control.ocp3.example.com
+master-1.control.ocp3.example.com openshift_hostname=master-1.control.ocp3.example.com openshift_public_hostname=master-1.control.ocp3.example.com
+master-2.control.ocp3.example.com openshift_hostname=master-2.control.ocp3.example.com openshift_public_hostname=master-2.control.ocp3.example.com
+
+[masters:vars]
+openshift_schedulable=true
+openshift_router_selector="region=infra"
```
cooktheryan: it may be worth pinning the registry to infra as well

reference-architecture/osp-cli/haproxy.cfg.template
```diff
@@ -0,0 +1,63 @@
...
+    default_backend             router443
+
+backend router443
+    balance source
+    mode tcp
+    # INFRA_443
+    #server infra-node-0.%OCP3_DOMAIN% %INFRA_0_IP%:443 check
+
+
+frontend  main *:8443
```
cooktheryan: Not required but in the inventory the port could be set to 443 but that's more of a nit

openshift_master_api_port: 443
openshift_master_console_port: 443

reference-architecture/osp-cli/ch4.8.4_install_docker_nodes.sh
```diff
@@ -0,0 +1,4 @@
+for H in $NODES
+do
+  ssh $H sudo yum -y install docker
```
cooktheryan: You should be able to remove this as the installer should take care of installing

Worth testing

reference-architecture/osp-cli/ch4.8.4_enable_docker_nodes.sh
```diff
@@ -0,0 +1,5 @@
+for H in $NODES
+do
+  ssh $H sudo systemctl enable docker
```
cooktheryan: You should be able to remove this as the installer should take care of starting and enabling

reference-architecture/osp-cli/ch4.4_security_groups.sh
detiber: I'm not sure that 8053 needs to be exposed on the master (I could be wrong, though) @sdodson can you confirm?

reference-architecture/aws-ansible/playbooks/roles/instance-groups/tasks/main.yaml
```diff
@@ -18,13 +18,11 @@
...
   add_host:
-    name: "{{ hostvars[item].ec2_tag_Name }}"
+    name: "{{ item }}"
     groups: primary_master
     openshift_node_labels:
       role: master
-  with_items: "{{ groups['tag_openshift_role_master'].0 }}"
-  when:
-    - hostvars[item]['ec2_tag_aws_cloudformation_stack_name'] == "{{ stack_name }}"
+  with_items: "{{ groups['masters'].0 }}"
```
detiber: instead of `groups['masters'].0` you can use `groups.masters.0`

reference-architecture/aws-ansible/playbooks/validation.yaml
```diff
@@ -3,36 +3,15 @@
   become: no
   vars_files:
       - vars/main.yaml
+  pre_tasks:
+      - name: set fact
+        set_fact:
+           openshift_master_cluster_public_hostname: "{{ openshift_master_cluster_public_hostname }}"
+      - name: set fact
+        set_fact:
+           openshift_master_cluster_hostname: "{{ openshift_master_cluster_hostname }}"
```
detiber: This isn't guaranteed to work, since you are setting the same variable name that you are using.

reference-architecture/osp-cli/OSv3.yml.template
```diff
@@ -0,0 +1,36 @@
+deployment_type: openshift-enterprise
```
detiber: It looks like this is used in group_vars, if that is the case, it should be named `OSEv3.yml` to match the byo groups.

reference-architecture/osp-cli/all-hosts.sh
```diff
@@ -0,0 +1,93 @@
+#!/bin/sh
+#
+# Execute these preparation steps on all hosts
+# Run on the bastion host
+#
+OCP_VERSION=${OCP_VERSION:-3.2}
```
detiber: Should this default to 3.3, since it is the latest released?

reference-architecture/osp-cli/ch4.4_user_data.sh
```diff
@@ -0,0 +1,46 @@
...
+fqdn: $1.$2
+
+--===============6355019966770068462==
+MIME-Version: 1.0
+Content-Type: text/x-shellscript; charset="us-ascii"
+Content-Transfer-Encoding: 7bit
+Content-Disposition: attachment; filename="allow-sudo-ssh.sh"
+
+#!/bin/sh
+sed -i "/requiretty/s/^/#/" /etc/sudoers
```
detiber: Instead of this, why not use a drop file: https://github.com/openshift/openshift-ansible/blob/master/playbooks/aws/openshift-cluster/templates/user_data.j2#L18-L21

reference-architecture/osp-cli/ch4.4_security_groups.sh
detiber: As long as the masters also run etcd (assuming multi-master) 2380 will need to be open as well.

reference-architecture/osp-cli/ch4.8_copy_ssh_key.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+OCP3_KEY_FILE=${OCP3_KEY_FILE:-keys/ocp3_rsa}
+BASTION_HOST=${BASTION_HOST:-bastion.${OCP3_BASE_DOMAIN}}
+scp -i ${OCP3_KEY_FILE} ${OCP3_KEY_FILE} cloud-user@${BASTION_HOST}:.ssh/id_rsa
+
+ssh -i ${OCP3_KEY_FILE} cloud-user@${BASTION_HOST} chmod 600 .ssh/id_rsa
```
detiber: can `ssh-copy-id` be used here instead?

reference-architecture/osp-cli/ch4.9_allow_docker_flannel.sh
```diff
@@ -0,0 +1,4 @@
+#ansible masters -i inventory -m shell \
+#  -a 'iptables -A DOCKER -p tcp -j ACCEPT'
+ansible nodes -i inventory -m shell \
+  -a 'iptables -A DOCKER -p tcp -j ACCEPT'
```
detiber: This seems like a bug if you need to add an iptables rule for flannel to work.

reference-architecture/osp-cli/docker-storage-setup
```diff
@@ -0,0 +1,2 @@
+DEVS=/dev/vdb
+VG=docker-vg
```
detiber: This seems like a good candidate for putting in cloud-init

reference-architecture/osp-cli/ifcfg-eth1
```diff
@@ -0,0 +1,10 @@
+# /etc/sysconfig/network-scripts/ifcfg-eth1
+DEVICE="eth1"
+BOOTPROTO="dhcp"
+BOOTPROTOv6="none"
+ONBOOT="yes"
+TYPE="Ethernet"
+USERCTL="no"
+PEERDNS="no"
+IPV6INIT="no"
+PERSISTENT_DHCLIENT="1"
```
detiber: This also seems like a good candidate for cloud-init

reference-architecture/aws-ansible/playbooks/validation.yaml
cooktheryan: will update in separate PR

reference-architecture/aws-ansible/playbooks/roles/instance-groups/tasks/main.yaml
cooktheryan: will update in separate PR

reference-architecture/osp-cli/ch4.4_boot_app_nodes.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+OCP3_DOMAIN=${OCP3_DOMAIN:-ocp3.example.com}
+for HOSTNAME in app-node-0 app-node-1 app-node-2
+do
+  VOLUMEID=$(cinder show ${HOSTNAME}-docker | grep ' id ' | awk '{print $4}')
+  
+  nova boot --flavor m1.medium --image rhel72 --key-name ocp3 \
```
markllama: Oops, missed a variableization

reference-architecture/osp-cli/ch4.4_boot_bastion.sh
markllama: More missed variableization.

reference-architecture/osp-cli/ch4.4_boot_masters.sh
markllama: Yep

reference-architecture/osp-cli/ch4.4_security_groups.sh
markllama: I will check again but I believe these all match the refarch and the heat templates.

reference-architecture/osp-cli/ch4.3_master_security_group.sh
markllama: See above

reference-architecture/osp-cli/ch4.4_security_groups.sh
markllama: Yep, that one should be to the bastion-sg only.

reference-architecture/osp-cli/haproxy.cfg.template
markllama: That would conflict with the app inbounds on port 443, would it not? With a single TCP haproxy, the app and master connections inbound must be on different ports to be forwarded to different hosts.

reference-architecture/osp-cli/ch4.8.4_install_docker_nodes.sh
markllama: I thought so too, but yes.

markllama: Oh wait.  No this is because it is configuring docker storage on the nodes.  Docker must be running after configuring the storage and before running docker-storage-config.

reference-architecture/osp-cli/ch4.8.4_enable_docker_nodes.sh
markllama: Enable could be left for the installer but since it was already installed and configured, not enabling it would raise questions.

reference-architecture/osp-cli/ch4.8.4_configure_docker_storage_nodes.sh
```diff
@@ -0,0 +1,6 @@
+for H in $NODES
+do
+  scp docker-storage-setup ${H}:
+  ssh $H sudo cp docker-storage-setup /etc/sysconfig
+  ssh $H sudo /usr/bin/docker-storage-setup
```
markllama: I was under the assumption that it would not, as the heat templates do it explicitly.

reference-architecture/aws-ansible/playbooks/validation.yaml
markllama: CHecking for missed rebase.

reference-architecture/osp-cli/OSv3.yml.template
markllama: Will update.  Matching the heat templates. Should it not now be changed universally to OCPv3.yml (gag)

reference-architecture/osp-cli/all-hosts.sh
markllama: Currently defaults to 3.2 since that was the latest tested on OSP8 and reviewed in the refarch.  Will update with the next refarch release.

reference-architecture/osp-cli/haproxy.cfg.template
detiber: That is true, would need to bind to separate interfaces, since tcp passthrough is being used for this to work.

reference-architecture/osp-cli/ch4.8_copy_ssh_key.sh
markllama: perhaps.  Will try.

reference-architecture/osp-cli/ch4.8.4_install_docker_nodes.sh
detiber: Actually, docker has a dependency on docker-storage-setup, so it will automatically run docker-storage-setup

reference-architecture/osp-cli/ch4.8.4_enable_docker_nodes.sh
detiber: Just dropping the docker-storage-setup config should be sufficient.

reference-architecture/osp-cli/ch4.8.4_configure_docker_storage_nodes.sh
detiber: There was previously a bug where docker-storage-setup would take too long and docker would fail to start, but I believe that is no longer an issue.

reference-architecture/osp-cli/docker-storage-setup
markllama: I agree, but to use cloud-init well in the refarch Id have had to also pull in core-os/cloud-init/tools/make-mime.py.  I can't find a RHEL package with make-mime.py and I didn't want to make the cloud-init script a gigantic mess with two or three replacements.

Several other things can go in cloud-init if we can dynamically create the host-data as a multi-part MIME file in a simple way.

reference-architecture/osp-cli/ifcfg-eth1
markllama: see above

reference-architecture/osp-cli/OSv3.yml.template
detiber: I think at this point we want to avoid breaking existing users/docs, so I have not heard any requests to rename from OSEv3 (yet).

markllama:
Rebased properly, reduced the changed files to those in osp-cli.

reference-architecture/osp-cli/docker-storage-setup
detiber: You s houldn't need make-mime, you should be able to do it through cloud-config using write_files (example here: https://github.com/openshift/openshift-ansible/blob/master/playbooks/aws/openshift-cluster/templates/user_data.j2)

reference-architecture/osp-cli/ch4.8.4_install_docker_nodes.sh
markllama: I'll try that then.

reference-architecture/osp-cli/ch4.4_user_data.sh
markllama: Because I have to do this *before* ansible can run.

reference-architecture/osp-cli/ch4.8_copy_ssh_key.sh
markllama: Actually no.  ssh-copy-id copies a local public key to a remote `authorized_keys` file.  That allows login from the local host to the remote one.

In this case I'm copying the _private_ key so that the bastion can log into the other instances for Ansible.

reference-architecture/osp-cli/docker-storage-setup
markllama: I am not using Ansible (yet) to do anything but the openshift installation.

The point isn't to get the job done, it's to expose what's happening so that deployers have a chance to understand what can go wrong and what it might look like.

In the next revision the refarch authors for OCP on cloud are looking at replacing the python and bourne shell we're using now with more ansible, but only if it doesn't obscure the underlying change.

cooktheryan: What Jason posted isn't really Ansible

reference-architecture/osp-cli/ansible.cfg
```diff
@@ -0,0 +1,19 @@
...
+#lookup_plugins = ./playbooks/lookup_plugins
+#log_path = /tmp/ansible.log
+
+[privilege_escalation]
+become = True
+
+[ssh_connection]
+ssh_args = -o ControlMaster=auto -o ControlPersist=900s -o GSSAPIAuthentication=no
+control_path = /var/tmp/%%h-%%r
+#pipelining = True
```
markllama: Trying

reference-architecture/osp-cli/inventory.template
markllama: Added

markllama:
All comments addressed or fixed.  Time to merge?

reference-architecture/osp-cli/ch4.4_user_data.sh
markllama: Correction: I do it this way so the size of the mime file is smaller.

markllama: Correction: I learn something new about how sudo works.  I'll try that.

reference-architecture/osp-cli/ch3_generate_keypair.sh
```diff
@@ -0,0 +1,18 @@
+ssh-keygen -f ocp3_rsa -N "" -C "OpenShift KeyPair"
+Generating public/private rsa key pair.
```
detiber: Lines 2-18 appear to be the output from the command in line 1

markllama: They are.  this is a bad hold over from the refarch doc sample, but not used in the scripts.
Removing.  *sigh*  Time to fine comb and not depend on you as my beta testers.

markllama: removed.

reference-architecture/osp-cli/README.adoc
```diff
@@ -0,0 +1,152 @@
...
+=== Prerequisites
+
+To use these scripts require a DNS server running BIND.
+It must have a delegated sub-domain for the OpenShift service.  In
+these examples that subdomain is `osp3.example.com`. Adjust this for
+the target environment.  The subdomain must allow dynamic updates using
+`nsupdate`. The process requires a copy of the `rndc.key`
+from the BIND server to allow updates.
+
+The scripts require an _haproxy_ proxy server configured. The user
```
themurph: "The scripts require an _haproxy_ proxy server configured" 

This sentence is strangely worded. I'm not sure if this means that an HAProxy server should already be provisioned or if the scripts are going to provision an HAProxy server.

reference-architecture/osp-cli/README.adoc
```diff
@@ -0,0 +1,152 @@
...
+must have an SSH key to allow access to the proxy server and the user
```
themurph: "must have **an SSH key** to allow access"

I know what you mean, but something like "must have Key-Based SSH passwordless access"

reference-architecture/osp-cli/README.adoc
```diff
@@ -0,0 +1,152 @@
...
+all of the OpenShift VMs from the bastion host.
+
+.Stage Bastion and VM Configuration Scripts
+----
+sh ./prepare_bastion.sh
+sh ./prepare_instances.sh
+----
+
+The final step is to run the OpenShift Ansible installer.  This script
+generates the ansible input files, pushes them to the bastion host and
```
themurph: Spelling: ansible -> Ansible

PR #115, opened 2016-11-16
Azure PR

cooktheryan:


reference-architecture/azure-ansible/azuredeploy.json
```diff
@@ -0,0 +1,1048 @@
...
+    },
+    "SubscriptionPoolId" : {
+      "type" : "string",
+      "minLength" : 1,
+      "metadata" : {
+        "description" : "Pool ID of the redhat subscritpion to use"
+      }
+    },
+    "sshPrivateData" : {
+      "type" : "string",
```
detiber: Should this be `securestring`?

reference-architecture/azure-ansible/azuredeploy.json
```diff
@@ -0,0 +1,1048 @@
...
+      "metadata" : {
+        "description" : "SSH rsa private key file as a base64 string."
+      }
+    },
+    "storeVMSize" : {
+      "type" : "string",
+      "defaultValue" : "Standard_DS3_v2",
+      "allowedValues" : [
+        "Standard_DS3_v2",
+        "Standard_DS12\t"
```
detiber: Any reason to limit this host to just these options? Other hosts seem to offer more VM size options.

reference-architecture/azure-ansible/azuredeploy.json
```diff
@@ -0,0 +1,1048 @@
...
+    "infranodeStorageName" : "[concat(uniqueString(resourceGroup().id), 'infranodesa')]",
+    "nodeStorageName" : "[concat(uniqueString(resourceGroup().id), 'nodesa')]",
+    "masterStorageName" : "[concat(uniqueString(resourceGroup().id), 'mastersa')]",
+    "vhdStorageType" : "Standard_LRS",
+    "vnetId" : "[resourceId('Microsoft.Network/virtualNetworks', variables('virtualNetworkName'))]",
+    "infranodeSubnetRef" : "[concat(variables('vnetId'), '/subnets/', variables('infranodesubnetName'))]",
+    "nodeSubnetRef" : "[concat(variables('vnetId'), '/subnets/', variables('nodesubnetName'))]",
+    "masterSubnetRef" : "[concat(variables('vnetId'), '/subnets/', variables('mastersubnetName'))]",
+    "centos" : {
+      "publisher" : "Openlogic",
```
detiber: Is there a more official CentOS image that can be used?

reference-architecture/azure-ansible/azuredeploy.json
```diff
@@ -0,0 +1,1048 @@
...
+            "name" : "[variables('mastersubnetName')]",
+            "properties" : {
+              "addressPrefix" : "[variables('mastersubnetPrefix')]"
+            }
+          }
+        ]
+      }
+    },
+    {
+      "name" : "logging",
```
detiber: What is the logging host used for?

reference-architecture/azure-ansible/azuredeploy.json
```diff
@@ -0,0 +1,1048 @@
...
+            "value" : "[parameters('SubscriptionPoolId')]"
+          },
+          "sshPrivateData" : {
+            "value" : "[parameters('sshPrivateData')]"
+          }
+        }
+      }
+    },
+    {
+      "name" : "infranode1",
```
detiber: Any reason not to make this more similar to nodeSet? I can see plenty of reasons one would want to scale the "infra" nodes.

reference-architecture/azure-ansible/azuredeploy.json
```diff
@@ -0,0 +1,1048 @@
...
+      "name" : "bastion",
+      "type" : "Microsoft.Resources/deployments",
+      "apiVersion" : "2013-03-01",
+      "dependsOn" : [
+        "[concat('Microsoft.Network/virtualNetworks/', variables('virtualNetworkName'))]",
+        "[concat('Microsoft.Storage/storageAccounts/', variables('masterStorageName'))]",
+        "[resourceId('Microsoft.Resources/deployments', 'master1')]",
+        "[resourceId('Microsoft.Resources/deployments', 'master2')]",
+        "[resourceId('Microsoft.Resources/deployments', 'master3')]",
+        "[resourceId('Microsoft.Resources/deployments', 'infranode1')]"
```
detiber: Are these dependencies accurate? I would expect the bastion host to not have dependencies on the master hosts or infra nodes (and that there is only one infra node listed).

reference-architecture/azure-ansible/azuredeploy.json
```diff
@@ -0,0 +1,1048 @@
...
+    },
+    {
+      "name" : "master2",
+      "type" : "Microsoft.Resources/deployments",
+      "apiVersion" : "2013-03-01",
+      "dependsOn" : [
+        "[concat('Microsoft.Network/virtualNetworks/', variables('virtualNetworkName'))]",
+        "[concat('Microsoft.Storage/storageAccounts/', variables('masterStorageName'))]",
+        "[resourceId('Microsoft.Resources/deployments', 'infranode1')]",
+        "[resourceId('Microsoft.Resources/deployments', 'logging')]"
```
detiber: Are these deps correct? It doesn't seem right that a master should be dependent on an infra node or logging.

reference-architecture/azure-ansible/azuredeploy.json
```diff
@@ -0,0 +1,1048 @@
...
+      }
+    },
+    {
+      "name" : "master3",
+      "type" : "Microsoft.Resources/deployments",
+      "apiVersion" : "2013-03-01",
+      "dependsOn" : [
+        "[concat('Microsoft.Network/virtualNetworks/', variables('virtualNetworkName'))]",
+        "[concat('Microsoft.Storage/storageAccounts/', variables('masterStorageName'))]",
+        "[resourceId('Microsoft.Resources/deployments', 'logging')]"
```
detiber: same question about dependencies here.

reference-architecture/azure-ansible/azuredeploy.json
```diff
@@ -0,0 +1,1048 @@
...
+    },
+    {
+      "apiVersion" : "[variables('tmApiVersion')]",
+      "type" : "Microsoft.Network/trafficManagerProfiles",
+      "name" : "MasterLB",
+      "location" : "global",
+      "dependsOn" : [
+        "[resourceId('Microsoft.Resources/deployments', 'master1')]",
+        "[resourceId('Microsoft.Resources/deployments', 'master2')]",
+        "[resourceId('Microsoft.Resources/deployments', 'master3')]"
```
detiber: Would it make more sense to use the equivalent of an ASG for masters to avoid having to map the hosts into a LB in this way?

reference-architecture/azure-ansible/azuredeploy.json
```diff
@@ -0,0 +1,1048 @@
...
+      }
+    },
+    {
+      "name" : "store1",
+      "type" : "Microsoft.Resources/deployments",
+      "apiVersion" : "2013-03-01",
+      "dependsOn" : [
+        "[concat('Microsoft.Network/virtualNetworks/', variables('virtualNetworkName'))]",
+        "[concat('Microsoft.Storage/storageAccounts/', variables('nodeStorageName'))]",
+        "[resourceId('Microsoft.Resources/deployments', 'logging')]"
```
detiber: Same question about deps

reference-architecture/azure-ansible/bastion.sh
```diff
@@ -0,0 +1,345 @@
...
+etcd
+nodes
+misc
+
+[OSEv3:vars]
+azure_resource_group=${RESOURCEGROUP}
+rhn_pool_id=${RHNPOOLID}
+openshift_install_examples=true
+deployment_type=openshift-enterprise
+openshift_master_identity_providers=[{'name': 'htpasswd_auth', 'login': 'true', 'challenge': 'true', 'kind': 'HTPasswdPasswordIdentityProvider', 'filename': '/etc/origin/master/htpasswd'}]
```
detiber: If using htpasswd, you will probably also want to specify either `openshift_master_htpasswd_users` or `openshift_master_htpasswd_file` to create and sync the initial users.

reference-architecture/azure-ansible/bastion.sh
```diff
@@ -0,0 +1,345 @@
...
+# default selectors for router and registry services
+openshift_router_selector='region=infra'
+openshift_registry_selector='region=infra'
+
+ansible_become=yes
+ansible_ssh_user=${AUSERNAME}
+remote_user=${AUSERNAME}
+
+openshift_master_default_subdomain=${ROUTEREXTIP}.xip.io 
+openshift_use_dnsmasq=False
```
detiber: Why is dnsmasq being disabled? It shouldn't be too big of an issue today, but eventually we'll need to resolve cluster addresses from the host and that will require dnsmasq being enabled (or a similar config being replicated).

reference-architecture/azure-ansible/bastion.sh
```diff
@@ -0,0 +1,345 @@
...
+
+openshift_master_cluster_method=native
+openshift_master_cluster_hostname=${RESOURCEGROUP}.trafficmanager.net
+openshift_master_cluster_public_hostname=${RESOURCEGROUP}.trafficmanager.net
+
+# Enable cockpit
+osm_use_cockpit=true
+
+# Set cockpit plugins
+osm_cockpit_plugins=['cockpit-kubernetes']
```
detiber: osm_use_cockpit and osm_cockpit_plugins are the defaults, they don't need to be specified here.

reference-architecture/azure-ansible/bastion.sh
```diff
@@ -0,0 +1,345 @@
...
+
+# default storage plugin dependencies to install, by default the ceph and
+# glusterfs plugin dependencies will be installed, if available.
+osn_storage_plugin_deps=['iscsi']
+
+[masters]
+master1 openshift_node_labels="{'role': 'master'}"
+master2 openshift_node_labels="{'role': 'master'}"
+master3 openshift_node_labels="{'role': 'master'}"
```
detiber: rather than specifying `openshift_node_labels` here, I would add `'role': 'master'` to the master{1,2,3} entries in the `nodes` group.

reference-architecture/azure-ansible/bastion.sh
```diff
@@ -0,0 +1,345 @@
...
+node[01:${NODECOUNT}] openshift_node_labels="{'region': 'primary', 'zone': 'default'}"
+infranode1 openshift_node_labels="{'region': 'infra', 'zone': 'default'}"
+infranode2 openshift_node_labels="{'region': 'infra', 'zone': 'default'}"
+infranode3 openshift_node_labels="{'region': 'infra', 'zone': 'default'}"
+
+[quotanodes]
+master1 openshift_node_labels="{'region':'master','zone':'default'}" 
+master2 openshift_node_labels="{'region':'master','zone':'default'}" 
+master3 openshift_node_labels="{'region':'master','zone':'default'}" 
+node[01:${NODECOUNT}] openshift_node_labels="{'region': 'primary', 'zone': 'default'}"
```
detiber: There should be no need to specify openshift_node labels here.

reference-architecture/azure-ansible/bastion.sh
```diff
@@ -0,0 +1,345 @@
...
+store1
+EOF
+
+
+cat <<EOF > /home/${AUSERNAME}/subscribe.yml
+---
+- hosts: all
+  vars:
+    description: "Subscribe OSE"
+  tasks:
```
detiber: This should probably be done in serial, there are known issues with doing parallel subscriptions at the same time (especially with large accounts).

reference-architecture/azure-ansible/bastion.sh
```diff
@@ -0,0 +1,345 @@
...
+- hosts: quotanodes
+  vars:
+    description: "Fix EP Storage/Quota"
+  tasks:
+  - name: Format xfs on disk
+    filesystem: dev=/dev/sdd fstype=xfs force=yes
+  - name: Update Mount to Handle Quota
+    mount: fstype=xfs name=/var/lib/origin/openshift.local/volumes src=/dev/sdd opts="gquota" state="mounted"
+  - name: Change Node perFSGroup Qouta Setting
+    lineinfile: dest=/etc/origin/node/node-config.yaml regexp=^perFSGroup line="    perFSGroup:512Mi"
```
detiber: Can perFSGroup not be specified during install?

Also, does Azure support cloud-init or something like it? It seems like the xfs config would make sense to do that way rather than as a separate step.

reference-architecture/azure-ansible/bastion.sh
detiber: Can this playbook be stored in the repo and transferred in some way? It would make it easier to re-use between cloud providers rather than having it hard coded in the script.

reference-architecture/azure-ansible/bastion.sh
```diff
@@ -0,0 +1,345 @@
...
+EOF
+
+
+cat <<EOF > /home/${AUSERNAME}/setupiscsi.yml
+- hosts: all
+  vars:
+    description: "Subscribe OSE"
+  tasks:
+  - name: Install iscsi initiator utils
```
detiber: This seems like a good candidate for a role in openshift-ansible-contrib rather than a playbook written out by a script.

reference-architecture/azure-ansible/bastion.sh
```diff
@@ -0,0 +1,345 @@
...
+cat <<EOF > /home/${AUSERNAME}/postinstall.yml
+---
+- hosts: masters
+  vars:
+    description: "auth users"
+  tasks:
+  - name: Create Master Directory
+    file: path=/etc/origin/master state=directory
+  - name: add initial user to OpenShift Enterprise
+    shell: htpasswd -c -b /etc/origin/master/htpasswd ${AUSERNAME} ${PASSWORD}
```
detiber: This can be done during install using openshift_master_htpasswd_users or openshift_master_htpasswd_file

reference-architecture/azure-ansible/bastion.sh
```diff
@@ -0,0 +1,345 @@
...
+local_tmp      = ~/.ansible/tmp
+host_key_checking = False
+forks=30
+gather_timeout=60
+timeout=240
+[ssh_connection]
+control_path = ~/.ansible/cp/ssh%%h-%%p-%%r
+ssh_args = -o ControlMaster=auto -o ControlPersist=600s -o ControlPath=~/.ansible/cp-%h-%p-%r
+EOF
+chown ${AUSERNAME} /home/${AUSERNAME}/.ansible.cfg
```
detiber: I would suggest putting the ansible.cfg file in repo (and running ansible from that working directory).

reference-architecture/azure-ansible/store.sh
```diff
@@ -0,0 +1,81 @@
...
+touch /root/.updateok
+pvcreate /dev/sdc /dev/sdd /dev/sde /dev/sdf /dev/sdg /dev/sdh /dev/sdi /dev/sdj
+parted --script -a optimal /dev/sdc mklabel gpt mkpart primary ext2 1M 100% set 1 lvm on
+parted --script -a optimal /dev/sdd mklabel gpt mkpart primary ext2 1M 100% set 1 lvm on
+parted --script -a optimal /dev/sde mklabel gpt mkpart primary ext2 1M 100% set 1 lvm on
+parted --script -a optimal /dev/sdf mklabel gpt mkpart primary ext2 1M 100% set 1 lvm on
+parted --script -a optimal /dev/sdg mklabel gpt mkpart primary ext2 1M 100% set 1 lvm on
+parted --script -a optimal /dev/sdh mklabel gpt mkpart primary ext2 1M 100% set 1 lvm on
+parted --script -a optimal /dev/sdi mklabel gpt mkpart primary ext2 1M 100% set 1 lvm on
+parted --script -a optimal /dev/sdj mklabel gpt mkpart primary ext2 1M 100% set 1 lvm on
```
detiber: shouldn't these be xfs and not ext2?

reference-architecture/azure-ansible/store.sh
```diff
@@ -0,0 +1,81 @@
...
+pvcreate /dev/sdc1 /dev/sdd1 /dev/sde1 /dev/sdf1 /dev/sdg1 /dev/sdh1 /dev/sdi1 /dev/sdj1
+vgcreate vg1 /dev/sdc1 /dev/sdd1 /dev/sde1 /dev/sdf1 /dev/sdg1 /dev/sdh1 /dev/sdi1 /dev/sdj1
+cat <<EOF | base64 --decode >  /root/ose_pvcreate_lun
```
ashcrow: Putting on my ops hat for a second, this would freak me out until I decoded and reviewed what it was I was about to run...

themurph: Why is there a base64 encoded file in the middle of this script? Can it be added to the git repo and retrieved from there during the install?

Side note: this will get flagged by security scanning tools.

reference-architecture/azure-ansible/azuredeploy.json
glennswest: Premium storage as standard storage is not recommended.

reference-architecture/azure-ansible/azuredeploy.json
glennswest: actually no centos support at all.

reference-architecture/azure-ansible/azuredeploy.json
glennswest: the infra nodes are always limited to three and there a interelationship to traffic manager, which would make doing so difficult.

reference-architecture/azure-ansible/azuredeploy.json
glennswest: Handled at ansible time. And also to prevent looping on dependency that causes a deadlock.

reference-architecture/azure-ansible/azuredeploy.json
glennswest: Handled in ansible level.

reference-architecture/azure-ansible/azuredeploy.json
glennswest: THis is just making sure there "created" so we can pull the IP's, otherwise a race condition can occur.

reference-architecture/azure-ansible/bastion.sh
glennswest: did have that I believe at one point, and something was not working, I switched to this to resolve it.

reference-architecture/azure-ansible/bastion.sh
glennswest: key work is "should" not be, I had a day where the labels were not getting actually tagged. This was to "resolve" getting the nodes tagged reliably.

reference-architecture/azure-ansible/bastion.sh
glennswest: This is all tested multiple times, with retry logic etc that has proven to work every time.
I've tried a variety of ways of doing this, and current logic resolves all. This includes rhn and ansible bugs, and hangs. Single Threaded stil ldoes not "solve" the varous and wonderful rhn and subscription manager issues, the retry logic works well.

glennswest: The overall azure architecture is one bash sh per host, and that sh creates all assets, this is true all the way true. I think u are right if: subscription-manager is fixed, rhn is fixed, channels are disabled properly. Even found that the ansible "subscription-manager" module that is supposed to handle things well, changes functionality from time to time. So short term, I'd say it decreases reliablity. Also note this may get better as RHN is undergoing massive changes. 

Im very comfortable with lots of runs, that the current code recovers from all displayed "variants". 

My biggest cause of failure of runs was related to subscription manager/subscription module thru changes in ansible. 

This is based on greater than 220 test runs,  really wish it was simpler, but reality seems to challenge simplicity.

reference-architecture/azure-ansible/bastion.sh
glennswest: cloudinit is a massive work in process, because microsoft "forked" there own, and eventually will converge back, but this is a base evangelistm, and a huge change in ARM as well. 

Also base storage setup really should be done before install, as its part of docker setup, which is a precursor. I do wish I could do different host groups in one playbook, but not seen that one done. Then could potentially combine things. I did find during development, that a bit of separation does improve debuggging, so some spliting was intention to be able to "retry" manually and look at cause of failure.

reference-architecture/azure-ansible/store.sh
glennswest: actually can change this, but the reality is this is a label and a unused partition. Note the size. 
The actual pv used are created by pv_create_lun later in the script.

reference-architecture/azure-ansible/README.md
```diff
@@ -0,0 +1,75 @@
+# RedHat Openshift Enterprise cluster on Azure
+
+When creating the RedHat Openshift Enterprise Cluster on Azure, you will need a SSH RSA key for access.
+
+## SSH Key Generation
+
+1. Windows - https://www.digitalocean.com/community/tutorials/how-to-create-ssh-keys-with-putty-to-connect-to-a-vps
+2. Linux - https://help.ubuntu.com/community/SSH/OpenSSH/Keys#Generating_RSA_Keys
```
glennswest: :) Links

reference-architecture/azure-ansible/bastion.sh
```diff
@@ -0,0 +1,345 @@
...
+sleep 30
+cat <<EOF > /root/setup_ssmtp.sh
+# \$1 = Gmail Account (Leave off @gmail.com ie user)
+# \$2 = Gmail Password
+# \$3 = Notification email address
+# Setup ssmtp mta agent for use with gmail
+yum -y install wget
+wget -c https://dl.fedoraproject.org/pub/epel/7/x86_64/e/epel-release-7-8.noarch.rpm
+rpm -ivh epel-release-7-8.noarch.rpm
+yum -y install ssmtp
```
glennswest: one package only.

reference-architecture/azure-ansible/bastion.sh
```diff
@@ -0,0 +1,345 @@
...
+yum-config-manager --disable epel
+yum-config-manager --disable epel-testing
+subscription-manager register --username $RHNUSERNAME --password $RHNPASSWORD
+subscription-manager attach --pool=$RHNPOOLID
+subscription-manager repos --disable="*"
+subscription-manager repos     --enable="rhel-7-server-rpms"     --enable="rhel-7-server-extras-rpms"
+subscription-manager repos     --enable="rhel-7-server-ose-3.3-rpms"
+yum -y install atomic-openshift-utils
+yum -y install git net-tools bind-utils iptables-services bridge-utils bash-completion httpd-tools
+yum -y install docker
```
glennswest: Yep, but it also makes it clear when it blows up. espesially on atomic-openshift-utils.

reference-architecture/azure-ansible/bastion.sh
glennswest: Im using Keycloak, and handle all via that. Its a better solution. Not integerated yet, in a separate repo.

reference-architecture/azure-ansible/bastion.sh
```diff
@@ -0,0 +1,345 @@
...
+    ignore_errors: yes
+  - name: attach sub
+    shell: subscription-manager attach --pool=$RHNPOOLID
+    register: task_result
+    until: task_result.rc == 0
+    retries: 10
+    delay: 30
+    ignore_errors: yes
+  - name: disable all repos
+    shell: subscription-manager repos --disable="*"
```
glennswest: Guys they dont work. subscription-manager ansible module is unreliable. This is done this way for reliablity based on 200+ runs. Even disable * does not work. Its wonderful to be "clean", but reality is to get things to run everytime, with ansible its the way it is. Changing it to the common role, is where i started from, and wasted lots of time. 
Changing this will "look" cleaner and fail often.

reference-architecture/azure-ansible/bastion.sh
```diff
@@ -0,0 +1,345 @@
...
+export ANSIBLE_HOST_KEY_CHECKING=False
+sleep 120
+ansible all --module-name=ping > ansible-preinstall-ping.out || true
+ansible-playbook  /home/${AUSERNAME}/subscribe.yml
+echo "${RESOURCEGROUP} Bastion Host is starting ansible BYO" | mail -s "${RESOURCEGROUP} Bastion BYO Install" ${RHNUSERNAME} || true
+ansible-playbook  /usr/share/ansible/openshift-ansible/playbooks/byo/config.yml < /dev/null &> byo.out
+
+# ssh gwest@master1 oadm registry --selector=region=infra
+# ssh gwest@master1 oadm router --selector=region=infra
+wget http://master1:8443/api > healtcheck.out
```
glennswest: Religion debaswe, I prefer wget.
really dont like curl.

reference-architecture/azure-ansible/bastion.sh
themurph: "The overall azure architecture is one bash sh per host, and that sh creates all assets..."

@glennswest can you briefly describe what "one bash sh per host" means? Do you mean one tty, therefore one ssh session? Are you using Bash in all places? I noticed that some scripts are started using "sh <script_file>" with no SheBang "#!/bin/bash" in the script, which causes Bash to run in a minimalist mode and cause many scripts will fail, even though they work fine when called with /bin/bash.

glennswest: In azure ARM, a "extension" is used to download and launch a shell script. This is the bridge 
between the ARM(Azure Resource Manager), and the resultant host. so bastion has "bastion.sh" store vm has "store.sh" and so on. there is a one to one match for each vm type. THe most complex of course is bastion as it running ansible.

reference-architecture/azure-ansible/bastion.sh
themurph: I prefer wget also, the only reason I brought this up was because you mentioned that your installs were taking 4 hours, so I was looking for places to speed up your install. The "yum -y install wget" on line 44 can sometimes take upwards of 3-4 minutes when mirrors are slow. You used curl in the azure-ansible/store.sh script so I just assumed you were comfortable with both. Feel free to call this comment resolved.

reference-architecture/azure-ansible/store.sh
themurph: I've read the script https://paste.fedoraproject.org/483513/36016614/ and it doesn't seem very "narly". It's pretty straight forward and could easily be included in this repo and cloned to the host or pulled using wget without causing any "arguments, etc getting translated".

glennswest:
Then store host is done differently than all the host. And the script is only used on store.sh

I prefer a consistency in ways things are done 

Each host type has one script

reference-architecture/azure-ansible/README.md
themurph: https://lmgtfy.com/?q=ssh+keys+site%3Aredhat.com

glennswest:
Thought u had a Favourite one vs the ones already in my documentation

reference-architecture/azure-ansible/README.md
themurph: No preference on a single link. I was trying to subtly suggest using our own documentation (from OpenShift Docs or Red Hat Docs) instead of a link from ubuntu.

glennswest: sad part is the ubuntu one is clearer :( 
and I dont like the digital ocean one much better, but putty not exactly documented in our docs.
Im still agravated that subscription-manager disable * actually is a No-Op, yet its in every install guide. In my case easy to fix, but hate to think the pain it causes.
Man its must be close to 1AM for u.

reference-architecture/azure-ansible/bastion.sh
glennswest: actually I was specifcally ask to include these quota settings by scott. That they "are" required as best practice, and are considered a must have. :) Im sure we can go around again on this, during running, I didnt see a net change.

reference-architecture/azure-ansible/store.sh
glennswest: resolved.

reference-architecture/azure-ansible/store.sh
glennswest: found the trick, its now in clear text. So totally transparent. - Resolved (Note till I test everything is going up to my github till I can test it all)

glennswest: actually found the way of not escaping the world, its pretty easy - just didnt realize you colud do it that way. So now the whole script is clear-text, testing now.
actually time is more "time of day" on azure, ie if u hit it right, its really fast.
45 minutes

Also suspect it would be quicker if I cached the rpms.

glennswest: actually its azure getting there vm's up, after paper I suspect I can speed things up. Have two concepts that could make it alot faster.

reference-architecture/azure-ansible/bastion.sh
detiber: dnsmasq isn't really redundant, because it is used to allow the pods/nodes to access OpenShift Cluster DNS records in skydns.

reference-architecture/azure-ansible/bastion.sh
ashcrow: Agreed. While I also like wget over curl it would speed things up slightly to use what's already available.

reference-architecture/azure-ansible/bastion.sh
ashcrow: @glennswest if there isn't an issue for that already created do you mind reporting it against the  appropriate installer roles?

reference-architecture/azure-ansible/bastion.sh
detiber: Right, I get wanting to set the params in an ansible.cfg file, I was just recommending the file exist as `ansible.cfg` in the working directory that you are running ansible from rather than setting it for the users. If users are going to write their own playbooks to augment the ones put in place, then it would be a poor experience if the `.ansible.cfg` files laid down break some of their assumptions.

reference-architecture/azure-ansible/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+```powershell
+New-AzureRmResourceGroupDeployment -Name <DeploymentName> -ResourceGroupName <RessourceGroupName> -TemplateUri https://raw.githubusercontent.com/openshift/openshift-ansible-contrib/master/reference-architecture/azure-ansible/azuredeploy.json
+```
+
+
+## Parameters
+SSHKeyData - Public Key - Copy/Past frome .ssh/id_rsa.pub - pbcopy < ~/.ssh/id_rsa.pub
+PoolId - From RREL - subscription-manager list --available
+SSHPrivate Data - Base64 Encoded id_rsa - cat ~/.ssh/id_rsa | base64 | pbcopy
```
ashcrow: May be helpful: http://superuser.com/questions/288320/whats-like-osxs-pbcopy-for-linux

reference-architecture/azure-ansible/bastion.sh
detiber: So, there are no official "installer roles" for this specifically because of the issues @glennswest mentions.

That said, I think there will be a certain scale where this approach will no longer work.

I've been working with the subscription-manager team, and there is work going on to help address the problems, but there are changes needed both on the server and client side to make this much  more reliable at scale.

glennswest: Dont forget there major rewrite on rhn as well, also the general transition to certificates that is on going. Will review this on next release of ansible, as I expect it will improve.

reference-architecture/azure-ansible/bastion.sh
glennswest: True, but considering the way these are "parsed" has changed quite a bit, I'd say it needs to be in several places for now. As I ran into issues with ansible not reading parameters. Actually this whole issue, is there default, buried deep in ansible was not set properly. There already a "bug" on it upstream, but I didnt see the timing when it would "flow" downwards, so till next "ansible", this was a work around. (And over 2 weeks of time tracking what the cause is.

reference-architecture/azure-ansible/bastion.sh
glennswest: actually, suprised this is not a default in the generated configs. Is there a way of doing this in the host files? would be more portable.

glennswest:
Actually my intention to limit single greenfield install to 30 nodes and the. Grow beyond with add node at ten at a time 

This is based on feedback from scot on Openstack testing

cooktheryan:
@glennswest what are your thoughts on me merging #125 into this pr that way we have a burn down list with the concerns

glennswest:
I did try to do a brief summary from this PR point of view. Not sure long term it helps merging it, as there so many changes already cooked in, and like to have a clean starting point in contrib. 
Also there is a complete commit history upstream, that I do reference in the README. 

So, in summary, I'd like to have a nice clean base to work from, and test, and then layer any additional small changes on top.

reference-architecture/azure-ansible/bastion.sh
detiber: openshift_node_local_quota_per_fsgroup

reference-architecture/azure-ansible/bastion.sh
detiber: The lookup order for the ansible config is as follows:
- cwd
- ~/.ansible.cfg
- /etc/ansible/ansible.cfg

The first file found is the one used, there is no merging of config files.

PR #116, opened 2016-11-16
better setting of vars to avoid confusion

cooktheryan:


reference-architecture/aws-ansible/playbooks/validation.yaml
```diff
@@ -6,10 +6,10 @@
   pre_tasks:
       - name: set fact
         set_fact:
-           openshift_master_cluster_public_hostname: "{{ openshift_master_cluster_public_hostname }}"
+           openshift_master_cluster_public_hostname: "openshift-master.{{ public_hosted_zone }}"
       - name: set fact
         set_fact:
-           openshift_master_cluster_hostname: "{{ openshift_master_cluster_hostname }}"
+           openshift_master_cluster_hostname: "internal-openshift-master.{{ public_hosted_zone }}"
```
detiber: Is there a private hosted zone that could be used instead of the public_hosted_zone? This should only be applied to and used for the internal-only load balancer

cooktheryan: Right so the play is to go to one of the masters and do a web call to ensure that they can hit the healthz page of the internal-only load balancer.  That way we can verify nothing crazy was set in regards to the cluster_hostname and the internal cluster vip is healthy. Making troubleshooting a bit easier if things weren't working as expected

detiber: Are you using split-horizon DNS? I'm just wondering about exposing out an internal-only DNS name in the public zone.

cooktheryan:
With the merging of validation this isn't required anymore

PR #117, opened 2016-11-16
New version of openshift playbooks better s3 support

cooktheryan:


dav1x:
LGTM - Also reminds me I need to do this and abstract out the registry role completely! Nice work

PR #119, opened 2016-11-18
Ports for metrics and logging 

cooktheryan:
Currently not offering metrics in logging in the reference architecture but if a customer were wanting to add the feature it would be good to have the ports open.

### DIFF
--- a/reference-architecture/aws-ansible/playbooks/roles/cloudformation-infra/files/brownfield-byo-bastion.json
+++ b/reference-architecture/aws-ansible/playbooks/roles/cloudformation-infra/files/brownfield-byo-bastion.json
@@ -407,8 +407,8 @@
       "Properties":{
         "GroupId": { "Fn::GetAtt": [ "MasterSG", "GroupId" ] },
         "IpProtocol": "tcp",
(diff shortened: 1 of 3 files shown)


reference-architecture/aws-ansible/playbooks/roles/cloudformation-infra/files/brownfield-byo-bastion.json
```diff
@@ -319,7 +319,7 @@
         "IpProtocol": "tcp",
         "FromPort": "10250",
         "ToPort": "10250",
-        "SourceSecurityGroupId": { "Fn::GetAtt": [ "MasterSG", "GroupId" ] }
+        "SourceSecurityGroupId": { "Fn::GetAtt": [ "NodeSG", "GroupId" ] }
```
detiber: I wonder if there should be rules for both MasterSG and NodeSG, not all SDN solutions will require that the Master be a Node.

cooktheryan: lemme know if the latest commit resolves that concern. I think thats what you meant

PR #124, opened 2016-11-21
More vagrant

cgwalters:


cgwalters:
Ah wait, bug in the last commit.  One sec.

PR #144, opened 2017-01-03
Enable dnsmasq for azure name resolution and fix app load balacer naming

glennswest:
Resolved 2 issues:
     1. No dns names on infra nodes resulting in failure of app load-balancer.
     2. Naming resolution for Azure Names not working due to dnsmasq changes in previous PR.


cooktheryan:
Talked to @glennswest and dnsmasq will be used by azure as well as the default use of ocp/origin.

+1

PR #146, opened 2017-01-03
fixed RH_PASSWORD variable to be compatible with passwords containing…

Nick-Harvey:
… special characters

### DIFF
--- a/reference-architecture/gce-cli/gcloud.sh
+++ b/reference-architecture/gce-cli/gcloud.sh
@@ -367,7 +367,7 @@ if ! gcloud --project "$GCLOUD_PROJECT" compute images describe "$REGISTERED_IMA
         sleep 5
     done
     if ! gcloud -q --project "$GCLOUD_PROJECT" compute ssh "cloud-user@${TEMP_INSTANCE}" --zone "$GCLOUD_ZONE" --ssh-flag="-t" --command "sudo bash -euc '
-        subscription-manager register --username=${RH_USERNAME} --password="\"${RH_PASSWORD}\"";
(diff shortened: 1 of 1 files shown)


cooktheryan:
@glennswest would this be applicable to Azure as well

glennswest:
I've added this support in, nice catch.

cooktheryan:
This LGTM

cooktheryan:
Hate to string this along PR much longer but @dav1x and @markllama can you see if your providers are affected by this as well. I will be testing AWS tomorrow but I believe I face the same issue.

reference-architecture/gce-cli/gcloud.sh
```diff
@@ -367,7 +367,7 @@ if ! gcloud --project "$GCLOUD_PROJECT" compute images describe "$REGISTERED_IMA
         sleep 5
     done
     if ! gcloud -q --project "$GCLOUD_PROJECT" compute ssh "cloud-user@${TEMP_INSTANCE}" --zone "$GCLOUD_ZONE" --ssh-flag="-t" --command "sudo bash -euc '
-        subscription-manager register --username=${RH_USERNAME} --password=${RH_PASSWORD};
+        subscription-manager register --username=${RH_USERNAME} --password="\"${RH_PASSWORD}\"";
```
pschiffe: @Nick-Harvey thanks for a catch and a PR, but shouldn't be `--password=\"${RH_PASSWORD}\";` enough? Could you check and possibly, update the PR? Thanks.

Nick-Harvey:
You're right. It should just be `password=\"${RH_PASSWORD}\"` the additional `"` were artifacts from some of my earlier troubleshooting. I'll make the changes and retest to make 100% sure that it works and then resubmit pull request. Standby

PR #150, opened 2017-01-05
Removal of strict ansible version

cooktheryan:
The dependencies ship working versions of both the playbooks/roles and ansible.


cooktheryan:
aos-ci-test

PR #154, opened 2017-01-07
Add network restart to increase local dns registration reliability

glennswest:
Restart the network service in first boot to make sure our local dns registration is working.


glennswest:
Actually good catch. The eye wanted all the systemctl in order, but realtiy is, the network should be done first. So fixed the order on the three files.

cooktheryan:
Approved as this is just a work around until the provider bug is resolved.

Both ifup and ifdown and network restart will be removed upon fix.

PR #155, opened 2017-01-07
Make the network restart faster/earlier in cycle

glennswest:
The nodes are registering properly, master/bastion are not. Difference appears to be timing.
Optimize and make the process occur sooner.


cooktheryan:
Part of the bug work around

PR #159, opened 2017-01-10
Temp roll back to RHEL 7.2

glennswest:
Change base version to 7.2 - Note that the upgrade process will still result in 7.3 being used.
This should make everything work properly, at the cost of time.


cooktheryan:
Did you mean to include all of the extra files?

PR #167, opened 2017-01-12
Add new variables, and fixup app load balancers.

glennswest:


cooktheryan:
Approved @glennswest can you catch your branch up when making the next PR that way there will only be commit entries relevant to the PR

PR #185, opened 2017-01-15
the fedora/25-cloud-base box is avaliable.

warmchang:
check the "https://atlas.hashicorp.com/fedora/boxes/25-cloud-base/".

### DIFF
--- a/vagrant/Vagrantfile
+++ b/vagrant/Vagrantfile
@@ -51,7 +51,7 @@ if deployment_type == 'openshift-enterprise'
 elsif origin_os == 'centos'
   box_name = 'centos/7'
 else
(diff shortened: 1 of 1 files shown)


cooktheryan:
@detiber do we have someone familiar with Vagrant that can +1 this?

detiber:
@cgwalters @dustymabe are either of you able to verify this change?

cgwalters:
I haven't tested it, but makes sense to me.

cgwalters:
The cloud base box is the yum managed, non-Atomic Host variant.

dustymabe:
@detiber - I don't have a setup for that right now. if @warmchang says it works then I say merge it.

Issue #175, opened 2017-01-13
[refarch/ocp-cli] script should create user-data directory if it does not exist and cleanup should clear it.


markllama:
The create_all.sh script (or the user-data script) should create the required `user-data` subdir if it does not exist.

If it does not exist the user data scripts are not created and instance boot fails.

markllama:
PR https://github.com/openshift/openshift-ansible-contrib/pull/190

PR #190, opened 2017-01-16
create the user-data directory if needed

markllama:
This patch creates the `user-data` directory if it does not exist.  If the directory is not created, following attempts to write user data files for new instances will also fail and the boot commands will then fail.

This patch removes the requirement that the caller pre-create the `user-data` direcory.


markllama:
Issue: https://github.com/openshift/openshift-ansible-contrib/issues/175

Issue #176, opened 2017-01-13
[refarch/ocp-cli] nova boot does not recognize `--nic net-name` in RHEL7 nova


markllama:
Boot scripts must use `--nic net-id` to attach networks to new instances.  `--nic net-name=` exists upstream but not yet in nova on rhel7.

markllama:
This will require a short function: 

----
function net_id_by_name() {
  # NETNAME=$1
  neutron net-list --field id --field name | grep "$1" | cut -d' ' -f2
}

function net_id_by_name() {
  openshift network list -f json | jq <query id when name=$1>
}
----

markllama:
PR: https://github.com/openshift/openshift-ansible-contrib/pull/191

PR #191, opened 2017-01-16
added function to find net id by name for use on rhel7.x

markllama:
This change causes the instance boot scripts to use the `--nic net-id=` form for identifying networks when booting new instances.  The `--nic net-name` form is not yet available on RHEL7.x for the nova CLI client.

This change adds a function `net_id_by_name` to each script that runs a `nova boot` and then uses that to look up the network id.

This change allows the refarch manual CLI scripts to execute properly on RHEL7x hosts.


markllama:
Issue https://github.com/openshift/openshift-ansible-contrib/issues/176

PR #195, opened 2017-01-17
Vmware guest

dav1x:
Added vmware guest module to ansible and removed a bunch of unneeded roles and variables.

Added INI support and added in 2 option netmask and v datacenter.

Tested this branch from scratch and works fine.


reference-architecture/vmware-ansible/README.md
```diff
@@ -19,18 +19,19 @@ rpm -Uvh https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm*
...
-                  python-httplib2
-git clone https://github.com/dannvix/pySphere && cd pySphere/ && python setup.py install
+                  python-simplejson \
+                  python-six \
+
 git clone https://github.com/vmware/pyvmomi && cd pyvmomi/ && python setup.py install
 
 # Grabbed the patched vsphere_guest until the PR is merged
-cp vsphere_guest.py /usr/lib/python2.7/site-packages/ansible/modules/core/cloud/vmware/
+cp vmware_guest.py /usr/lib/python2.7/site-packages/ansible/modules/core/cloud/vmware/
```
cooktheryan: Davis can you see if you can use the module similar to this?

https://github.com/openshift/openshift-ansible-contrib/tree/master/reference-architecture/aws-ansible/playbooks/library

dav1x:
@cooktheryan you mean to just store the module for use?

reference-architecture/vmware-ansible/README.md
cooktheryan: From the comment below.

Yea just store it within the repo and don't copy it to the ansible directory. One concern I would have would be overwriting something in the current Ansible directory or if an Ansible patch comes out the module may be overwrote or deleted and that could break us. If we ship it within our repo then we know we have a known good module until upstream makes it to fully supported.

roles/validate-etcd/tasks/main.yaml
```diff
@@ -5,7 +5,7 @@
   register: etcd_result
 
 - name: Validate etcd
-  command: "etcdctl -C https://{{ ansible_nodename }}:2379 --ca-file=/etc/origin/master/master.etcd-ca.crt --cert-file=/etc/origin/master/master.etcd-client.crt --key-file=/etc/origin/master/master.etcd-client.key cluster-health | grep 'cluster is'"
+  command: "etcdctl -C https://{{ ansible_fqdn }}:2379 --ca-file=/etc/origin/master/master.etcd-ca.crt --cert-file=/etc/origin/master/master.etcd-client.crt --key-file=/etc/origin/master/master.etcd-client.key cluster-health | grep 'cluster is'"
```
cooktheryan: This works perfectly.

reference-architecture/vmware-ansible/README.md
dav1x: The only problem is I can't call the module as you would in a normal playbook if I dont move it to that directory. 

I could rename it as to not overwrite the existing vmware_guest. vmware_guest_ocp or something like that

dav1x: done!

PR #199, opened 2017-01-18
Brownfield is not calling the template correctly

cooktheryan:
This fixes an a missing variable issue when it comes to a brownfield deployment and allows for the correct stack to be called when asking for cloudformation outputs.


dav1x:
LGTM

PR #205, opened 2017-01-20
Vmware readme.md updates

dav1x:


detiber:
@dav1x commit history looks a bit messy for the changeset.

PR #214, opened 2017-01-22
VMware - Fix ssh_key filename to match ansible.cfg

### LINKED ISSUE #213: VMware - fix the ssh_key filename
Following recent change where the ssh private key was renamed from `ose3-installer` to `ocp-installer` we missed to sync the change in the ssh_key dir

DanyC97:
Fixes #213


dav1x:
LGTM

PR #220, opened 2017-01-22
VMware - parameterize the deployment_type

### LINKED ISSUE #219: VMware - parameterize the deployment_type
In a lab is quite useful to use a CentOS and deploy the community OpenShift origin. 
In order to do that we should allow the deployment_type to be parameterized rather than being hardcoded to `openshift-enterprise`

DanyC97:
Fixes #219


dav1x:
tested LGTM!

PR #221, opened 2017-01-22
Fix Vagrant roles symlink

chekolyn:
Looks like a couple of role symlinks are broken for vagrant


detiber:
@chekolyn Thanks for the contribution!

PR #223, opened 2017-01-23
Delete post-install-check.yaml file and replace it with post-validati…

### LINKED ISSUE #222: Remove duplicated playbook file
In `https://github.com/openshift/openshift-ansible-contrib/tree/master/playbooks` we have `post-install-check.yaml` and `post-validation.yaml` which are identical and where only the latter one is used.

From naming pov it make sense for me to use the latter one `post-validation.yaml` and delete the former one

DanyC97:
…on.yaml

Fixes #222


cooktheryan:
aos-ci-test

DanyC97:
@cooktheryan fyi the failed test is all down to your infra, nothing to do with the change (just checked the logs)

cooktheryan:
Agree I dont believe it should be a blocker. I believe it was just a fluke
occurance

cooktheryan:
@dav1x if you do not see any issues with this we can merge

dav1x:
LGTM!

PR #229, opened 2017-01-25
WIP: Introduction of logging and metrics

cooktheryan:


reference-architecture/aws-ansible/playbooks/openshift-setup.yaml
```diff
@@ -53,3 +53,7 @@
     openshift_hosted_registry_pullthrough: true
     openshift_hosted_registry_acceptschema2: true
     openshift_hosted_registry_enforcequota: true
+    openshift_hosted_metrics_deploy: "{{ metrics }}" 
+    openshift_hosted_logging_deploy: "{{ logging }}"
```
detiber: might want to use the `bool` filter for these just to be safe. Also, shouldn't these be `openshift_metrics` and `openshift_logging` based on the variables being set in ose-on-aws.py?

PR #230, opened 2017-01-25
refactor restart atomic-openshift-node task to work with origin

bdurrow:
Instead of calling service atomic-openshift-node restart use the correct service and for other deployment types (like origin)


cooktheryan:
reviewing now

PR #234, opened 2017-01-26
[WIP] AWS Provisioning changes

detiber:
Still rough, but wanted to start getting something out there.

Current goal with this initial PR is to start on the VPC configuration, other components will be follow up PRs

Still to do:
- subnets
- routes
- igw
- nat gateway
- ???


playbooks/provisioning/aws/provision_aws.yml
```diff
@@ -0,0 +1,16 @@
...
+  become: no
+  gather_facts: no
+  pre_tasks:
+  - debug: var=openshift_cluster_id
+  - debug: var=openshift_env_id
+
+  # TODO: properly handle these defaults, currently role defaults are overriding
+  - fail:
+      msg: "Missing required variable: openshift_cluster_id"
+    when: openshift_cluster_id is not defined
```
stevekuznetsov: I wish I had written down why I ended up choosing this route, but there were some cases where `is not defined` didn't catch things so I ended up using `when: item not in vars and item not in hostvars[inventory_hostname]`. I'll try to remember why.

roles/openshift_provisioning_aws_facts/README.md
```diff
@@ -0,0 +1,43 @@
...
+------------
+
+None
+
+Role Variables
+--------------
+
+| Name                                   | Default value                                       |                     |
+|----------------------------------------|-----------------------------------------------------|---------------------|
+| openshift_cluster_id                   | default                                             | Cluster ID (combination of openshift_cluster_id and openshift_env_id are expected to be unique for all OpenShift clusters) |
```
stevekuznetsov: I wonder if it would be better to try to get `-i one -i two` upstream so we can just spatially separate inventories for clusters.

playbooks/provisioning/aws/provision_aws.yml
detiber: hmm, seems odd, maybe that avoids values that are set in defaults, which is what I'm hitting.

roles/openshift_provisioning_aws_facts/README.md
detiber: I don't think this prevents that approach, but we still need a way to do unique naming by default.

playbooks/provisioning/aws/cloudformation/cloudformation_setup.yml
```diff
@@ -1,61 +0,0 @@
...
-  gather_facts: no
-  vars_files:
-  - vars.yml
-  vars:
-    vpc_subnet_azs: "{{ lookup('ec2_zones_by_region', ec2_region) }}"
-    r53_subdomain: "{{ cluster_id }}.{{ env_id }}.{{ r53_zone }}"
-  tasks:
-  - name: 'Validating options'
-    fail:
-      msg: required values not set
```
dgoodwin: Will this print the actual option not set?

roles/openshift_provisioning_aws_vpc/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+Role Variables
+--------------
+
+| Name                                   | Default value                                       |                     |
+|----------------------------------------|-----------------------------------------------------|---------------------|
+| openshift_provisioning_aws_vpc_id      | UNDEF                                               | Existing VPC ID to use |
+| openshift_provisioning_aws_vpc_name    | openshift_<openshift_cluster_id>_<openshift_env_id> | name for the created vpc |
+| openshift_provisioning_aws_vpc_cidr    | 172.18.0.0/16 | cidr for the created vpc |
+| openshift_provisioning_aws_vpc_tenancy | default | type of tenancy for VPC (default or dedicated) |
```
dgoodwin: opav_id, opav_name, how do we feel about adopting this format when we get into variables being passed specifically into roles. (not globally used)

roles/openshift_provisioning_aws_vpc/README.md
```diff
@@ -0,0 +1,65 @@
...
+------------
+
+boto
+
+Role Variables
+--------------
+
+| Name                                   | Default value                                       |                     |
+|----------------------------------------|-----------------------------------------------------|---------------------|
+| openshift_provisioning_aws_vpc_id      | UNDEF                                               | Existing VPC ID to use |
```
dgoodwin: Is this one just used when deleting a VPC?

roles/openshift_provisioning_aws_vpc/meta/main.yml
```diff
@@ -0,0 +1,16 @@
...
+  platforms:
+  - name: Fedora
+    versions:
+    - 24
+    - 25
+  - name: EL
+    versions:
+    - 7
+  galaxy_tags: []
+dependencies: []
```
dgoodwin: Needs a dep on the facts role here?

roles/openshift_provisioning_aws_facts/README.md
dgoodwin: I assume this parallels the ops concept of cluster ID and env? How will these relate to or be used with an inventory as we would have today?

playbooks/provisioning/aws/cloudformation/cloudformation_setup.yml
detiber: The cloudformation stuff was a previous attempt, I'm working on getting that removed in a separate PR to clean this PR up.

roles/openshift_provisioning_aws_vpc/README.md
detiber: No, it is also used for short-circuiting the vpc creation to use an existing vpc. Idea being if the user specifies openshift_provisioning_aws_vpc_id, then we attempt to verify that it already exists and meets the config criteria that we have. If the user has permissions to query vpcs and the vpc config does not meet the criteria for OpenShift, then we will fail with an error message.

roles/openshift_provisioning_aws_vpc/README.md
detiber: It breaks with upstream convention, I worry about confusing users with cryptic names, and further confusion when there are two roles that would have the same abbreviation (openshift_router, openshift_registry are the first that come to mind).

detiber: That said, I did use the abbreviations for variables meant for internal to the role use.

roles/openshift_provisioning_aws_vpc/meta/main.yml
detiber: I purposely left this out, since it goes against the Ops conventions, and opted to enforce the role dependencies at the openshift_provisioning_aws role.

roles/openshift_provisioning_aws_facts/README.md
detiber: Yes, it is meant to mirror the Ops concept of cluster id and env. It is still to be determined how we would handle it with current inventory, but it is something we will have to define through the docs.

detiber:
> Is this an open repo or private? (noticed some repo URLs and such)

It is public

> I assume this is intended for ops as well and will cover their functionality implemented today?

Yes, goal is to start with the VPC and then work our way down through the rest of the infrastructure.

> The tox/molecule stuff looks interesting, will it be possible to use it for actual AWS provisioning testing? (I can't quite see how given the credentials and cleanup required for creating in AWS)

That is what I'm trying to figure out. I plan on letting travis/tox/molecule test the roles that don't require AWS tasks, and then adding testing with the internal ci host for provisioning related tasks. Current thought is that if AWS env vars are present to do the AWS testing (or at least attempt to) otherwise skip those tests.

detiber:
I've broken out the testing and cleanup changes into https://github.com/openshift/openshift-ansible-contrib/pull/237/files, once that merges this PR will be much cleaner

playbooks/provisioning/aws/provision_aws.yml
jstuever: Is it an option to not default the openshift_cluster_id and openshift_env_id?

detiber: I've considered that, at some point we'll want an opinionated deployment that would set defaults, but that can happen at a higher level.

roles/openshift_provisioning_aws_vpc/molecule/default/playbook.yml
```diff
@@ -0,0 +1,13 @@
...
+- hosts: all
+  become: no
+  pre_tasks:
+  - name: Start moto service
+    shell: nohup moto_server ec2 </dev/null >/dev/null 2>&1 &
+    changed_when: false
+  - name: Write boto config
+    copy:
+      src: boto_config
+      dest: /root/.boto
```
detiber: @cooktheryan The pre_tasks here configure the moto ec2 service, and write a boto config that uses the service.

roles/openshift_provisioning_aws_vpc/molecule/default/tests/test_default.py
```diff
@@ -0,0 +1,21 @@
+# import boto
+def test_vpc_exists(Ansible):
```
detiber: @cooktheryan since this is using an Ansible task and the ec2_vpc_net_facts module, I can verify that the vpc is configured the way I expect

playbooks/provisioning/aws/provision_aws.yml
detiber: @stevekuznetsov @jstuever I am going to go ahead and update this to remove the defaults and require they be set explicitly

roles/openshift_provisioning_aws_facts/README.md
detiber: @dgoodwin I think we would require that a user explicitly set the variables, I think we might want to default them in the current byo/config.yml playbook for backward compatibility, but it would need to be explicitly set for new playbooks that would leverage the provisioning.

detiber:
@mwoodson would really like your input on this approach.

kwoodson:
I think there is some value in the tests written with molecule for this PR.  I'm not sure what the plans are for using molecule tests for openshift-ansible.  

The AWS provisioning went into openshift-ansible for 3.7.  The 3.7 release for provisioning is mostly an internal one so that we can adopt our own usage, patterns, and deployment of the AWS objects.  Once we test it, and migrate to it, and harden it we will be recommending others do the same.

mazzystr:
Closing with no merge per pull/868

PR #235, opened 2017-01-26
gce origin deployments

bdurrow:
Here are the changes required to deploy origin to gce using the reference arch gcloud.sh script.  This work is in preparation for the multizone work I hope to do next as I don't have redhat subscriptions available to me.

My config.sh is here:
https://gist.github.com/bdurrow/c6e5a2e1e4e92dc1940f29083d34508e


bdurrow:
Replaced by #243

PR #238, opened 2017-01-27
retain or delete registered image upon revert

cooktheryan:
Adding in the option to retain or remove the registered image created by gce-cli


cooktheryan:
@pschiffe or @themurph can on you give a +1 on this patch?

cooktheryan:
@pschiffe can I get a review of this? Customers may want to retain the RHEL image uploaded to GCE

PR #242, opened 2017-01-30
Enable containerized deployments in GCE

bdurrow:
Containerized deployments currently fail because master and node containers don't have gce.ini mounted into the container.  It looks like openshift-ansible handles this by volume mounting /etc/origin/cloudprovider when openshift_cloudprovider_kind is defined.


cooktheryan:
Can you remove the lin in openshift-setup.yaml?
```
    containerized: false
```

It should be covered my your yaml update

reference-architecture/gce-ansible/playbooks/openshift-setup.yaml
```diff
@@ -9,19 +9,20 @@
       cloud-provider:
       - "gce"
       cloud-config:
-      - "/etc/origin/gce.ini"
+      - "/etc/origin/cloudprovider/gce.conf"
```
detiber: Instead of configuring osm_controller_args for the cloud provider, should be using the openshift-ansible provided config.

detiber: For now, you'll probably need to update openshift_hostname with the short hostname to ensure everything works.

cooktheryan:
Might want to merge in the conflicts then we can test

cooktheryan:
@bdurrow if you fix this one up it can be my next test

cooktheryan:
in the gcloud.sh file the following needs added near line number 675

```
OS_CONTAINERIZED
```

cooktheryan:
@bdurrow I am going to close this due to all of the changes currently taking place with the GCE deployment. Containerizing GCE will not be forgotten though.

PR #243, opened 2017-01-30
Allow for CentOS Images and parametarize many deployment options.

bdurrow:


reference-architecture/gce-cli/gcloud.sh
```diff
@@ -360,24 +383,38 @@ if ! gcloud --project "$GCLOUD_PROJECT" compute project-info describe | grep -q
...
-            --enable=\"rhel-7-server-extras-rpms\" \
-            --enable=\"rhel-7-server-ose-${OCP_VERSION}-rpms\";
+        if which subscription-manager; then
+            subscription-manager register --username=${RH_USERNAME} --password=\"${RH_PASSWORD}\";
+            subscription-manager attach --pool=${RH_POOL_ID};
+            subscription-manager repos --disable=\"*\";
+            subscription-manager repos \
+                --enable=\"rhel-7-server-rpms\" \
+                --enable=\"rhel-7-server-extras-rpms\" \
+                --enable=\"rhel-7-server-ose-${OCP_VERSION}-rpms\";
```
detiber: Is subscription-manager installed on CentOS hosts that are registered to a Katello system?

bdurrow: Based on what I read [here](https://fedorahosted.org/katello/wiki/GuideSystemRegistrationClient) it looks like it is.  I won't have a way to test it but should I use subscription-manager to enable these repos on centos boxes if it is available  or just use different detection logic to determine what kind of system I am on?

detiber: @bdurrow woudl probably be better to key off of the deployment_type if possible.

pschiffe:
This PR is now outdated and the support for centos and origin was added in #483

Issue #96, opened 2016-11-05
Instruction should contain ntp installation and setup


tahonen:
When doing installation from vagrant image etc. time might not be synced correctly. This may cause python errors that are really hard to figure out.

detiber:
@tahonen we should be enabling either ntpd or chronyd at install time, did you hit an issue with this?

tahonen:
Installed hosts are running fine and in correct time. I had time issue with vm that I used to install. I used  vagrant and clock was no in sync after host laptop when to sleep. When clock was no sync I got very weird errors from python code accessing AWS. Installed aws cli and noted ant SSL was failing due time problems.

dwojciec:
@tahonen on my vagrant project here: https://github.com/dwojciec/ocp-on-aws  I setup ntp. I include Vagrant-timezone and I setup ntp inside setup.sh script. By doing that I'm able to rerun the provisonning later when I want.

PR #245, opened 2017-01-31
Vagrant fixes

chekolyn:
1) Added ntp
2) Fix glitch with vagrant and eth1 not having an active ip after first boot.

### DIFF
--- a/vagrant/provision/setup.sh
+++ b/vagrant/provision/setup.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+
+# Capture network variable:
+NETWORK_BASE=$1
+sudo -i
+
+echo "Check eth1 ip config for netbase: $NETWORK_BASE"
+
+if ip a show dev eth1 | grep -q "inet $NETWORK_BASE"; then
+  echo "eth1 ip detected"
+else
+  echo "eth1 missing ip; restaring interface"
+  ifdown eth1 && ifup eth1
+fi
+
+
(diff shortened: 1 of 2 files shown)


vagrant/provision/setup.sh
```diff
@@ -0,0 +1,21 @@
...
+else
+  echo "eth1 missing ip; restaring interface"
+  ifdown eth1 && ifup eth1
+fi
+
+
+echo "Install ntp server to avoid to have desync issues"
+yum -y install ntp 
+systemctl enable ntpd
+systemctl start ntpd
```
detiber: hmm, I would expect that the openshift install would be taking care of configuring ntpd/chronyd, is that not the case?

chekolyn: I think you are right. I was looking at this issue https://github.com/openshift/openshift-ansible-contrib/issues/96 but as you mentioned ansible already handles [this](https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_clock/tasks/main.yaml). 

Since the ntp portion is unnecessary I'll remove it from the pull request; the eth1 vagrant issue is legit.

spicysomtam:
This change breaks my attempt in using vagrant2 using rhel7.4/openshift 3.6 ocp, because the network interfaces are enp0 devices rather than eth devices. I commented out the line you added in the Vagrantfile to work round the issue.

PR #262, opened 2017-02-16
git is required when running containerized

cooktheryan:
When running the validation testing in a containerized environment git is not installed which causes issues when attempting to deploy the validation project.

### DIFF
--- a/playbooks/prerequisite.yaml
+++ b/playbooks/prerequisite.yaml
@@ -14,7 +14,7 @@
   - prerequisites
 
 - hosts: master
-  gather_facts: no
+  gather_facts: yes
   become: yes
   roles:
   - master-prerequisites
--- a/playbooks/prerequisite.yaml
+++ b/playbooks/prerequisite.yaml
@@ -12,3 +12,9 @@
   roles:
   - rhsm-repos
   - prerequisites
+
+- hosts: masters 
+  gather_facts: no
+  become: yes
+  roles:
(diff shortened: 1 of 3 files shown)


cooktheryan:
aos-vmw-ci

cooktheryan:
aos-vmw-test

cooktheryan:
aos-ci-test

roles/master-prerequisites/tasks/main.yaml
```diff
@@ -0,0 +1,9 @@
+---
+- block:
+  - name: Install git
+    package:
+      name: "{{ item }}"
+      state: latest
+    with_items:
+    - git
```
detiber: Since this is only a single package, it might make sense to drop the loop.

The block seems unnecessary as well, since there is only the single task.

playbooks/prerequisite.yaml
```diff
@@ -12,3 +12,9 @@
   roles:
   - rhsm-repos
   - prerequisites
+
+- hosts: master
+  gather_facts: no
```
detiber: One thing to keep in mind here is that the package module makes use of facts, so If for some reason it ends up being possible to call just this play (through tags), then the package task will fail for undefined variables.

cooktheryan:
aos-vmw-test

cooktheryan:
aos-ci-test

PR #264, opened 2017-02-17
adding KubernetesCluster variable

cooktheryan:
Missing variable in AWS

### DIFF
--- a/reference-architecture/aws-ansible/playbooks/roles/cloudformation-infra/files/add-infra-node.json
+++ b/reference-architecture/aws-ansible/playbooks/roles/cloudformation-infra/files/add-infra-node.json
@@ -10,7 +10,7 @@
     "AmiId": {
       "Type": "AWS::EC2::Image::Id"
     },
-    "ExistingStack,": {
+    "ExistingStack": {
       "Type": "String"
     },
     "InstanceType": {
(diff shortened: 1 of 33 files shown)


reference-architecture/aws-ansible/playbooks/roles/cloudformation-infra/files/add-infra-node.json
```diff
@@ -10,6 +10,9 @@
     "AmiId": {
       "Type": "AWS::EC2::Image::Id"
     },
+    "ExistingStack": {
```
detiber: Since There could be multiple existing stacks, should this be `PrimaryStack` instead?

detiber: You could also use the [cloudformation_facts module](http://docs.ansible.com/ansible/cloudformation_facts_module.html) to validate the stack already exists. It could also help to provide defaults for the new stack based on the primary stack.

reference-architecture/aws-ansible/playbooks/roles/cloudformation-infra/files/brownfield-byo-bastion.json
```diff
@@ -755,6 +755,9 @@
             { "Key": "Name",
               "Value": {"Fn::Join": [".", ["ose-master01",{"Ref": "PublicHostedZone"}]]}
             },
+            { "Key": "KubernetesCluster",
```
detiber: Since this key is being added to all of the hosts, instead of setting it on each host, you can set it for the entire stack using the `tags` param with the cloudformation module

cooktheryan:
aos-ci-test

cooktheryan:
@detiber excellent suggestion. It made the change 100 times easier

cooktheryan:
aos-ci-test

PR #272, opened 2017-02-28
Changed registered to patched

cooktheryan:
Change of variable name within config.sh.example and gcloud.sh to better reflect the image


pschiffe:
We can drop this PR now, the name is already updated to the `GOLD` image in the gce-cli.

PR #277, opened 2017-03-04
make the installer setup the registry

cooktheryan:
Trying to use the installer to configure the registry rather than have our scripts do the work


cooktheryan:
aos-vmw-test

cooktheryan:
Code is in upstream but not in official rpms yet

cooktheryan:
@dav1x I'm going to roll back the whole /exports things but I think we are onto something good here after testing today on 3.4

dav1x:
love it. removed more code than added. lgtm

reference-architecture/vmware-ansible/ocp-on-vmware.py
```diff
@@ -519,7 +523,7 @@ def launch_refarch_env(console_port=8443,
     rhsm_password=%s \
     rhsm_activation_key=%s \
     rhsm_org_id=%s \
-    rhsm_pool=%s \
+    rhsm_pool="%s" \
```
cooktheryan: @dav1x added the quotes because sub manager was attaching every pool with the word "Red" because of the space between Red Hat

roles/rhsm-subscription/tasks/main.yaml
```diff
@@ -30,6 +35,6 @@
         password: "{{ rhsm_password }}"
         state: present
         pool: "{{ rhsm_pool }}"
-      when: rhsm_user is defined and rhsm_user
+      when: "'Current' not in subscribed.stdout and rhsm_user is defined and rhsm_user"
```
cooktheryan: @pschiffe and @dav1x I think this should work for skipping the step rather than Subscribed where we ran into "Not Subscribed"

PR #282, opened 2017-03-13
Migrating to deployment manager, part 1

pschiffe:
This is a first step in migrating the gcp ref. arch. away from shell script to
Google's gcp deployment manager and ansible. Most of the resources is now
migrated, more complicated resources will be migrated later. Naming was redone
- before, it was possible to choose arbitrary name for almost any resource -
now it's done by changing the 'prefix'. This way, it should be possible to
deploy multiple clusters under single account by choosing different prefix.

### DIFF
--- a/reference-architecture/gce-cli/config.sh.example
+++ b/reference-architecture/gce-cli/config.sh.example
@@ -4,11 +4,11 @@
 # Path to a RHEL image on local machine, downloaded from Red Hat Customer Portal
 RHEL_IMAGE_PATH="${HOME}/Downloads/rhel-guest-image-7.3-35.x86_64.qcow2"
 
-# Choose to delete or retain the unregistered image
-DELETE_IMAGE=true
+# Choose to delete or retain the clean image
+DELETE_IMAGE='false'
 
-# Choose to delete or retain the registered image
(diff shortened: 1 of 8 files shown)


reference-architecture/gce-cli/config.sh.example
```diff
@@ -4,11 +4,11 @@
...
 RHEL_IMAGE_PATH="${HOME}/Downloads/rhel-guest-image-7.3-35.x86_64.qcow2"
 
-# Choose to delete or retain the unregistered image
-DELETE_IMAGE=true
+# Choose to delete or retain the clean image
+DELETE_IMAGE='false'
 
-# Choose to delete or retain the registered image
-REGISTERED_IMAGE_DELETE=true
+# Choose to delete or retain the golf image
```
cooktheryan: Just quick spelling NIT from golf to gold

cooktheryan:
@pschiffe the code looks good. Do you need to make any changes to the ref arch before this is merged? If not will merge after I get the ok from you

pschiffe:
No, not yet. Most of the changes are done under the hood, user shouldn't notice. Once the migration is more complete, I've update the paper.

PR #286, opened 2017-03-20
Handle non standard OCP_PREFIX with GCE setup

dhananjaysathe:
If someone sets a non-default OCP_PREFIX the following stages such as r/reference-architecture/gce-ansible/playbooks/create-inventory-file.yaml and the /reference-architecture/gce-ansible/playbooks/roles/instance-groups/tasks/main.yaml that rely on these groups fail. This patch takes an approach of handling this at the inventory generation phase to account for a non-standard prefix.

### DIFF
--- a/reference-architecture/gce-ansible/inventory/gce/hosts/gce.py
+++ b/reference-architecture/gce-ansible/inventory/gce/hosts/gce.py
@@ -330,6 +330,7 @@ class GceInventory(object):
             else: groups[zone] = [name]
 
             tags = node.extra['tags']
+            tag_prefix = os.environ.get('OCP_PREFIX', 'ocp')
             for t in tags:
                 if t.startswith('group-'):
                     tag = t[6:]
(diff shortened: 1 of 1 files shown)


dhananjaysathe:
An alternate approach is to pass this variable into ansbile-config then have ansible convert the groups based on this. I am not an expert in ansible on how I parse and rename the groups within its syntax so this patch tries to deal with that issue at an earlier phase. Yielding the same result, albeit some extra groups are generated as a small overhead.

pschiffe:
Thanks for the PR, I'll look into it. We are currently in the rewrite process of the gcp ref arch - going away from shell to deployment manager and ansible. `OCP_PREFIX` is also a new stuff, which needs to be yet incorporated into relevant ansible playbooks.

dhananjaysathe:
@pschiffe I see you have pr 282 merged , is there a piece i am missing or should wait for I want to go ahead deploy a fresh cluster. If you have something on staging and would want me to review/try i'd be more than glad to

pschiffe:
@dhananjaysathe give me a couple days to migrate the rest and check all the playbooks. I'll let you know once it's ready.

pschiffe:
Sorry it took so long :-( but PR #356 should fix this problem.

PR #288, opened 2017-03-20
static inventory resolves #285

### LINKED ISSUE #285: Inventory file doesn't exist after reference architecture is deployed on AWS
After the reference architecture is deployed on EC2 using ose-on-aws.py no inventory file is created, I believe this is important as one might want to modify the cluster using the playbooks from openshift-ansible, according to the advanced installation docs.

Even the AWS reference architecture instructions README have a teardown section that needs an inventory file: https://github.com/openshift/openshift-ansible-contrib/tree/master/reference-architecture/aws-ansible#teardown

Is there an easy way someone can generate the inventory file after the install or should the installer be modified to create this file?

cooktheryan:


cooktheryan:
@dav1x if you get the cycles can you review this?

dav1x:
lgtm

PR #300, opened 2017-03-24
Migrating to ansible, next part

pschiffe:


reference-architecture/gce-cli/ansible/playbooks/roles/rhel-image/defaults/main.yaml
```diff
@@ -0,0 +1,9 @@
+---
+rhel_image_path: '{{ ansible_env.HOME }}/Downloads/rhel-guest-image-7.3-35.x86_64.qcow2'
```
cooktheryan: Only small thing would be possibly removing or variablizing /Downloads/

cooktheryan: Not a required fix but just a potential for some confusion

pschiffe: Well, this is just a default for variable `rhel_image_path`.. It can be redefined by user altogether. Also, when the user downloads the image with browser, it will be stored in `Downloads` folder by default.

cooktheryan: That's fine. Once it passes travis Ill give it one more look and then merge or ping you. Thanks!!

pschiffe: Also I should note, that later this variable will be probably moved to main config file, similar as currently `config.sh.example`, so it will have better visibility.

Issue #181, opened 2017-01-13
[refarch/osp-cli] Reduce host name variables in scripts to those needed.


markllama:
Several scripts that use only the NODE list also define the BASTION and MASTER lists.  This obscures the values in use, adds unneeded code and increases the size of the doc.

Each script which works on multiple hosts should define only the the values needed.

Issue #180, opened 2017-01-13
[refarch/osp-cli] Add documentation for proxy host SSH access and sudo.


markllama:
The documentation mentions the need for passwordless SSH access to the haproxy host, but does not detail how to set a variable for the SSH key.

Issue #179, opened 2017-01-13
[refarch/osp-cli] Remove unneeded line in ch4.6_generate_dns_updates.sh


markllama:
Line 5: HOSTNAME=$1

This value is unused in the script.  Remove it.

markllama:
Removing that line and several others that are extraneous.

Issue #178, opened 2017-01-13
[refarch/osp-cli] DOMAIN setup variable duplicates/conflicts with other DNS related vars


markllama:
The variables that define the host naming internally and externally sometimes duplicate or conflict.

DOMAIN can duplicate with OCP3_DNS_PREFIX among others.

Clean up input variables and de-duplicate.

markllama:
DOMAIN and OCP3_BASE_DOMAIN use are unclear and confusing.  The control net name should be derived from the base domain.

IE `DOMAIN=control.${OCP3_BASE_DOMAIN}`

Possibly rename DOMAIN as well.

PR #306, opened 2017-03-30
adding scripts dir to osp-cli dir

rlopez133:
This contains the 'openstack' commands, however, some scripts here are older and are not called within the paper so they aren't changed. We can remove them as needed.

### DIFF
--- a/reference-architecture/osp-cli/scripts/all-hosts.sh
+++ b/reference-architecture/osp-cli/scripts/all-hosts.sh
@@ -3,7 +3,7 @@
 # Execute these preparation steps on all hosts
 # Run on the bastion host
 #
-OCP_VERSION=${OCP_VERSION:-3.2}
+OCP_VERSION=${OCP_VERSION:-3.4}
 
 MASTERS="master-0 master-1 master-2"
 INFRA_NODES="infra-node-0 infra-node-1"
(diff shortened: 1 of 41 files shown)


cooktheryan:
Resolves issues #181 #180 #179 #178 #177 #173

reference-architecture/osp-cli/scripts/all-hosts.sh
```diff
@@ -0,0 +1,93 @@
+#!/bin/sh
+#
+# Execute these preparation steps on all hosts
+# Run on the bastion host
+#
+OCP_VERSION=${OCP_VERSION:-3.2}
```
cooktheryan: Should this be 3.4?

reference-architecture/osp-cli/scripts/bastion_host.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/sh
+
+OSP_VERSION=${OSP_VERSION:-8}
+OCP_VERSION=${OCP_VERSION:-3.2}
```
cooktheryan: Should these values be updated as well

reference-architecture/osp-cli/scripts/install_docker.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+for H in $INFRA_NODES $APP_NODES
+do
+  ssh $H sudo yum install -y docker
```
cooktheryan: I would remove this and let the installer take care if it by default

PR #311, opened 2017-04-05
Drive the DNS installation completely with Ansible

tomassedovic:
This replaces the current glue code and inventory generation with a playbook that does everything: deploys the Heat stack, configures the inventory and installs the DNS.

All configuration now happens with Ansible vars and the deployment now consists of a single `ansible-playbook` invocation.


tomassedovic:
@markllama would appreciate your review since this touches the code you wrote.

PR #312, opened 2017-04-07
Migrating to ansible, continued

pschiffe:


cooktheryan:
@pschiffe  just making sure but does the redhat_subscriptipn.py handle the pool id and pool name ? I think with the library that currently exists it can handle the both.

pschiffe:
@cooktheryan currently we are using our custom modified version of redhat_subscriptipn.py in library folder, which handles both. Version which will come with ansible 2.3 will also handle both (pool id and name).

PR #317, opened 2017-04-18
timeout increase for subscription-manager

cooktheryan:


cooktheryan:
@pschiffe if you have the cycles I think this will make rhsm subscriptions a little safer

/cc @dav1x

PR #319, opened 2017-04-19
Attach subscription during provisioning.

### LINKED ISSUE #318: RHEL Provisioning Fails
Because of the changes made in commit [bd1e2b2](https://github.com/openshift/openshift-ansible-contrib/commit/bd1e2b29b1a69c09cb85cb8b69fe190d9f1c9a23), Ansible no longer attaches the provided subscription (variable `rhsm_pool`) to RHEL machines. These tasks are skipped because vagrant-registration registers the VMs prior to provisioning. As a result, the role `rhsm-repos` fails with "rhel-7-server-ose-3.4-rpms is not a valid repository ID."

ghyde:
Fixes #318

### DIFF
--- a/roles/rhsm-subscription/tasks/main.yaml
+++ b/roles/rhsm-subscription/tasks/main.yaml
@@ -38,4 +38,25 @@
         pool: "{{ rhsm_pool }}"
       when: "'Current' not in subscribed.stdout and rhsm_user is defined and rhsm_user"
 
+    - name: Check if subscription is attached
+      command: subscription-manager list --consumed --pool-only --matches="{{ rhsm_pool }}"
+      register: subscription_attached
+      changed_when: no
+
+    - block:
+        - name: Get pool id
+          shell: subscription-manager list --available --pool-only --matches="{{ rhsm_pool }}" | head -n 1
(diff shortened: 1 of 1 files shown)


ghyde:
@cooktheryan would you take a look at this?

cooktheryan:
@ghyde ill take a look at this. I need to brush off some vagrant cobwebs

ghyde:
@cooktheryan Vagrant doesn't really apply to this patch.

What's happening is `vagrant-registration` registers a VM during a `vagrant up`, but it attaches the VM to the first available subscription. Your [patch](https://github.com/openshift/openshift-ansible-contrib/commit/bd1e2b29b1a69c09cb85cb8b69fe190d9f1c9a23) then sees that the VM is registered, but it doesn't check the subscription. As a result, I sometimes get a subscription that doesn't contain the `rhel-7-server-ose-3.4-rpms` repo, and the playbook fails.

This patch ensures that the correct subscription, provided by the variable `rhsm_pool`, is attached before the playbook attempts to enable OpenShift repos.

cooktheryan:
Makes sense

PR #328, opened 2017-04-25
Using the shared validation plays

cooktheryan:


pschiffe:
This PR doesn't solve the issues with the validation plays, I'm working on the fix.

pschiffe:
Replaced by #343

PR #332, opened 2017-04-28
Replace rhsm_* with rhel_subscribe

dav1x:
- Modified add-node.py and ocp-on-vmware.py to use the same variable names as rhel_subscribe in openshift-ansible
- Modified all the plays to point to that role
- Added the roles from ansible -  /usr/share/ansible/openshift-ansible/roles to role_path


cooktheryan:
aos-vmw-test

cooktheryan:
aos-vmw-test

PR #337, opened 2017-05-02
remove ssh keys for new instances on heat stack create

markllama:
On creating new servers in a heat stack, remove any old SSH keys from the localhost known_hosts file.
This will prevent erroneous messages from SSH when ansible attempts to log into the instances for the first time.

### DIFF
--- a/reference-architecture/osp-dns/deploy-dns.yaml
+++ b/reference-architecture/osp-dns/deploy-dns.yaml
@@ -69,6 +69,13 @@
       ansible_user: "{{ ssh_user }}"
     with_items: "{{ stack_output.slaves }}"
 
+  - name: Clear saved SSH keys
+    known_hosts:
+      name: "{{ item }}"
+      state: absent
+    with_items: "{{ groups['all'] }}"
+    when: stack_check.rc != 0
+
   - name: Wait for the deployed servers
     wait_for:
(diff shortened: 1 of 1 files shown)


tomassedovic:
I'm not sure it's a good idea to update the local known_hosts file. I'd prefer if Ansible didn't modify the local environment at all.

I've always seen either setting the `ANSIBLE_HOST_KEY_CHECKING=False` environment variable or [`host_key_checking` in ansible.cfg](http://docs.ansible.com/ansible/intro_configuration.html#host-key-checking) and I assumed it's the standard practice.

What do you think about documenting the above in the readme instead?

cooktheryan:
I +1 Tomas's suggestion. Just in event someone needs to keep other entries.

Ryan


Ryan

markllama:
It's only re-setting the address which OSP has re-assigned, and only during stack create.  If you follow the "don't remove it" logic, we should be clearing those entries when the stack is removed as well.

tomassedovic:
We should clear them only if we set them (e.g. when it's not overridden with `ssh_args` in their ansible.cfg or ~/.ssh/config). But this is standard SSH behaviour and Ansible just reflects that.

markllama:
Withdrawn.

PR #348, opened 2017-05-10
[WIP] Fixes #347

### LINKED ISSUE #347: Expose some more variables in the roles
As a user, I want to avoid modifying the ansible roles to modify some parameters currently hardcoded. Also, when something like some additional repo is required, it would be easier to modify just the variable or use the extra-vars command in ansible.

e-minguez:
Also:
- The openshift_node_local_quota_per_fsgroup variable was explicitly set but
not used anywhere
- Moved the docker-storage-setup file to a template to support variables
- Use grep -c instead grep | wc -l


roles/openshift-emptydir-quota/tasks/main.yaml
```diff
@@ -3,7 +3,7 @@
   replace:
     dest: /etc/origin/node/node-config.yaml
     regexp: '^(.*)perFSGroup: (\s+.*)?$'
-    replace: '\1 perFSGroup: 512Mi\2'
+    replace: '\1 perFSGroup: {{ openshift_node_local_quota_per_fsgroup }}\2'
```
cooktheryan: This is done during the openshift install on aws. I believe we should deprecate role for gce/vmw and let the installer handle it.

reference-architecture/aws-ansible/playbooks/openshift-setup.yaml
```diff
@@ -25,12 +25,11 @@
     openshift_hosted_router_replicas: 3
     openshift_hosted_registry_replicas: 3
     openshift_master_cluster_method: native
-    openshift_node_local_quota_per_fsgroup: 512Mi
```
cooktheryan: I don't believe we should remove the line but change the value to be a variable

reference-architecture/aws-ansible/playbooks/node-setup.yaml
```diff
@@ -24,7 +24,6 @@
     openshift_router_selector: "role=infra"
     openshift_hosted_router_replicas: 3
     openshift_hosted_registry_replicas: 3
-    openshift_node_local_quota_per_fsgroup: 512Mi
```
cooktheryan: Keep the value and change to 512Mi to a variable

reference-architecture/aws-ansible/playbooks/roles/inventory-file-creation/files/inventory
```diff
@@ -17,7 +17,6 @@ openshift_router_selector="role=infra"
 openshift_hosted_router_replicas=3
 openshift_hosted_registry_replicas=3
 openshift_master_cluster_method=native
-openshift_node_local_quota_per_fsgroup=512Mi
```
cooktheryan: Let this in but as a variable

roles/openshift-emptydir-quota/tasks/main.yaml
e-minguez: It looks like it is in the installer https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_node/templates/node.yaml.v1.j2#L58 so yeah, I agree on deprecate it on every provider

reference-architecture/aws-ansible/playbooks/openshift-setup.yaml
e-minguez: As it is in the installer and the value is the default one, I think it safe to remove it and add support for extra-vars when calling the playbooks/openshift-install.yaml playbook. What do you think?

roles/openshift-emptydir-quota/tasks/main.yaml
cooktheryan: @e-minguez in this pull then we will need to include the other providers openshift install vars file

reference-architecture/aws-ansible/playbooks/openshift-setup.yaml
cooktheryan: If its not defined though like above will the value be defined ?

roles/openshift-emptydir-quota/tasks/main.yaml
e-minguez: I'll try to fix it

reference-architecture/aws-ansible/playbooks/openshift-setup.yaml
e-minguez: You're right, if it is not defined the default value is none https://github.com/openshift/openshift-ansible/blob/81cf5123366814d098b7ca169a6e4deb2759507d/roles/openshift_node/tasks/main.yml#L33 , I'll try to fix it

e-minguez:
I'll close this PR in order to split it in different commits

PR #357, opened 2017-05-14
let the installer handle excluder

cooktheryan:
Removal of excluder installation and configuration as the installer now does this.


cooktheryan:
aos-ci-test

PR #361, opened 2017-05-15
Add subscribe_rhel to only vmw

dav1x:
-Reverted the rhel_subsribe to work for VMware
-Added compatibiilty with the shared roles
-Fixed playbook issue caused by add-node pr


cooktheryan:
aos-vmw-test

cooktheryan:
Lgtm

PR #362, opened 2017-05-16
Add ansible_ssh_private_key_file to the SSH test

### LINKED ISSUE #355: [osp-dns] - playbook should accept --key-name or --private-key for ssh to the DNS host.
when running the playbook with --key-name or --private-key and the file path to the private key file associated with the openstack keypair, ansible fails to log in to the new DNS server.  Allowing ansible to proceed requires placing the private key file in `.ssh/id_rsa` so that ansible's call to SSH uses the default private key.

tomassedovic:
We're waiting for the OpenStack VMs to be able to SSH into before
proceeding with the DNS playbook (because until 2.3 Ansible doesn't
handle that gracefully).

But if the private key for the SSH connection was only specified using
the `--private-key` argument to `ansible-playbook`, the test was
failing because it didn't take that possibility into the account.

Now it does.

Fixes #355

### DIFF
--- a/reference-architecture/osp-dns/deploy-dns.yaml
+++ b/reference-architecture/osp-dns/deploy-dns.yaml
@@ -75,9 +75,15 @@
       port: 22
     with_items: "{{ groups['all'] }}"
 
+  # NOTE(shadower): This is necessary until we switch to Ansible 2.3,
+  # which has the `wait_for_connection` module for this exact purpose.
+  # https://docs.ansible.com/ansible/wait_for_connection_module.html
   - name: Verify the servers can be SSHd into
     command: >
       {{ansible_ssh_executable}} {{ansible_ssh_common_args}}
+      {% if ansible_ssh_private_key_file|default(None) %}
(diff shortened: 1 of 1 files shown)


pschiffe:
For this purpose, I'm using this with ansible 2.2. Wouldn't that be also applicable to you?
```
  - name: wait for the temp instance to come up
    wait_for:
      host: '{{ hostvars[prefix + "-tmp-instance"]["ansible_ssh_host"] }}'
      port: 22
      state: started
```

tomassedovic:
Thanks @pschiffe! I've tried `wait_for` (indeed, it's in there, too), but it doesn't always work -- as in, you can end up in a situation where sshd is up, but the login fails.

I think what happens is that that it's cloud-init's job to set up the pubkey to `authorized_hosts` and sometimes sshd comes up before that's done -- so you can connect to the port but the authentication will fail because the key is not in the right place yet.

Which is why there is a task that explicitly tries to SSH in.

PR #370, opened 2017-05-18
safer subscriptions until we move to openshift-ansible

cooktheryan:
This is the safer route to subscribe a server to rhsm. This will be the stop gap until the next openshift release which should have a fix included for us.

### DIFF
--- a/playbooks/add-node-prerequisite.yaml
+++ b/playbooks/add-node-prerequisite.yaml
@@ -4,7 +4,7 @@
   become: yes
   serial: 1
   roles:
-    - rhsm-subscription
+    - aws-rhsm-subscription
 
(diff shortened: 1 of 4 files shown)


cooktheryan:
aos-ci-test

cooktheryan:
aos-ci-test

pschiffe:
@cooktheryan why only for aws?

cooktheryan:
@pschiffe I would need to write the code to be able to handle the sub pool id rather than the sub pool name

pschiffe:
Ah, now I see that you are switching from the pool id to the pool name which is provided by user in aws. OK then, as it's temporary.

Issue #377, opened 2017-05-19
AWS and Atomic Host Error


christianh814:
Using the `./ose-on-aws.py` command with the `--containerized=True` flag returns the following (this was done with `ami-67389007` in west-2 )

```
fatal: [ose-infra-node01.breakne.ws]: FAILED! => {
    "assertion": "l_atomic_docker_version.stdout | replace('\"', '') | version_compare('1.12','>=')",
    "changed": false,
    "evaluated_to": false,
    "failed": true
}

MSG:

Installation on Atomic Host requires Docker 1.12 or later. Please upgrade and restart the Atomic Host.
```

Logging onto the instance  you see

```
[root@ip-10-20-1-8 ~]# docker info
Containers: 0
 Running: 0
 Paused: 0
 Stopped: 0
Images: 0
Server Version: 1.10.3
```

The script should probably issue an `atomic host upgrade` to the proper version.

cooktheryan:
This has been fixed

PR #379, opened 2017-05-22
Atomic docker version

cooktheryan:
Work will patch the atomic instance if it does not meet the correct version. Will resolve issue #377

### DIFF
--- a/roles/prerequisites/tasks/main.yaml
+++ b/roles/prerequisites/tasks/main.yaml
@@ -81,4 +81,3 @@
     when: patched.changed
 
   when: openshift.common.is_atomic | bool
-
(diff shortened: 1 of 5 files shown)


roles/prerequisites/tasks/main.yaml
```diff
@@ -50,3 +50,34 @@
...
+
+  - name: Reboot when patched
+    shell: sleep 5 && shutdown -r now "Reboot due to Atomic Patching"
+    async: 1
+    poll: 0
+    ignore_errors: true
+    when: patched.changed
+
+  - name: Wait for hosts to be back
+    pause:
```
pschiffe: `wait_for` module could be more suitable here

PR #395, opened 2017-05-24
Ref. Arch. says gquota instead grpquota

e-minguez:


cooktheryan:
@e-minguez can you verify grpquota would work rather than gquota

e-minguez:
It was qrpquota before, I've modified to be "gquota" like every other ref. arch. Reading the man[1], it should be the same:
```
gquota/grpquota/gqnoenforce
              Group disk quota accounting enabled and limits (optionally) enforced.  Refer to xfs_quota(8) for further details.
```

[1]https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/filesystems/xfs.txt?id=HEAD#n181

cooktheryan:
@e-minguez nothing like a bunch of names to do the same things. Excellent find

e-minguez:
👍

PR #397, opened 2017-05-24
Add an Openstack provider

tomassedovic:
This adds playbooks that create an environment ready for
openshift-ansible on OpenStack.

This commit just moves code over from casl-ansible:

https://github.com/redhat-cop/casl-ansible

I've tested this against OSP 10 with OCP 3.5 and RHEL 7 guests in a 1
master, 1 infra & 2 app nodes scenario with LDAP auth.

To my knowledge, multi-master, OCP 3.4, centos guests and non-Red Hat
openstack should all work but I didn't get to test them with these
playbooks yet.

## Notes & Questions

The playbooks require Ansible 2.3 and shade installed. Is that a problem?

The initial inventory setup is a bit fiddly -- I'd like to streamline it, and will appreciate any suggestions (check out the readme and the attached sample inventory). I'm not against writing a tailored dynamic inventory if need be.

The playbook var names come from casl-ansible. Are they okay or do we want them changed? Since they form the public interface, better settle on the final ones now than changing them later when people use this.

This is an initial work -- I'd like to merge this as a functioning base that we can make more flexible and useful (say by integrating openstack's native lbaas etc.) later.

Oh also, it comes with its own role for doing Red Hat subscriptions because we clearly don't have enough of those. We'll probably want to consolidate those at some point.


cooktheryan:
@tomassedovic no issues with ansible 2.3 and shade especially if we call it out in the readme

playbooks/provisioning/openstack/server-setup.yaml
```diff
@@ -0,0 +1,72 @@
...
+  pre_tasks:
+  - include: pre_tasks.yml
+  roles:
+  - role: hostnames
+
+# Subscribe Hosts
+- hosts: cluster_hosts
+  become: true
+  roles:
+  - role: subscription-manager
```
cooktheryan: this is probably incredibly premature request but it may be worth looking into using the rhel_subscribe role

/usr/share/ansible/openshift-ansible/roles/rhel_subscribe/

oybed: agreed - I believe we need to have a bigger discussion around what components will move over from CASL to contrib. I'll add it to the agenda for this Friday's meeting.

roles/docker/tasks/main.yml
```diff
@@ -0,0 +1,41 @@
+---
+- name: "Setting Docker Facts"
+  set_fact:
+    docker_storage_block_device: "{{ docker_storage_block_device | default(default_docker_storage_block_device) }}"
+    docker_storage_volume_group: "{{ docker_storage_volume_group | default(default_docker_storage_volume_group) }}"
+
+- name: "Install Docker"
```
cooktheryan: If this role is used exclusively for docker with OpenShift I would suggest just placing the docker-storage-setup file and then letting the OpenShift installer install docker and start the service. That way we potentially don't install a version that may be too new

roles/openstack-stack/tasks/main.yml
```diff
@@ -0,0 +1,41 @@
+---
+- name: create HOT stack template prefix
+  register: stack_template_pre
+  tempfile:
+    state: directory
+    prefix: casl-ansible
```
cooktheryan: We can accept this as is with the casl-ansible tag but in the future we may want to make this a configurable variable

roles/openstack-stack/templates/user_data.j2
```diff
@@ -0,0 +1,13 @@
+#cloud-config
+disable_root: true
+
+system_info:
+  default_user:
+    name: openshift
+    sudo: ["ALL=(ALL) NOPASSWD: ALL"]
+
```
cooktheryan: Potentially we could also put the docker-storage-setup file in user_data 

write_files:
- content: |
    DEVS='/dev/xvdb'
    VG=docker_vol
    DATA_SIZE=95%VG
    EXTRA_DOCKER_STORAGE_OPTIONS="--storage-opt dm.basesize=3G"
  path: /etc/sysconfig/docker-storage-setup
  owner: root:root

Also, if we set the openshift_node_local_quota_per_fsgroup during install we could place the required fstab options here

roles/docker/tasks/main.yml
oybed: yes, this is what should happen for contrib and what we have been moving towards in CASL, implementation is just a bit behind

playbooks/provisioning/openstack/server-setup.yaml
etsauer: we would have to heavily enhance the rhel_subscribe role, as it doesn't deal with satellite.

cooktheryan: sounds good to me. We can say this is out of scope

tomassedovic:
Thanks for the feedback! I'll update the PR such that it preserves the history from casl-ansible and base the rest of my work on top of it.

playbooks/provisioning/openstack/pre_tasks.yml
```diff
@@ -0,0 +1,37 @@
...
+
+- name: Setting Common Facts
+  set_fact:
+    env_id: "{{ env_id | default(default_env_id) }}"
+  delegate_to: localhost
+
+- name: Set Dynamic Inventory Filters
+  become: false
+  shell: >
+    export OS_INV_FILTER_KEY=clusterid && OS_INV_FILTER_VALUE={{ env_id }}
```
bogdando: the latter seems missing the ``export`` cmd

tomassedovic:
Please ignore the commit above for now -- it's just a workaround I need for a public cloud I'm using that doesn't like having DNS recusive open to everyone. We'll want to tighten this up later and I need to update the pull request with the history of the original CASL roles & playbooks first.

tomassedovic:
Okay, I've just updated the pull request to include the history of the copied roles and playbooks.

playbooks/provisioning/openstack/pre_tasks.yml
bogdando: I double-checked on my pc, and it seems the second part will not get exported, so looks like a typo

playbooks/provisioning/openstack/sample-inventory/clouds.yaml
```diff
@@ -0,0 +1,5 @@
+ansible:
```
bogdando: Please put as well
```
#clouds:                                                                                                                                                                           
#	<cloud-name:
#		auth:
#			auth_url: <auth_url>
#			project_name: <project_name>
#			username: <username>
#			password: <password>
#		region: RegionOne
```

playbooks/provisioning/openstack/README.md
```diff
@@ -0,0 +1,113 @@
...
+* Optionally attach Red Hat subscriptions
+* Set up a bind-based DNS server
+* When deploying more than one master, set up a HAproxy server
+
+
+## Set up
+
+### Copy the sample inventory
+
+    cp openshift-ansible-contrib/playbooks/provisioning/openstack/sample-inventory inventory
```
bogdando: cp -r

playbooks/provisioning/openstack/README.md
```diff
@@ -0,0 +1,113 @@
...
+    cp openshift-ansible-contrib/playbooks/provisioning/openstack/sample-inventory/clouds.yaml clouds.yaml
+
+### Update `inventory/group_vars/all.yml`
+
+Pay special attention to the values in the first paragraph -- these
+will depend on your OpenStack environment.
+
+The `env_id` and `openstack_dns_domain` will form the DNS domain all
+your servers will be under. With the default values, this will be
+`openshift.example.com`.
```
bogdando: Let's insist for the guide to update it to something like `openshift.cluster.lc`, which has nothing to publicly DNS-resolvable domains.

playbooks/provisioning/openstack/sample-inventory/group_vars/all.yml
```diff
@@ -0,0 +1,39 @@
...
+
+docker_volume_size: "15"
+
+# TODO(shadower): this is identical to `openstack_dns_domain`.
+# We should make it so it's not duplicated here.
+dns_domain: "example.com"
+
+# TODO(shadower): this is identical to `openstack_nameservers`.
+# We should make it so it's not duplicated here.
+public_dns_forwarder: "192.168.1.1"
```
bogdando: please use "{{openstack_nameservers}}" then, or {{openstack_nameservers[0]}} perhaps

playbooks/provisioning/openstack/README.md
```diff
@@ -0,0 +1,113 @@
...
+You can see your flavors with `openstack flavor list`.
+
+`openstack_external_network_name` is the name of the Neutron network
+providing external connectivity. It is often called `public`,
+`external` or `ext-net`. You can see your networks with `openstack
+network list`.
+
+The `openstack_num_masters`, `openstack_num_infra` and
+`openstack_num_nodes` values specify the number of Master, Infra and
+App nodes to create.
```
bogdando: anything to note wrt the `openstack_subnet_prefix` defaults please?

playbooks/provisioning/openstack/sample-inventory/group_vars/all.yml
```diff
@@ -0,0 +1,39 @@
...
+
+openstack_num_masters: 1
+openstack_num_infra: 1
+openstack_num_nodes: 2
+
+docker_volume_size: "15"
+
+# TODO(shadower): this is identical to `openstack_dns_domain`.
+# We should make it so it's not duplicated here.
+dns_domain: "example.com"
```
bogdando: please use "{{openstack_dns_domain}}" then

playbooks/provisioning/openstack/sample-inventory/group_vars/OSEv3.yml
```diff
@@ -0,0 +1,10 @@
+---
+openshift_deployment_type: openshift-enterprise
+openshift_release: v3.5
+openshift_master_default_subdomain: "apps.openshift.example.com"
```
bogdando: please use "apps.{{openstack_dns_domain}}"

playbooks/provisioning/openstack/README.md
```diff
@@ -0,0 +1,113 @@
...
+For example, if your final domain is `my.cloud.com`, you can run this
+command to fix update the `hosts` file:
+
+    sed -i 's/openshift.example.com/my.cloud.com/' inventory/hosts
+
+### Configure the OpenShift parameters
+
+Finally, you need to update the DNS entry in
+`inventory/group_vars/OSEv3.yml` (look at
+`openshift_master_default_subdomain`).
```
bogdando: this step may be removed, if the default value adjusted (see the comment below)

playbooks/provisioning/openstack/pre_tasks.yml
bogdando: Could you please move these to the `environment:` clause of the tasks below that want these env vars?

playbooks/provisioning/openstack/README.md
```diff
@@ -0,0 +1,113 @@
...
+
+
+## Deployment
+
+### Run the playbook
+
+Assuming your OpenStack (Keystone) credentials are in the `keystonerc`
+file, this is how you stat the provisioning process:
+
+    . keystonerc
```
bogdando: could we make this do not rely on the stackrc file, but only the clouds.yaml?

bogdando:
@tomassedovic please introduce a "flat-secgrp" that merges all rules into a one set. I have exceeded my quota with testing these RDO cloud. This creates 34 rules, and the quota is 50 :(

roles/openshift-prep/tasks/prerequisites.yml
```diff
@@ -0,0 +1,36 @@
...
+    name: "{{ item }}"
+    state: latest
+  with_items:
+    - wget
+    - git
+    - net-tools
+    - bind-utils
+    - bridge-utils
+    - bash-completion
+    - atomic-openshift-utils
```
bogdando: this doesn't work for centos7 w/o rhsm repos registered. Could you please ensure centos7 compatibility?

playbooks/provisioning/openstack/sample-inventory/openstack.py
```diff
@@ -0,0 +1,252 @@
+#!/usr/bin/env python
```
bogdando: Could we just rely on the shade inventory directly, w/o this wrapper script?

roles/openshift-prep/tasks/prerequisites.yml
bogdando: to clarify, this package limits the use cases for the provider to the OpenShift enterprise only. Please let's assure it's working with origin on centos7 as well. Perhaps we could you this docker images for the upstream origin deployment cases https://hub.docker.com/r/karafior/atomic-openshift-utils/  ?

playbooks/provisioning/openstack/sample-inventory/openstack.py
tomassedovic: Ohh, good point! I didn't know it had one. For the record, this is the openstack inventory script that comes from Ansible, not something we wrote or going to maintain ourselves. But it's worth investigating shade's inventory, too.

roles/openshift-prep/tasks/prerequisites.yml
```diff
@@ -0,0 +1,35 @@
...
+  yum:
+    name: "{{ item }}"
+    state: latest
+  with_items:
+    - wget
+    - git
+    - net-tools
+    - bind-utils
+    - bridge-utils
+    - bash-completion
```
cooktheryan: @tomassedovic no rush now but you may want to look at copying the prereq role the other providers use to be able to handle atomic

playbooks/provisioning/openstack/pre-install.yml
```diff
@@ -0,0 +1,72 @@
...
+    include_vars: /tmp/records.yml
+  roles:
+    - role: dns-server
+    - role: dns
+
+# OpenShift Pre-Requisites
+- hosts: OSEv3
+  become: true
+  tasks:
+  - name: "Edit /etc/resolv.conf on masters/nodes"
```
bogdando: this should be persisted by network manager I believe, see https://github.com/openshift/openshift-ansible-contrib/issues/455

bogdando: for example, here is an example implementation we could try to benefit from https://github.com/kubernetes-incubator/kargo/tree/master/roles/kubernetes/preinstall/tasks

oybed: This overall needs a better solution as this was originally introduced as a "quick fix" to allow for the internally provisioned DNS server. I'd suggest getting this *huge* PR merged first, however, then go back and clean-up this area.

playbooks/provisioning/openstack/sample-inventory/group_vars/OSEv3.yml
oybed: This is changing per PR https://github.com/redhat-cop/casl-ansible/pull/48

playbooks/provisioning/openstack/sample-inventory/group_vars/all.yml
oybed: this is changing per PR https://github.com/redhat-cop/casl-ansible/pull/48

playbooks/provisioning/openstack/sample-inventory/group_vars/all.yml
oybed: This is changing per PR https://github.com/redhat-cop/casl-ansible/pull/48

playbooks/provisioning/openstack/templates/named_views.template
```diff
@@ -0,0 +1,14 @@
+---
```
oybed: file removed per PR https://github.com/redhat-cop/casl-ansible/pull/48

playbooks/provisioning/openstack/templates/records.template.yml
```diff
@@ -0,0 +1,23 @@
+---
```
oybed: file removed per PR https://github.com/redhat-cop/casl-ansible/pull/48

roles/openshift-prep/tasks/prerequisites.yml
oybed: I really like this and looking forward to be handling atomic. However, could that be a follow-up PR? (and issue?)

roles/openstack-stack/README.md
```diff
@@ -0,0 +1,9 @@
+# Role openstack-stack
+
+Role for spinning up instances using OpenStack Heat.
+
+## To Test
+
+```
+ansible-playbook casl-ansible/roles/openstack-stack/test/stack-create-test.yml
```
oybed: needs to be updated to not reference the `casl-ansible` repo

roles/openstack-stack/tasks/main.yml
oybed: Agreed - I think we should aim to merge as much as-is to have a working implementation, then do smaller follow-up PRs to do clean-up / renaming.

roles/openshift-prep/tasks/prerequisites.yml
```diff
@@ -0,0 +1,35 @@
...
+    - vim-enhanced
+
+- name: "Update all packages (this can take a very long time)"
```
oybed: Note that this may require a host reboot - i.e.: if the kernel changed, etc. We  have a small role that does the update and reboot the host if needed (and wait for it to come back). I'd suggest consider something like that for future in a follow-up PR.
https://github.com/redhat-cop/infra-ansible/tree/master/roles/update-host

playbooks/provisioning/openstack/pre-install.yml
bogdando: yeah, that works. Follow up patches could fix that as well

bogdando:
LGTM, would be nice to have it merged, then having the comments addressed by follow-up patches

playbooks/provisioning/openstack/sample-inventory/group_vars/all.yml
```diff
@@ -0,0 +1,42 @@
+env_id: "openshift"
+openstack_dns_domain: "example.com"
+openstack_nameservers: ["192.168.1.1"]
```
bogdando: the fix for networkmanager config that worked for me was using:
```
- "{{ hostvars[groups.dns|first].openstack.private_v4 }}"
- 8.8.8.8
```
This allows the nodes to resolve both external and internal names, with a priority for the intranet-first

oybed: That's not necessary, and could potentially even cause problems with a circular resolution. With PR https://github.com/redhat-cop/casl-ansible/pull/48, the DNS name server config is being consolidated and named `public_dns_nameservers` as a more appropriate name - i.e.: what's the "fall-back" DNS server(s) in case the internal ones don't resolve the requested name.

bogdando: JFYI, I have a WIP patch what addresses the hosts DNS setup as a persisted config, via the heat stack diven dns node IPs merged with custom vars. It helped to address my dev env blocker.
Will polish it and submit soon.

tomassedovic:
Alright, the casl-ansible team merged the nsupdate DNS work. I'll add it here, test it and then we can start thinking about merging this.

tomassedovic:
Okay, this includes the DNS changes (tested with ocp 3.5 & rhel). I'll fix yaml lint and then we can talk about merging this.

tomassedovic:
@cooktheryan assuming the tests pass, I'd like to get this merged.

It's got a bunch of issues with it (casl references we want to update, the docker storage stuff you suggested, etc.), but it should also have everything necessary to do a deployment and maintaining the PR is a bit of a hassle.

So if we can, I'd like to merge this soon and then keep improving it.

cooktheryan:
@tomassedovic i will monitor the results

PR #403, opened 2017-05-25
gitless validation testing - safer for atomic

cooktheryan:
Resolves issue #402

### DIFF
--- a/roles/validate-app/tasks/main.yaml
+++ b/roles/validate-app/tasks/main.yaml
@@ -8,24 +8,21 @@
       command: "{{ openshift.common.client_binary }} new-project validate"
 
     - name: Create Hello world app
-      shell: "{{ openshift.common.client_binary }} new-app https://github.com/openshift/cakephp-ex.git"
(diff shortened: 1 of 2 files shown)


roles/validate-app/tasks/main.yaml
```diff
@@ -8,32 +8,29 @@
       command: "{{ openshift.common.client_binary }} new-project validate"
 
     - name: Create Hello world app
-      shell: "{{ openshift.common.client_binary }} new-app https://github.com/openshift/cakephp-ex.git"
+      shell: "{{ openshift.common.client_binary }} new-app cakephp-mysql-example"
```
e-minguez: I'd say to set cakephp-mysql-example to a variable to avoid rewritting it a few times (and change it if needed)

roles/validate-app/tasks/main.yaml
```diff
@@ -8,32 +8,29 @@
       command: "{{ openshift.common.client_binary }} new-project validate"
```
e-minguez: I'd say to set validate to a variable to avoid rewritting it a few times (and change it if needed)

PR #404, opened 2017-05-25
ensure that directories exist before beginning installation

cooktheryan:
Pre-flight check that will ensure that /usr/share/ansible/openshift-ansible is populated


cooktheryan:
@dav1x and @pschiffe flag a github issue if you want me to integrate this into your providers

pschiffe:
I've actually stopped using the `/usr/share/ansible/openshift-ansible` dir because the rpm is available only in rhel and for the git clone, you need the root access. So I'm currently using the `openshift-ansible` folder next to the `openshift-ansible-contrib` folder on the local machine, and I'm thinking about the ansible role, which would clone the `openshift-ansible` repo there and kept it updated to the latest version of current / chosen ocp release. Only challenge I can currently see is how to programmatically determine the latest `openshift-ansible` release within the specific ocp release..

PR #407, opened 2017-05-28
gce-cli: reference config.yaml instead of config.sh

abn:
Straightforward change changing `config.sh` references to the new `config.yaml` in the README file.


pschiffe:
Thank you @abn !

abn:
@pschiffe  :+1:

PR #411, opened 2017-05-30
azure.conf should be ini format, no json

e-minguez:
https://docs.openshift.com/container-platform/3.5/install_config/configuring_azure.html#azure-configuration-file
I've tried to create a PVC with the previous format and it failed with "Failed to create provisioner: Failed to get Azure Cloud Provider. GetCloudProvider returned <nil> instead" and after doing this modifications and restarting master service, it works.


e-minguez:
Tested again with the old format and it works, so it should be related with the service restart.

PR #412, opened 2017-05-30
ocp_tmp_instance variable

eye0fra:
As you know, ansible_ssh_host contains the "ocp-tmp-instance" as value, which is not a valid host. Because you cannot reference the "**hostsvars**" inside  **hosts** keyword, the same fix to use the **gce_public_ip** is not applicable.

So I had to externalise the var **ocp_tmp_instance** inside **gold-image.yaml**


pschiffe:
Thanks for the PR, I just used slightly different, more robust, approach here - https://github.com/openshift/openshift-ansible-contrib/pull/414/commits/32743e26a4faf22885e3a13cc1b2247f126a39e4

eye0fra:
That's great! Thanks

PR #414, opened 2017-05-30
Configure ssh for the temp instance when creating gold image

pschiffe:


pschiffe:
@cooktheryan how do I restart the checks? They failed but I don't see any reason..

cooktheryan:
@pschiffe try to trigger a change in the PR...add a space or an empty line somewhere. A follow up commit to this PR can remove it if required

cooktheryan:
@pschiffe I think you have the magic that breaks travis

PR #415, opened 2017-05-30
more vars for static inventory

### LINKED ISSUE #406: No hosts file under /etc/ansible
I've installed Openshift Origin by executing `ose-on-aws.py`. The cluster is now operative but I'd like to install metrics and logging. According to the documentation I should follow the [Advanced Installation](https://docs.openshift.org/latest/install_config/install/advanced_install.html#configuring-ansible). I'm not able to find `/etc/ansible/hosts` file to configure Ansible. Any idea about that? My idea was to use `- include: /usr/share/ansible/openshift-ansible/playbooks/byo/config.yml` but what should I use as inventory? The EC2 hosts inventory or a plain file configured as wanted?

### LINKED ISSUE #408: Containerized not as parameter
I don't know If I'm wrong but `ose-on-aws.py` has a parameter `containerized`. This parameter is not parametrized in the `openshift-setup.yaml`. Is it the expected behaviour?

cooktheryan:
Resolves #406

### DIFF
--- a/reference-architecture/aws-ansible/playbooks/roles/inventory-file-creation/files/inventory
+++ b/reference-architecture/aws-ansible/playbooks/roles/inventory-file-creation/files/inventory
@@ -25,7 +25,7 @@ osm_default_subdomain={{ wildcard_zone }}
 openshift_master_default_subdomain={{ wildcard_zone }}
 osm_default_node_selector="role=app"
 deployment_type={{ deployment_type | default('openshift-enterprise') }}
-os_sdn_network_plugin_name={{ openshift_sdn | default('openshift-ovs-subnet') }}
+os_sdn_network_plugin_name={{ openshift_sdn | default('redhat/openshift-ovs-subnet') }}
 #openshift_master_identity_providers=[{'name': 'github', 'challenge': 'false', 'login': 'true', 'kind': 'GitHubIdentityProvider', 'mapping_method': 'true', 'clientID': '{{ github_client_id }}', 'clientSecret': '{{ github_client_secret }}', 'organizations': {'['{{ github_organization }}']'}}]
 osm_use_cockpit=true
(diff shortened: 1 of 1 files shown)


reference-architecture/aws-ansible/playbooks/roles/inventory-file-creation/files/inventory
```diff
@@ -24,11 +24,11 @@ openshift_master_cluster_public_hostname=openshift-master.{{ public_hosted_zone
 osm_default_subdomain={{ wildcard_zone }}
 openshift_master_default_subdomain={{ wildcard_zone }}
 osm_default_node_selector="role=app"
-deployment_type=openshift-enterprise
-os_sdn_network_plugin_name=redhat/{{ openshift_sdn | default('openshift-ovs-subnet') }}
+deployment_type={{ deployment_type | default('openshift-enterprise') }}
+os_sdn_network_plugin_name={{ openshift_sdn | default('redhat/openshift-ovs-subnet') }}
```
e-minguez: I'd say to use the multitenant plugin by default for security reasons

eye0fra: I agree

glennswest:
Think it's safer to do it that way 
It was a intentional choice

glennswest:
The other reason you uses this, is it should make mdns discovery work properly, otherwise mdns will go all over the place, and not what you want.

cooktheryan:
@e-minguez you good with approving this one and a subsequent change of me switching the default SDN there. It would require some doc updates that I don't want to have to force in at the moment.

e-minguez:
aos-ci-test

e-minguez:
aos-ci-test

PR #419, opened 2017-05-31
Missing node type

dav1x:


playbooks/add-node-prerequisite.yaml
```diff
@@ -4,6 +4,7 @@
   become: yes
   serial: 1
   roles:
+    - instance-groups
```
cooktheryan: Ill need to test this really quick

PR #422, opened 2017-06-01
Propose Get pool id maches fix.

eye0fra:
The --matches filters only the subscription or product information.

Sample tests:

subscription-manager list --available --pool-only | grep 1234
1234

subscription-manager list --available --pool-only --matches 1234

subscription-manager list --available --pool-only --matches '*OpenShift*' | grep 1234
1234

### DIFF
--- a/roles/rhsm-subscription/tasks/main.yaml
+++ b/roles/rhsm-subscription/tasks/main.yaml
@@ -45,19 +45,19 @@
       when: "'Current' not in subscribed.stdout and rhsm_user is defined and rhsm_user"
 
     - name: Check if subscription is attached
-      command: subscription-manager list --consumed --pool-only --matches="{{ rhsm_pool }}"
+      shell: "subscription-manager list --consumed --pool-only --matches '*OpenShift*' | grep {{ rhsm_pool }}"
       register: subscription_attached
       changed_when: no
 
     - block:
         - name: Get pool id
-          shell: subscription-manager list --available --pool-only --matches="{{ rhsm_pool }}" | head -n 1
+          shell: "subscription-manager list --available --pool-only --matches '*OpenShift*' | grep {{ rhsm_pool }} | head -n 1"
           register: pool_id
           changed_when: no
 
         - name: Fail if no pool ID is returned
           fail:
-            msg: No subscription matching "{{ rhsm_pool }}" found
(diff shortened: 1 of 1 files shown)


pschiffe:
This is not a good fix because not every pool which contains OpenShift has OpenShift in the name. Also with this approach, it would be possible only to use pool IDs, but I guess we want to have the possibility to use both, pool names and IDs. But currently only pool names work.

eye0fra:
Are you sure ? 

It should contain at least the name of the product under "Provides".

For instance:

subscription-manager list --available --matches '*OpenShift*'

Subscription Name:   Employee SKU
Provides:            Oracle Java (for RHEL Server) - AUS
                     Oracle Java (for RHEL Client)
                     Red Hat Enterprise Linux 7 High Availability High Touch Beta
                     Red Hat Enterprise Linux High Availability (for IBM z Systems)

                     Red Hat OpenShift Enterprise JBoss FUSE add-on
                     Red Hat Enterprise Linux Scalable File System (for RHEL Server) - Extended Update Support
 
                     Red Hat Enterprise Linux Server - Extended Life Cycle Support (for IBM z Systems)
                     Red Hat JBoss A-MQ Interconnect
                     Red Hat Certificate System with Advanced Access
                     Red Hat Enterprise Linux High Performance Networking (for RHEL Server)


                      .…
                     Red Hat **OpenShift** Enterprise Infrastructure
                      …. 

                     Red Hat **OpenShift** Container Platform

Anyway the matches is used only to speed up the search in subscription-manager, because you cannot use the poolId as matches, it will always empty.

pschiffe:
Ah, OK, so it does match also in the provides field and this is only a speed up in the search.. I got it wrong before.

pschiffe:
aos-ci-test

pschiffe:
In that case, I think it's OK, @cooktheryan what do you think? If you don't see any obvious problem, I think we can merge this.

mazzystr:
Hello,
I did test the diff and it does work so thank you for that.

Unfortunately we cannot merge.  The team had a discussion regarding this role.  There are other pitfalls that need to be addressed.  The openshift-ansible team has developed an rhsm role.  The team is in favor of dropping this role and integrating into that role.  That work be happening in the very near future.

Thanks!

PR #428, opened 2017-06-02
allow for metrics size to be chosen

### LINKED ISSUE #426: Add the ability to specify the size of metrics persistent storage in aws-ansible scripts
As a user I want the ability to specify the size of the persistent storage behind cluster metrics. Right now I believe it defaults to 20Gi statically.

cooktheryan:
Resolves #426

### DIFF
--- a/reference-architecture/gce-cli/ansible/playbooks/openshift-minor-upgrade.yaml
+++ b/reference-architecture/gce-cli/ansible/playbooks/openshift-minor-upgrade.yaml
@@ -27,4 +27,3 @@
     containerized: false
     openshift_hosted_metrics_storage_kind: dynamic
     openshift_master_metrics_public_url: 'https://metrics.{{ osm_default_subdomain }}/hawkular/metrics'
-    openshift_hosted_metrics_storage_volume_size: 20Gi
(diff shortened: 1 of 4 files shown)


reference-architecture/gce-cli/ansible/playbooks/openshift-minor-upgrade.yaml
```diff
@@ -27,4 +27,3 @@
     containerized: false
     openshift_hosted_metrics_storage_kind: dynamic
     openshift_master_metrics_public_url: 'https://metrics.{{ osm_default_subdomain }}/hawkular/metrics'
-    openshift_hosted_metrics_storage_volume_size: 20Gi
```
pschiffe: You also need to do this change in https://github.com/openshift/openshift-ansible-contrib/blob/master/reference-architecture/gce-cli/ansible/playbooks/openshift-setup.yaml#L25

cooktheryan: @pschiffe good catch. The run time values defined should now be correct and allow for metrics to be deployed at a specific dize

PR #441, opened 2017-06-08
WIP: Add retry logic for restart of atomic-openshift services

glennswest:
Randomly in the final stage of the install when the cloud provider was being setup,  during the restart of atomic-openshift services, a error could occur, and only some of the masters would get the azure.conf setup properly. The ansible script would terminate with a partial config of azure.conf so 1, 2 of the masters are correct and 1 or 2 are not configured. Depending on which master is processing the request for a PV, a nil would be returned, and the PV setup would hang.

As the ansible systemd module does not have retry support, switched to the shell module.

Also add a marker to allow for automated test after a completed install.


cooktheryan:
We should move up the population of azure.conf to before the installation. We can always fire these restarts after but those should be unnecessary in the future. Did you have a BZ open in regards to the azure cloud provider? There has been one release since Summit and another around the corner. If it hasn't been resolved yet we need to sync with the openshift-ansible team immediately as people in the field may not know this work around.

glennswest:
I'm trying it 
Actually the whole azure conf is badly documented

reference-architecture/azure-ansible/bastion.sh
```diff
@@ -294,48 +207,18 @@ cat > /home/${AUSERNAME}/setup-azure-node.yml <<EOF
...
-
-
 cat <<EOF > /etc/ansible/hosts
 [OSEv3:children]
 masters
 nodes
 etcd
 
 [OSEv3:vars]
+osm_controller_args={'cloud-provider': ['azure'], 'cloud-config': ['/etc/azure/azure.conf']}
```
e-minguez: Just in case, according to this[1], containerized installations mount /etc/origin and /var/lib/origin only, so they recommend to store cloud files in those folders.

[1] https://docs.openshift.com/container-platform/3.5/install_config/configuring_openstack.html#openstack-configuring-nodes-manually

glennswest:
Switch to rcook method

PR #449, opened 2017-06-09
Change to let installer setup azure.conf

glennswest:
Use the ansible installer to write the azure.conf settings.
Verifyied as working.
Remove the restarts fo services.
Make sure the touch on completion of install is supported for automated test.


reference-architecture/azure-ansible/bastion.sh
```diff
@@ -178,9 +178,9 @@ EOF
 
 # Create Azure Cloud Provider configuration Playbook
 
-cat > /home/${AUSERNAME}/setup-azure-master.yml <<EOF
+cat > /home/${AUSERNAME}/azure-config.yml <<EOF
 #!/usr/bin/ansible-playbook
-- hosts: masters
+- hosts: all
   gather_facts: no
   serial: 1
```
e-minguez: I think "serial" can be removed, because it can be done on all the hosts at the same time. Thoughts?

reference-architecture/azure-ansible/bastion.sh
```diff
@@ -190,94 +190,7 @@ cat > /home/${AUSERNAME}/setup-azure-master.yml <<EOF
     azure_conf_dir: /etc/azure
     azure_conf: "{{ azure_conf_dir }}/azure.conf"
     master_conf: /etc/origin/master/master-config.yaml
```
e-minguez: master_conf variable is not longer used I think this can be safely removed too. Thoughts?

PR #451, opened 2017-06-09
Download openshift installer from github and verify req. packages

pschiffe:


reference-architecture/gce-cli/ansible/playbooks/roles/openshift-ansible-installer/tasks/main.yaml
```diff
@@ -0,0 +1,6 @@
+---
+- name: download openshift ansible installer from github
+  git:
+    repo: https://github.com/openshift/openshift-ansible.git
+    dest: '{{ playbook_dir }}/../../../../../openshift-ansible'
```
cooktheryan: I'm not sure that we should be doing this outside of origin installs. If a version/branch is updated before RPMs are cut we may get into an issue with support. Although our installations are not supported by RH support but the resulting OpenShift environment is required to be supported by RH. I think by using git we may be setting ourselves up for some future issues

pschiffe: I've considered this risk, but looking at the history of the release branches it's clear that the development is done in the `master` branch and the changes in the `release` branches are usually quickly followed by a release tag (and the rpm). So the risk is there, but quite low I would say. Main reason why I'm doing this is that the rpm is not available in Fedora and CentOS and o-a-c is also not in rpm, so.. 

But I can add support for installer in rpm later, currently, only installer next to the o-a-c repo folder works.

e-minguez: Maybe we can check the installation type. If it is enterprise and RHEL involved, then install the RPMs, otherwise (origin, Fedora, CentOS), go with the git repo.
My 2 cents.

cooktheryan: I'm ok with this if we add support for it in the future. It does currently stink that you have to use a license for OCP on your deployer box.

e-minguez: OCP subscription provides 10x subscriptions[1] for infra stuff (including infra nodes, masters, bastions) on every node subscription, so I agree it stinks but it shouldn't be a problem regarding subscriptions.
[1] https://access.redhat.com/solutions/2851701

pschiffe: I was looking into this, but I'm blocked by static playbook includes in ansible [1]. Tasks can be included dynamically, but playbooks not. So if I determine dynamically which playbook to include in openshift-setup.yaml [2] (based on check if installer is available as rpm), inclusion fails :-( You can see my non-working commit here [3]. Do you have any ideas how this could be solved?

[1] http://docs.ansible.com/ansible/playbooks_roles.html#dynamic-versus-static-includes
[2] https://github.com/openshift/openshift-ansible-contrib/blob/master/reference-architecture/gce-cli/ansible/playbooks/openshift-setup.yaml#L2
[3] https://github.com/pschiffe/openshift-ansible-contrib/commit/0229ae77c6e9aca2eb95a073266071147d34ae4a

e-minguez: It is quite ugly, but you can use the same file path and mess with symlinks... so if it is RHEL, create a symlink to /usr/share/ansible/openshift-ansible/playbooks/... otherwise, leave it as it should.

glennswest:
Using the git repo is not synced to rpm relesaes, big problems

reference-architecture/gce-cli/ansible/playbooks/roles/openshift-ansible-installer/tasks/main.yaml
pschiffe: Good point @e-minguez , that might work. I'll play with it.

PR #460, opened 2017-06-15
Use ansible installer hosted registry role to deploy registry

pschiffe:


reference-architecture/gce-cli/ansible/playbooks/openshift-setup.yaml
```diff
@@ -10,14 +10,19 @@
     openshift_master_api_port: '{{ console_port }}'
     openshift_master_console_port: '{{ console_port }}'
     osm_cluster_network_cidr: 172.16.0.0/16
-    openshift_registry_selector: 'role=infra'
-    openshift_router_selector: 'role=infra'
-    openshift_hosted_router_replicas: 3
```
cooktheryan: Are you removing the replicas scaling completely?

pschiffe: No, the default is the number of selected hosts (in this case, hosts with `role=infra`). So there's no change.

cooktheryan: wow didn't know that. good work

PR #462, opened 2017-06-15
Finish crs

dav1x:
1. Added heketi install and config
2. Added plays to install crs vms
3. Added heketi auto secret and topology generation


cooktheryan:
If you lint this I'll give it a check

Issue #468, opened 2017-06-19
[Openstack provider] Installation of provisioning prerequisites must be automated for cluster nodes


bogdando:
Provision prerequisites (for cluster nodes) must be covered by some automation. We can not expect users to ssh to nodes and install prerequisites once they provisioned nodes on top of OpenStack cloud with a Heat stack. User_data could help to address that.

For localhost played roles/tasks, it is OK to leave them only documented perhaps.

bogdando:
@oybed @tomassedovic

tomassedovic:
It's already there, isn't it?

https://github.com/openshift/openshift-ansible-contrib/blob/master/playbooks/provisioning/openstack/README.md#dependencies

bogdando:
@tomassedovic yes, in docs. I'll submit a patch to install the package as well

bogdando:
Ok, there is localhost requirements, which is ok to not automate I suppose. This ticket is for addressing requirements (if any) for the cluster nodes. Sorry for confusion!

tomassedovic:
Closing this. All packages the cluster nodes require are installed by the playbooks, no manual work required.

PR #469, opened 2017-06-19
[WIP] Install DNS config prerequisites

bogdando:
Ansible nsupdate module requires python-dns
installed for provisioned nodes.

Closes: https://github.com/openshift/openshift-ansible-contrib/issues/468

### DIFF
--- a/roles/dns/tasks/main.yml
+++ b/roles/dns/tasks/main.yml
@@ -1,4 +1,9 @@
 ---
+- name: Install prerequisites
+  package:
+    name: python-dns
+  become: true
+
 - name: "Remove any deleted DNS A records"
   nsupdate:
     key_name: "{{ item.0.key_name }}"


roles/dns/tasks/main.yml
```diff
@@ -1,4 +1,8 @@
 ---
+- name: Install prerequisites
+  package:
+    name: python-dns
```
tomassedovic: This package is called "python2-dns" on Fedora, so running it there would fail the playbook.

tomassedovic:
I'm not sure we want to do this in the playbooks. First, that will fail on any system that has the package called something else (like Fedora) and second, I'd like the provision playbook to be runnable without root on the ansible host and without modifying it..

I'd be open to adding say a "prerequisites" playbook that would install the host prereqs other than ansible (so shade, python-dns and anything else we might need later), but I'm not keen on having it here deep in the role.

roles/dns/tasks/main.yml
cooktheryan: Might be a good time to gather_facts about localhost and then start adding in blocks

bogdando:
@tomassedovic ack. Let's do that as a separate action item. Provision prerequisites must be covered by some automation. I mean we would not expect users to ssh to nodes and install prerequisites once they provisioned nodes on top of OpenStack cloud with heat.

PR #470, opened 2017-06-19
Add profiling and skippy stdout

bogdando:
Tune an example ansible.cfg to include
tasks profiling info and improve displaying
of skipped tasks.


tomassedovic:
@bogdando Thanks! I'm running it now and see that it prints out each tasks' duration (which is awesome!).

Will it show an  overview/collect all timings at the end, too? I wonder whether I should pipe this somewhere for later inspection...

tomassedovic:
@cooktheryan can we get this merged?

bogdando:
@tomassedovic 
> Will it show an overview/collect all timings at the end, too?

yeah, it does that

Issue #455, opened 2017-06-13
[Openstack provider] persist nameservers by networkmanager via dnsmasq deployed by openshift-ansible


bogdando:
Testing OpenShift origin on OpenStack, with Centos7 servers provisioned by Heat.
The networkmanager config needs to persist nameservers to the hosts' /etc/resolv.conf or better off to the dnsmasq deployed. The current [code](https://github.com/openshift/openshift-ansible-contrib/blob/d81278c8e4b28f0ff756219d174a3ceef57056e4/playbooks/provisioning/openstack/post-provision-openstack.yml#L46) does not persist that `private_dns_server` and the hacked /etc/resolv.conf entry will be removed after the ``systemctl restart network`` command or a reboot.

bogdando:
https://github.com/openshift/openshift-ansible-contrib/pull/397/files#r121722908

bogdando:
While the fix works for hosts, containers seem still receiving nameservers in a wrong order:

See the snippet that demonstrates the name resolve issues, both for the service FQDN from within deployed openshift cluster, and from the container of an example cakePHP app to external resources (github) https://pastebin.com/V1c91juW

The failed container's config:
```
[openshift@app-node-0 ~]$ sudo docker ps -a
CONTAINER ID        IMAGE                                                                                             COMMAND                  CREATED             STATUS                      PORTS               NAMES
564693ddddb3        openshift/origin-sti-builder:v1.5.1                                                               "/usr/bin/openshift-s"   19 minutes ago      Exited (1) 18 minutes ago                       k8s_sti-build.8a7c96a8_cakephp-mysql-example-1-build_foo_370d1312-529e-11e7-a284-fa163e18994e_78b81144

[openshift@app-node-0 ~]$ sudo docker cp 564693ddddb3:/etc/resolv.conf /tmp/resolv.conf
[openshift@app-node-0 ~]$ cat /tmp/resolv.conf
search foo.svc.cluster.local svc.cluster.local cluster.local openstacklocal openshift.cluster.lc
nameserver 172.30.0.1
nameserver 192.168.99.8
nameserver 192.168.1.1
nameserver 8.8.8.8
```

bogdando:
The command dig cakephp-mysql-example-foo.apps.openshift.cluster.lc @<dns-0-floating-IP> works from my laptop. So it seems I have only bad DNS config for container that prevents it to resolve external names.

bogdando:
[...]

PR #471, opened 2017-06-19
[WIP] Persist DNS configuration for nodes for openstack provider

bogdando:
Persist DNS configuration for nodes for openstack provider

* Add the custom dnsmasq config for openstack provider needs.
* Point the dnsmasq config to the defined private_dns_server.
* Speed up DNS queries against bogus/malformed domain names with
  NXDOMAIN reported early.
* Add the dnsmasq role to ship the custom config file for servers.
* Drop post-provision /etc/reslov.conf nameserver hacks to be
  persisted by openshift-ansible dnsmasq role via the custom config.
* Fix dns floating IPs output and add the priv IPs output as well.
* Update docs, clarify localhost vs servers requirements, add
  required Network Manager setup step.

TODO add missing docs for external_nsupdate_keys, public_dns_server.

Closes https://github.com/openshift/openshift-ansible-contrib/issues/455

### DIFF
--- a/playbooks/provisioning/openstack/README.md
+++ b/playbooks/provisioning/openstack/README.md
@@ -5,12 +5,18 @@ OpenStack resources (servers, networking, volumes, security groups,
 etc.). The result is an environment ready for openshift-ansible.
 
 
-## Dependencies
+## Dependencies for localhost (ansible admin node)
 
 * [Ansible 2.3](https://pypi.python.org/pypi/ansible)
 * [shade](https://pypi.python.org/pypi/shade)
-* python-dns
+* python-dns / [dnspython](https://pypi.python.org/pypi/dnspython)
+* Become (sudo) is not required.
 
+## Dependencies for OpenStack hosted cluster nodes (servers)
+
+There are no additional dependencies for the cluster nodes. Required
+configuration steps are done by Heat given a specific user data config
+that normally should not be changed.
 
 ## What does it do
 
@@ -41,12 +47,22 @@ etc.). The result is an environment ready for openshift-ansible.
 Pay special attention to the values in the first paragraph -- these
 will depend on your OpenStack environment.
 
-The `env_id` and `openstack_dns_domain` will form the DNS domain all
+The `env_id` and `public_dns_domain` will form the cluster's DNS domain all
 your servers will be under. With the default values, this will be
-`openshift.example.com`.
+`openshift.example.com`. For workloads, the default subdomain is 'apps'.
+That sudomain can be set as well by the `openshift_app_domain` variable in
+the inventory.
 
-`openstack_nameservers` is a list of DNS servers accessible from all
-the created Nova servers. These will be serve as your DNS forwarders.
+The `public_dns_nameservers` is a list of DNS servers accessible from all
+the created Nova servers. These will be serving as your DNS forwarders for
+external FQDNs that do not belong to the cluster's DNS domain and its subdomains.
+
+The `openshift_use_dnsmasq` controls either dnsmasq is deployed or not.
+By default, dnsmasq is deployed and comes as the hosts' /etc/resolv.conf file
+first nameserver entry that points to the local host instance of the dnsmasq
+daemon that in turn proxies DNS requests to the authoritative DNS server.
+When Network Manager is enabled for provisioned cluster nodes, which is
+normally the case, you should not change the defaults and always deploy dnsmasq.
 
 `openstack_ssh_key` is a Nova keypair -- you can see your keypairs with
 `openstack keypair list`.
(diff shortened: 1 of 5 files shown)


playbooks/provisioning/openstack/README.md
```diff
@@ -41,12 +41,23 @@ etc.). The result is an environment ready for openshift-ansible.
...
 will depend on your OpenStack environment.
 
-The `env_id` and `openstack_dns_domain` will form the DNS domain all
+The `env_id` and `public_dns_domain` will form the cluster's DNS domain all
 your servers will be under. With the default values, this will be
-`openshift.example.com`.
-
-`openstack_nameservers` is a list of DNS servers accessible from all
-the created Nova servers. These will be serve as your DNS forwarders.
+`openshift.example.com`. For workloads, the default subdomain is 'apps'.
```
oybed: It's probably worth mentioning that the "apps." sudomain can be set by the `openshift_master_default_subdomain` variable in the inventory.

playbooks/provisioning/openstack/README.md
```diff
@@ -41,12 +41,23 @@ etc.). The result is an environment ready for openshift-ansible.
...
+
+The `public_dns_nameservers` is a list of DNS servers accessible from all
+the created Nova servers. These will be serving as your DNS forwarders for
+external FQDNs that do not belong to the cluster's DNS domain.
+
+The `openshift_use_dnsmasq` controls either dnsmasq is deployed or not.
```
oybed: It's not clear why you may want to **not** use `dnsmasq` ... I believe there are some additional upcoming changes around DNS, so IMHO we should just leave it as-is and don't support not deploying without dnsmasq. Can you please elaborate why we should allow to deploy without?

playbooks/provisioning/openstack/README.md
```diff
@@ -41,12 +41,23 @@ etc.). The result is an environment ready for openshift-ansible.
...
+When dnsmasq is not used, you may only effectively configure a 2nd and a 3rd
+"fallback" nameservers of the maximum of a 3 nameserver records allowed for
```
oybed: Keep in mind that if nameserver 1 responds, none of the other nameservers will be used, so not sure why more than 3 is needed anyway. (we typically run with just 2 in the list).

Typo below `reoslv`...

playbooks/provisioning/openstack/README.md
bogdando: @oybed I was advised to consider that deployment mode, can't recall the context of that conversation happended at the openshift-dev IRC. But given [docs](https://docs.openshift.com/enterprise/3.2/install_config/install/prerequisites.html) I tend to agree we better off to just ignore that option and always deploy with dnsmasq

roles/dnsmasq/templates/openstack-dns.conf.j2
```diff
@@ -0,0 +1,14 @@
...
+cache-size=1000
+dns-forward-max=150
+max-cache-ttl=10
+max-ttl=20
+
+server=/{{ public_dns_domain }}/{{ private_dns_server }}
+# Reply NXDOMAIN to bogus domains requests like com.{{ public_dns_domain }}.{{ public_dns_domain }}
+local=/{%- for d in [ public_dns_domain, full_dns_domain, openshift_master_default_subdomain ] -%}
+{{public_dns_domain}}.{{d}}./{{d}}.{{d}}./com.{{d}}./
+{%- end -%}
```
bogdando: [Kargo](https://github.com/kubernetes-incubator/kargo)-culting this WIP config here. I'd been working on the same topic there, thus let's evaluate if that could benefit the provider as well

roles/dnsmasq/templates/openstack-dns.conf.j2
```diff
@@ -0,0 +1,12 @@
...
+bogus-priv
+no-negcache
+cache-size=1000
+dns-forward-max=150
+max-cache-ttl=10
+max-ttl=20
+
+server=/{{ public_dns_domain }}/{{ private_dns_server }}
+# Reply NXDOMAIN to bogus domains requests like com.{{ public_dns_domain }}.{{ public_dns_domain }}
+local=/{% for d in [ public_dns_domain, full_dns_domain, openshift_master_default_subdomain ] %}{{public_dns_domain}}.{{d}}./{{d}}.{{d}}./com.{{d}}./{% endfor %}
```
bogdando: this may be needed if we will need to change defaults for `ndots: 1`

bogdando:
The current blocker for this persisting DNS config approach is that the ``openshift_master : Start and enable master`` is called before dnsmasq configured with the custom override config (it is done by the ``Configure nodes for dnsmasq`` of the common openshift-cluster playbooks, IIUC. So we need the latter to prepend the openshift cluster nodes' services configuration, it seems.

Such a logic change should be done in the byo playbooks, AFAICT.

@tomassedovic PTAL ^^ thoughts on this?
@oybed is it goes off-road the way you had implemented DNS configs persisting for the infra repo?
Could you point me to that solution, if any, please? Otherwise I'll switch this PR to the infra repo shortly, once I inspect it for the recent changes and rebase.

bogdando:
@tomassedovic recommended to abandon this approach and ensure a consistent DNS nameserver config that comes from the Heat stack's Neutron subnet. Once we have an authoritative DNS server configured, just switch the subnet to it.

PR #483, opened 2017-06-19
Support origin deployments on centos

pschiffe:


pschiffe:
@bdurrow would this work for you?

PR #484, opened 2017-06-20
Use cached facts, do not become for localhost

bogdando:
Prohibit sudoing for localhost played tasks, like DNS setup.
Re-use cached facts to speed up deployment.

### DIFF
--- a/playbooks/provisioning/openstack/provision-openstack.yml
+++ b/playbooks/provisioning/openstack/provision-openstack.yml
@@ -1,6 +1,7 @@
 ---
 - hosts: localhost
   gather_facts: True
+  become: False
   pre_tasks:
   - include: pre_tasks.yml
   roles:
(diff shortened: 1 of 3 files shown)


bogdando:
@tomassedovic PTAL

playbooks/provisioning/openstack/provision-openstack.yml
```diff
@@ -31,12 +32,13 @@
...
   hosts: localhost
   connection: local
+  become: False
   gather_facts: False
   tasks:
   - meta: refresh_inventory
 
 - hosts: cluster_hosts
-  gather_facts: false
+  gather_facts: True
```
bogdando: note: this is a trick to collect all facts for cluster nodes while waiting for provisioned servers to come up. I hope that works.

tomassedovic:
Do you have `become: true` in your ansible.cfg? Ansible should not do sudo by default (or at least it doesn't in my case).

Not that I'm opposed to putting it explicitly there, just wondering.

bogdando:
@tomassedovic AFAIK, putting it top level prevents anything downhill from becoming as well, even if a task wants it, it won't get it. So this kind of enforces a no become requirement.

tomassedovic:
Cheers. Testing it now. LGTM if it passes.

bogdando:
@tomassedovic it fails "dns : Add DNS A records" w/o sudoing on localhost I guess. It seems we've missed the point we could disable become for localhost :(

tomassedovic:
@cooktheryan @pschiffe we can merge this now

PR #488, opened 2017-06-21
Fix flat sec group and infra/dns sec rules

bogdando:
Make flat sec group to only merge node/master/etcd sec rules.
Add basic dns/ssh sec group and assign it to all but dns node groups.
Assign only dns sec group for dns nodes.
Assign only infra (and basic) sec groups for ingra nodes.
Add security notes for openstack provider.

Closes https://github.com/openshift/openshift-ansible-contrib/issues/489

### DIFF
--- a/roles/openstack-stack/templates/heat_stack.yaml.j2
+++ b/roles/openstack-stack/templates/heat_stack.yaml.j2
@@ -142,6 +142,34 @@ resources:
 #            cluster_id: {{ stack_name }}
 #      public_key: {{ ssh_public_key }}
 
+  common-secgrp:
+    type: OS::Neutron::SecurityGroup
+    properties:
+      name:
+        str_replace:
+          template: openshift-ansible-cluster_id-common-secgrp
+          params:
+            cluster_id: {{ stack_name }}
+      description:
+        str_replace:
+          template: Basic ssh/dns security group for cluster_id OpenShift cluster
+          params:
+            cluster_id: {{ stack_name }}
+      rules:
+        - direction: ingress
+          protocol: tcp
+          port_range_min: 22
+          port_range_max: 22
+          remote_ip_prefix: {{ ssh_ingress_cidr }}
+        - direction: ingress
+          protocol: tcp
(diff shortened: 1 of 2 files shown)


bogdando:
@tomassedovic

bogdando:
Please test and merge this ASAP, it fixes unrestricted public DNS access for RDO cloud

roles/openstack-stack/templates/heat_stack.yaml.j2
```diff
@@ -142,18 +142,17 @@ resources:
 #            cluster_id: {{ stack_name }}
 #      public_key: {{ ssh_public_key }}
 
-{% if openstack_flat_secgrp|bool %}
-  flat-secgrp:
+  basic-secgrp:
```
tomassedovic: Can we call this `common-secgrp` instead? I was really confused at first, trying to figure out what a basic security group would mean.

If I understand it correctly, this contains rules that are common to all nodes in the cluster, right?

tomassedovic:
Thanks! I'm testing it now but I have a small naming suggestion inline.

roles/openstack-stack/templates/heat_stack.yaml.j2
bogdando: not really, all but dns. I tried to explain that both in commit msg and the description below

roles/openstack-stack/templates/heat_stack.yaml.j2
```diff
@@ -142,18 +142,17 @@ resources:
...
       name:
         str_replace:
-          template: openshift-ansible-cluster_id-flat-secgrp
+          template: openshift-ansible-cluster_id-basic-secgrp
           params:
             cluster_id: {{ stack_name }}
       description:
         str_replace:
-          template: Security group for cluster_id OpenShift cluster
+          template: Basic ssh/dns security group for cluster_id OpenShift cluster
```
bogdando: see? :) @tomassedovic

PR #491, opened 2017-06-21
Add node_removal_policies variable to openstack provisioning to allow for scaling down

tzumainn:


playbooks/provisioning/openstack/README.md
```diff
@@ -67,6 +67,9 @@ The `openstack_num_masters`, `openstack_num_infra` and
 `openstack_num_nodes` values specify the number of Master, Infra and
 App nodes to create.
 
+The `openstack_node_removal_policies` allows you to specify which App nodes to
+remove.
```
tomassedovic: So when I first tried this, I put in the names of the servers (e.g. `app-node-1.openshift.example.com`) and that didn't do anything. What needs to go here is the numerical index (e.g. `1`) of the node.

Can you update the description here to make that clear (and maybe give an example?)

Eventually, I'd like us to support the hostnames as well, but this is fine for now.

tomassedovic: Oh, and I'd also like to call it something more obvious, such as: "nodes_to_remove". Ideally, our users shouldn't need to know a ton about openstack and heat.

playbooks/provisioning/openstack/provision-openstack.yml
```diff
@@ -23,6 +23,7 @@
     num_nodes: "{{ openstack_num_nodes }}"
     num_infra: "{{ openstack_num_infra }}"
     num_dns: "{{ openstack_num_dns | default(1) }}"
+    node_removal_policies: "{{ openstack_node_removal_policies | to_yaml }}"
```
tomassedovic: can you put `"{{ openstack_node_removal_policies | default([]) | to_yaml }}"` here?

Without the `default` filter, everyone has to put this in their inventory even if they don't use it.

playbooks/provisioning/openstack/sample-inventory/group_vars/all.yml
```diff
@@ -12,6 +12,8 @@ openstack_num_masters: 1
 openstack_num_infra: 1
 openstack_num_nodes: 2
 
+openstack_node_removal_policies: []
```
tomassedovic: Can you put a short comment here describing what is this for?

Also, with the `default` fix in the provision-openstack playbook, this can be commented out (so please do that).

tomassedovic:
Thanks! I've tested this and it seems to work great \o/.

I have a few suggestions inline if you could take a look please.

playbooks/provisioning/openstack/README.md
tzumainn: Done!

playbooks/provisioning/openstack/provision-openstack.yml
tzumainn: Ah, that makes sense - done!

playbooks/provisioning/openstack/sample-inventory/group_vars/all.yml
tzumainn: Yep, done!

tzumainn:
Thanks for the review!  I've made the changes you suggested - let me know if there's anything else that should be changed!

tomassedovic:
I'm testing the changes now and will merge assuming they work out.

PR #492, opened 2017-06-22
Add ansibledeployocp to support ansible deployment of arm template

glennswest:
Add scripts to support deployment of the Azure OCP Reference architecture from ansible.


e-minguez:
Regarding the test playbook, I'd say to rework it as it contains a few hardcoded values (I didn't have the chance to do it)

glennswest:
Rename it as example ?

e-minguez:
Ok to me (and create a card to follow up the testing and don't forget about it)

glennswest:
Actually the testing had a card 
I use the script with the test every run as well

PR #495, opened 2017-06-23
WIP Openstack provisioning history

tomassedovic:
This preserves the history of the openstack provisioning playbooks over the move from `playbooks/openshift` to `playbooks/provisioning/openstack`.


tomassedovic:
Preserving history across `git mv` as requested by @oybed @etsauer 

I'm honestly unsure we want to do this. The files were originally moved with `git mv` and the history works just fine if you do `git log --follow -- filename` (as opposed to a vanilla `git log`). The fact that github doesn't know how to handle standard moves is not a good reason for us to bend over backwards to show it in the UI.

But it's here if y'all want it.

cc @cooktheryan

oybed:
@tomassedovic yeah, this doesn't look correct. For the playbooks, I believe they all have a 3-4 commit history, so not sure why this many commits are showing.

tomassedovic:
@oybed yeah I thought so, too. `git filter-branch` is basically the dark arts so it's quite possible I messed it up somewhere. Would appreciate any suggestions.

But my question stands: do you do something like this every time you move files around inside a repo? Git can resolve the moves as they are, it's just github's UI that's not as sophisticated.

Issue #494, opened 2017-06-23
[Openstack provider] re-apply of the post-provision tasks fails with "Add DNS A records"


bogdando:
If provision playbook re-applied, the task "Add DNS A records" fails with rc:2 and logs looking like:

```
failed: [localhost] (item=({u'key_algorithm': u'hmac-sha256', u'zone': u'openshift.cluster.lc', u'key_name':
u'private-openshift.cluster.lc', u'server': u'<dns_public_ip>', u'key_secret': u'<some_secret>', u'view':
u'private'}, {'ip': u'192.168.99.4', 'hostname': u'master-0', 'type': u'A'})) => {"changed": false, "dns_rc": 
2, "dns_rc_str": "SERVFAIL", "failed": true, "item": [{"key_algorithm": "hmac-sha256", "key_name": 
"private-openshift.cluster.lc", "key_secret": "<some_secret>", "server": "<dns_public_ip>", "view": 
"private", "zone": "openshift.cluster.lc"}, {"hostname": "master-0", "ip": "192.168.99.4", "type": "A"}], 
"msg": "Failed to create DNS record (rc: 2)"}
```

tomassedovic:
I've verified this. Running the provision playbook and then re-running the same thing without changes fails updating the DNS records.

The nsupdate keys are different each time.

tomassedovic:
I think we absolutely want to keep the provision playbooks idempotent. So I propose we require that the users generate their keys up front for now (and document how to do it in the readme as well as the sample `all.yml`)

We will want to autogenerate the keys going forward, but I'd like to have that in a separate playbook/role that would make sure the keys doesn't change across runs.

/cc @oybed

bogdando:
I found there is named startup issues preventing it from reading zones:
```
Jun 23 14:33:44 host-192-168-99-12 named[937]: zone openshift.cluster.lc/IN/public: journal rollforward failed: journal out of sync with zone
Jun 23 14:33:44 host-192-168-99-12 named[937]: zone openshift.cluster.lc/IN/public: not loaded due to errors.
```

bogdando:
Perhaps the fix will be both updating facts to catch up new keys AND do some dances with jorunal files as well

oybed:
@bogdando as mentioned before, these roles are moving out of `contrib` and it has already been changed in the other repo:
https://github.com/redhat-cop/infra-ansible/pull/24

oybed:
@tomassedovic FYI ^^

[...]

PR #498, opened 2017-06-23
[WIP] Make idempotent zone updates

bogdando:
* Refresh nsupdates facts for dns node to fix obsoleted secrets
  used for nsupdates tasks issued against localhost (control node)
* Purge zone journal files if zone db files has been changed by
  the dns-server role applied.

Closes https://github.com/openshift/openshift-ansible-contrib/issues/494

### DIFF
--- a/playbooks/provisioning/openstack/post-provision-openstack.yml
+++ b/playbooks/provisioning/openstack/post-provision-openstack.yml
@@ -35,6 +35,13 @@
   roles:
   - role: dns-server
 
+- name: Refresh facts for nsupdate keys
+  hosts: dns
+  tasks:
+  - setup:
+      gather_subset: 'facter'
+      filter: nsupdate_keys
+
 - name: Build and process DNS Records
   hosts: localhost
   gather_facts: False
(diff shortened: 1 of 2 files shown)


bogdando:
@tomassedovic @oybed PTAL (I'm aware now of https://github.com/redhat-cop/infra-ansible/pull/24 but this still has to be fixed for the contrib repo anyway)

oybed:
@bogdando it will be a nightmare to maintain the same roles in more than one place, so really don't want these changes to go in (especially since I gave a heads-up at time of initial merge that these roles will be moved out of contrib)

bogdando:
@oybed agree. We will rework the roles to be consumed as external modules from your repos perhaps

oybed:
Also "purging journal files" is very risky in general as records may get lost. The better approach is to freeze the journal files before a change and "thaw" them afterwards. The change made in `infra-ansible`, however, is to avoid changes at all - i.e.: if a key already exists, don't touch it. (we may want to introduce a flag in the future to "refresh" keys when/if needed but that's a TBD)

bogdando:
Superseded by https://github.com/openshift/openshift-ansible-contrib/pull/529

PR #500, opened 2017-06-24
GCP: Allow for custom VPC subnet

bdurrow:
We need our cluster hosts to be in a different CIDR range than the default so we use a custom subnet.  This allows others to do the same with a simple addition to their config.yaml


pschiffe:
Merged as #530

PR #501, opened 2017-06-26
Add ISSUE/PR github templates

bogdando:

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,92 @@
+<!-- Thanks for filing an issue! Before hitting the button, please answer these questions.-->
+
+**Is this a BUG REPORT or FEATURE REQUEST?** (choose one):
+
+<!--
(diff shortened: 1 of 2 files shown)


bogdando:
@tomassedovic

tomassedovic:
Pinging @cooktheryan (please add anyone else who might have a feedback on this, Ryan!).

I like the PR template (seems to come from casl-ansible and I really love their "how should this be tested" idea).

But the issues one feels incredibly heavy-handed and not necessarily providing a lot of benefit.

Also note that there's a lot of different work going into this repository (for gce, aws, vmware, scripts etc. related to the various reference architectures, etc.) so for a lot of the issues here things like Ansible or Shade are not applicable whatsoever.

bogdando:
@cooktheryan PTAL ^^

cooktheryan:
@bogdando and @tomassedovic I am ok with this as sometimes things that come in are sometimes closer to feature requests than issues. I +1 this

Issue #485, opened 2017-06-20
add ICMP to security groups that create the ocp instances


rlopez133:
The playbooks in https://github.com/openshift/openshift-ansible-contrib/tree/master/playbooks/provisioning/openstack , when creating the instances don't have ICMP enabled on the security groups. If you could please add that as not being able to ping systems can make it hard to troubleshoot.

@tomassedovic

bogdando:
will do that as a follow up to that https://github.com/openshift/openshift-ansible-contrib/pull/488 . Adding ICMP to the common sec group (and dns secgroup) will make more sens after that patch merged.

PR #502, opened 2017-06-26
Modify sec groups for provisioned openstack servers

bogdando:
Drop ingress DNS rules from the common secgrp.
Add an ingress ICMP rule, restricted by the ssh ingress cidr,
to the common secgrp. This allows to ping servers from the
control node (ansible admin node).
Add dns servers into the common secgrp as well.

Closes https://github.com/openshift/openshift-ansible-contrib/issues/485

### DIFF
--- a/roles/openstack-stack/templates/heat_stack.yaml.j2
+++ b/roles/openstack-stack/templates/heat_stack.yaml.j2
@@ -152,7 +152,7 @@ resources:
             cluster_id: {{ stack_name }}
       description:
         str_replace:
-          template: Basic ssh/dns security group for cluster_id OpenShift cluster
+          template: Basic ssh/icmp security group for cluster_id OpenShift cluster
           params:
             cluster_id: {{ stack_name }}
       rules:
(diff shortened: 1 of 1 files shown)


bogdando:
@tomassedovic

roles/openstack-stack/templates/heat_stack.yaml.j2
```diff
@@ -162,13 +162,8 @@ resources:
...
         - direction: ingress
-          protocol: tcp
-          port_range_min: 53
-          port_range_max: 53
-        - direction: ingress
-          protocol: udp
-          port_range_min: 53
-          port_range_max: 53
+          protocol: icmp
+          remote_ip_prefix: {{ ssh_ingress_cidr }}
```
tomassedovic: Can we add another rule that makes the hosts pingable within the network as well (i.e. master can ping the infra, etc.). We do a similar thing for the DNS rules.

tomassedovic: So the nodes are actually pingable from each other. Fair.

PR #504, opened 2017-06-26
WIP: Disable swap for Azure

e-minguez:
If restaring waagent makes the installation fail, then do not restart it but leave the swap disabled in the configuration file as well as disabled with swapoff, that way, even if you restart the node, the swap will be disabled.

Tested by redeploying an Azure environment (and modifying the template to fit the github user variable), and it seems to be working:

```
[root@node01 ~]# grep -R -i "swap" /etc/waagent.conf 
# Create and use swapfile on resource disk.
ResourceDisk.EnableSwap=n
# Size of the swapfile.
ResourceDisk.SwapSizeMB=2048
[root@node01 ~]# free -m
              total        used        free      shared  buff/cache   available
Mem:          28136         381       25893           8        1860       27323
Swap:             0           0           0
```

### DIFF
--- a/reference-architecture/azure-ansible/azuredeploy.json
+++ b/reference-architecture/azure-ansible/azuredeploy.json
@@ -276,8 +276,8 @@
     }
   },
   "variables" : {
-    "gituser" : "openshift",
-    "branch" : "master",
+    "gituser" : "e-minguez",
+    "branch" : "azure-swap",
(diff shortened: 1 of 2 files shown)


e-minguez:
I've tried to reboot 1 node and it didn't come back... but with other and setting other parameter (swapsize=0) seems to work...

PR #505, opened 2017-06-26
GCP: Refactor data disk(s) creation and setup to instance template

bdurrow:
This work brings the GCP deployment closer to instances groups launched instances being functional.  It also supports a ./gcloud.sh --scaleup target I'm working on.

PR #508, opened 2017-06-26
GCP: Allow finer control of node instance and disk sizes

bdurrow:
Dependent on https://github.com/openshift/openshift-ansible-contrib/pull/505
Backward compatible with config.yaml written prior to this patch giving the old behavior.


pschiffe:
Resubmitted as #606

PR #506, opened 2017-06-26
GCP: Add --scaleup flag to gcloud.sh

bdurrow:

PR #509, opened 2017-06-27
GCP:custom cluster network cidr

bdurrow:
Make the osm_cluster_network_cidr configurable with a default of 172.16.0.0/16

This branch also includes all commits from https://github.com/openshift/openshift-ansible-contrib/pull/506


pschiffe:
Hi @bdurrow . This PR is really not needed any more, as you can override any ansible variable in the config.yaml file by default without any changes needed (just copy/paste it there and modify it's value). Also, I don't want to pollute the `config.yaml` file if not really necessary.

BTW you can also pass any ansible variable to the `gcloud.sh` command directly, for example `./gcloud.sh -e cluster_network_cidr='172.16.0.0/16'`

PR #512, opened 2017-06-27
Put back node/flat secgrp for infra nodes on openstack

bogdando:
Partially undo 2028883e936c8a1a0be031a19d531d0804a32b68
to unblock end-to-end deployments

### DIFF
--- a/roles/openstack-stack/templates/heat_stack.yaml.j2
+++ b/roles/openstack-stack/templates/heat_stack.yaml.j2
@@ -674,6 +674,12 @@ resources:
           net:         { get_resource: net }
           subnet:      { get_resource: subnet }
           secgrp:
+# TODO(bogdando) filter only required node rules into infra-secgrp
+{% if openstack_flat_secgrp|bool %}
+            - { get_resource: flat-secgrp }
+{% else %}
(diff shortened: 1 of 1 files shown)


roles/openstack-stack/templates/heat_stack.yaml.j2
```diff
@@ -674,6 +674,11 @@ resources:
           net:         { get_resource: net }
           subnet:      { get_resource: subnet }
           secgrp:
+# TODO(bogdando) filter only required node rules into infra-secgrp
+{% if openstack_flat_secgrp|bool %}
+            - { get_resource: flat-secgrp }
+{% else %}
+            - { get_resource: node-secgrp }
```
tomassedovic: This is missing an `{% endif %}`

tomassedovic:
I've tested this with putting `{% endif %}` between `{ get_resource: infra-secgrp }` and `{ get_resource: common-secgrp }` and I was able to build an app and access it via its route.

roles/openstack-stack/templates/heat_stack.yaml.j2
bogdando: done

PR #513, opened 2017-06-27
GCP: remove public IPs from cluster nodes; nat through bastion

bdurrow:
This branch also includes commits from https://github.com/openshift/openshift-ansible-contrib/pull/505 and https://github.com/openshift/openshift-ansible-contrib/pull/508.  I can rebase to exclude most of those commits but I need at least the cloud-init commit from https://github.com/openshift/openshift-ansible-contrib/pull/505.

Possible security risk. Google metadata service is at 169.254.169.254. If someone can get a packet to this server with the destination addresss of 169.254.169.254 it would allow them to see the metadata from this instance. I couldn’t figure out how to do this so I couldn’t test possible remedies.

I should also make this fault tolerant but that would require quite a bit more configuration; perhaps a little later after I have had a discussion with Peter about how to handle load balancers moving forward.


Fale:
Since #505 got then transformed in #537 and merged, I think this PR will need a rebase/conflict check

pschiffe:
So, when looking at this, I have mixed feelings. I know that before this was a recommended and documented way how to remove external ip addresses from instances in GCE, but it's not any more. Now you are supposed to use Organization Policy [1] (which is still beta). Now, I don't like any of these two approaches, both seem to be too complicated, and the result doesn't feel to be worth it. @bdurrow if you want, please take a look at the Organization Policy and let me know how difficult would it be to implement it, but I'm not going to invest my time into this.

[1] https://cloud.google.com/compute/docs/ip-addresses/reserve-static-external-ip-address#disableexternalip

pschiffe:
Closing for now, feel free to reopen with more info.

PR #516, opened 2017-06-28
Adds aws credentials cfn stack lookup

propyless:
Adds (what I believe is) missing aws credentials as an input to cfn-outputs task.

without this I have not been able to use the playbooks that generate the ansible inventory or validates the openshift deployment.

### DIFF
--- a/reference-architecture/aws-ansible/playbooks/roles/cfn-outputs/tasks/main.yaml
+++ b/reference-architecture/aws-ansible/playbooks/roles/cfn-outputs/tasks/main.yaml
@@ -3,6 +3,8 @@
   cloudformation_facts:
     stack_name: "{{ stack_name }}"
     region: "{{ region }}"
+    aws_access_key: "{{ aws_access_key }}"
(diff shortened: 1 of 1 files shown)


propyless:
Will probably need to update the reference architecture documentation as well since that fails to mention the need for aws credentials.

cooktheryan:
@propyless I'm not sure if this is required because the values should be exported by the local system before attempting to perform this

propyless:
Hi yeah I had them exported. But it didn't read those environment
variables. That's why I added this.

propyless:
the same is required for the validation scripts. It is also unable to fetch data from aws..

propyless:
I've verified it twice.

https://github.com/openshift/openshift-ansible-contrib/blob/master/reference-architecture/aws-ansible/playbooks/library/cloudformation_facts.py

This module thats used in the anisble playbooks doesn't load/read the environment variables for AWS at least.

propyless:
hve you been able to verify this as well?

PR #518, opened 2017-06-28
Playbook prerequisites.yml checks that prerequisites are met before provisioning

Tlacenka:
cc @tomassedovic

### DIFF
--- a/playbooks/provisioning/openstack/prerequisites.yml
+++ b/playbooks/provisioning/openstack/prerequisites.yml
@@ -9,8 +9,7 @@
     register: ansible_result
   - name: Check Ansible version
     assert:
-      that:
-        - '{{ ansible_result.stdout }} >= 2.3'
+      that: '{{ ansible_result.stdout }} >= 2.3'
       msg: "Ansible version must be at least 2.3"
 
   # Check shade
@@ -20,19 +19,16 @@
     register: shade_result
   - name: Check if shade is installed
     assert:
-      that:
-        - '{{ shade_result.rc }} == 0'
+      that: '{{ shade_result.rc }} == 0'
       msg: "Python module shade is not installed"
 
   # Check python-dns
   - name: Try to import python DNS module
     command: python -c "import dns"
-    fail: msg: "Python module python-dns is not installed"
     register: pythondns_result
   - name: Check if python-dns is installed
     assert:
-      that:
-        - '{{ pythondns_result.rc }} == 0'
+      that: '{{ pythondns_result.rc }} == 0'
       msg: "Python module python-dns is not installed"
 
   # Check jinja2
@@ -42,8 +38,7 @@
     register: jinja_result
   - name: Check if jinja2 is installed
     assert:
-      that:
-        - '{{ jinja_result.rc }} == 0'
+      that: '{{ jinja_result.rc }} == 0'
       msg: "Python module jinja2 is not installed"
 
   # Check Glance image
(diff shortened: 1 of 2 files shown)


playbooks/provisioning/openstack/prerequisites.yml
```diff
@@ -0,0 +1,72 @@
...
+    ignore_errors: yes
+    register: jinja_result
+  - name: Check if jinja2 is installed
+    assert:
+      that: '{{ jinja_result.rc }} == 0'
+      msg: "Python module jinja2 is not installed"
+
+  # Check Glance image
+  - name: Try to show image
+    shell: openstack image show {{ openstack_default_image_name }}
```
tomassedovic: Please take a look at this module: http://docs.ansible.com/ansible/os_image_facts_module.html

If it doesn't work, use the `command` module here instead of `shell`. We don't do any piping here so it's not necessary.

playbooks/provisioning/openstack/prerequisites.yml
```diff
@@ -0,0 +1,72 @@
...
+    ignore_errors: yes
+    register: image_result
+  - name: Check that image is available
+    assert:
+      that: '{{ image_result.rc }} == 0'
+      msg: "Image {{ openstack_default_image_name }} is not available"
+
+  # Check keypair
+  - name: Try to show keypair
+    shell: openstack keypair show {{ openstack_ssh_public_key }}
```
tomassedovic: There's this module but I'm not sure we can use it to check for a keypair's presense: http://docs.ansible.com/ansible/os_keypair_module.html

If we can't, please use the `command` module here.

playbooks/provisioning/openstack/prerequisites.yml
```diff
@@ -0,0 +1,72 @@
...
+    ignore_errors: yes
+    register: key_result
+  - name: Check that keypair is available
+    assert:
+      that: '{{ key_result.rc }} == 0'
+      msg: "Keypair {{ openstack_ssh_public_key }} is not available"
+
+  # Check network name
+  - name: Try to show network
+    shell: openstack network show {{ openstack_external_network_name }}
```
tomassedovic: Can you try this? http://docs.ansible.com/ansible/os_networks_facts_module.html

Again, if not applicable, please use `command` instead of `shell` here.

playbooks/provisioning/openstack/prerequisites.yml
```diff
@@ -0,0 +1,71 @@
...
+      name: "{{ openstack_external_network_name }}"
+    register: network_result
+  - name: Check that network is available
+    assert:
+      that: "network_result.ansible_facts.openstack_networks"
+      msg: "Network {{ openstack_external_network_name }} is not available"
+
+  # Check keypair
+  - name: Try to show keypair
+    command: openstack keypair show {{ openstack_ssh_public_key }}
```
tomassedovic: I've just realized that we don't actually use the python-openstackclient anywhere in the playbooks and since it's a bunch of complex dependencies, I'd like it to stay that way.

Turns out, it's reasonably simple to do this with `shade` (which we do depend on) directly. Just replace the command above with this one:

    python -c 'import shade; cloud = shade.openstack_cloud(); exit(cloud.get_keypair("{{ openstack_ssh_public_key }}") is None)'

Sorry to ask for another change but could you do that, please?

tomassedovic: Oh and please add a `TODO` comment explaining that there's no Ansible module to do this now and we don't want to force the python-openstackclient dependency.

PR #520, opened 2017-06-29
Enable dnsmasq or it fails resolving k8s svc

e-minguez:
Trying to deploy metrics, the hawkular-metrics pod failed to deploy:

```
[edu@bastion ~]$ oc logs -f hawkular-metrics-05mmk
2017-06-29 09:28:50 Starting Hawkular Metrics
Error: the service account for Hawkular Metrics does not have permission to view resources in this namespace. View permissions are required for Hawkular Metrics to function properly.
Usually this can be resolved by running: oc adm policy add-role-to-user view system:serviceaccount:openshift-infra:hawkular -n openshift-infra
```

What it was really happening was DNS is broken and it cannot access the k8s svc as it cannot find the hostname (I've tried all the k8s svc combinations, none worked).

I think other OCP components (like router/registry) uses the service IP instead the hostname so they were running properly...


cooktheryan:
@e-minguez if you tested this with success then im good with it

e-minguez:
I've tested it and it works (otherwise the monitoring doesn't work as it cannot resolve kubernetes.default.svc) @glennswest do you mind to approve it and merge it if it worked for you? Thanks.

PR #522, opened 2017-06-29
Persist DNS configuration for nodes for openstack provider

bogdando:
* Firstly, provision a Heat stack with given public resolvers.
* After the DNS node configured as an authoritative server,
  switch the Heat stack's Neutron subnet to that resolver
  (private_dns_server) the way it to become the first entry pushed
  into the  hosts /etc/resolv.conf. It will be serving the cluster
  domain requests for OpenShift nodes and workloads.
* Drop post-provision /etc/reslov.conf nameserver hacks as not
  needed anymore.
* Fix dns floating IPs output and add the priv IPs output as well.
* Update docs, clarify localhost vs servers requirements, add
  required Network Manager setup step.
* Use post-provision task names instead of comments.

Closes https://github.com/openshift/openshift-ansible-contrib/issues/455


bogdando:
This is blocked by https://bugs.launchpad.net/heat/+bug/1241553 . While the bug states it was fixed since Havana, the described update scenario is exactly that I'm doing in this patch as well, but it doesn't work.

bogdando:
And we can't use the subnet update CLI command as it brings a bloat of openstack/neutron clients dependencies for the control node. Let's try with the shade library then! (thanks @tomassedovic for an idea!) The tale of a simple task becoming a nightmare.

bogdando:
@tomassedovic works for me \o/

PR #523, opened 2017-06-29
Use wait_for_connection for the Heat nodes

tomassedovic:
The `wait_for_connection` module is more reliable as it uses Ansible's `ping`
to verify the nodes are really accessible. Using `wait_for` and checking that
port 22 is open runs into the possibility of SSH being up but the public keys
or users not being set up yet (as that's done with cloud-init).

In addition, we were gathering facts before running the wait_for task which
rendered it useless.

### DIFF
--- a/playbooks/provisioning/openstack/provision-openstack.yml
+++ b/playbooks/provisioning/openstack/provision-openstack.yml
@@ -43,7 +43,7 @@
   become: False
   gather_facts: False
   tasks:
-    - wait_for_connection:
+  - wait_for_connection:
 
 - hosts: cluster_hosts
   gather_facts: True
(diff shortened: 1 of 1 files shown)


playbooks/provisioning/openstack/provision-openstack.yml
```diff
@@ -39,14 +39,18 @@
   - meta: refresh_inventory
 
 - hosts: cluster_hosts
+  name: Wait for the the nodes to come up
+  become: False
+  gather_facts: False
```
bogdando: then please set gather `Facts: true` as well

tomassedovic: @bogdando we can't gather facts before the nodes are accessible (and `gather_facts` happens before `tasks`  so before `wait_for_connection` happens).

If we need to gather facts in this particular section, we can add this task after wait_for_connection:

      - name: Gather facts for the created nodes
        setup:

But why would we need to? Nothing uses them in this particular play, and they're gathered right after this one.

PR #530, opened 2017-07-03
GCP: Allow for custom VPC subnet II

pschiffe:
#### What does this PR do?
This is slightly modified PR #500 . It enables you to specify custom VPC subnet when creating network in GCP.

In config file configure custom subnet, e.g. `gce_vpc_custom_subnet_cidr: '10.160.0.0/20'`, then verify that the subnet was created and that instances are connected to it:
```
$ gcloud compute networks subnets list
NAME                           REGION        NETWORK            RANGE
ocp-peter-europe-west1-subnet  europe-west1  ocp-peter-network  10.160.0.0/20

$ gcloud compute instances describe ocp-peter-bastion --format='value(networkInterfaces.subnetwork)'
https://www.googleapis.com/compute/v1/projects/ose-refarch/regions/europe-west1/subnetworks/ocp-peter-europe-west1-subnet
```

Just the PR #500 

cc: @bdurrow @e-minguez  PTAL

### DIFF
--- a/reference-architecture/gce-cli/config.yaml.example
+++ b/reference-architecture/gce-cli/config.yaml.example
@@ -86,4 +86,5 @@ node_docker_disk_size: 25
 node_openshift_disk_size: 50
 
 # Custom VPC Subnet, example value: '10.160.0.0/20'
+# Default value is empty, when random subnet in form of 10.x.0.0/20 will be used
 gce_vpc_custom_subnet_cidr: ''
--- a/reference-architecture/gce-cli/config.yaml.example
+++ b/reference-architecture/gce-cli/config.yaml.example
@@ -24,10 +24,6 @@ delete_gold_image: false
 gcloud_project: 'project-1'
 gcloud_zone: 'us-central1-a'
 
-# Custom Subnet
-# If not set will use default subnets in custom VPC
-gce_vpc_custom_subnet_cidr: 10.160.0.0/20
-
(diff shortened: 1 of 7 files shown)


bdurrow:
Thanks @pschiffe, The reason that I chose to do things like this in my PRs was so that people using an old config.yaml would get the previously expected behavior:

gce_vpc_custom_subnet_cidr: {{ gce_vpc_custom_subnet_cidr|default('') }}

Based on your commits that is not something that is worth making the code more complex for.  I'll keep this in mind in the future.

pschiffe:
@bdurrow exactly. Thanks.

reference-architecture/gce-cli/config.yaml.example
```diff
@@ -84,3 +84,6 @@ master_boot_disk_size: 35
 node_boot_disk_size: 25
 node_docker_disk_size: 25
 node_openshift_disk_size: 50
+
+# Custom VPC Subnet, example value: '10.160.0.0/20'
+gce_vpc_custom_subnet_cidr: ''
```
e-minguez: If the variable is empty as the default value, it will create a new subnet, am I right? (I didn't even deploy a thing in GCP yet)
If that's the case, I'd say to add an explanation as "if it is empty, it will create a new subnet with this address a.b.c.d/x"

pschiffe:
@e-minguez I've added better description.

PR #532, opened 2017-07-06
Make enabling/disabling repos idempotent

ghyde:
#### What does this PR do?
When enabling RHEL repos, Ansible disables all repos and then enables only those that are listed in `openshift_required_repos`. It does this every time the playbook is executed. This patch only disables those repos not listed in ` openshift_required_repos`, and it only enables those repos than aren't already enabled.

In addition to making this play idempotent, it also shortens run times by not having to run `subscription-manager` every time the play is ran.

`vagrant provision admin1`

The first time this is executed, extra repos will be disabled, and missing repos will be enabled. The second time this is executed, no changes will be made.

No

cc: @tomassedovic @bogdando @cooktheryan PTAL


ghyde:
This was implemented by @dav1x in commit 9ce488459753a79381cdd15f1fe14e39a46fa7f8.

PR #535, opened 2017-07-10
Options for bastion, SSH config, static inventory autogeneration

bogdando:
#### What does this PR do?

* At the provisioning stage, allow users to auto-generate SSH config,
  when using a static inventory.
* Run playbooks to provsion and post-provision as a separate, when
  using a bastion. This re-applies the SSH config, which ansible can't
  do on the fly.
* Support a pre-installed bastion node, colocated with the 1st infra
  node.
* With a bastion enabled, reduce floating IP footprint to infra and
  dns nodes only, effectively isolating a cluster in a private
  network.

Depends-on https://github.com/openshift/openshift-ansible-contrib/pull/549

Manual e2e deployments by the existing guide with and w/o a bastion node enabled and `inventory: static`.

cc: @tomassedovic @oybed PTAL


playbooks/provisioning/openstack/post-provision-openstack.yml
```diff
@@ -1,70 +1,93 @@
 ---
+- hosts: cluster_hosts
+  name: Wait for the the nodes to come up
+  become: False
+  gather_facts: False
+  tasks:
+    - ping:
```
tomassedovic: Ping is going to fail if the hosts are not up, isn't it? This should be moved to later (but why is it necessary at all?).

playbooks/provisioning/openstack/openstack_dns_records.yml
```diff
@@ -1,25 +1,52 @@
 ---
 - name: "Generate list of private A records"
   set_fact:
-    private_records: "{{ private_records | default([]) + [ { 'type': 'A', 'hostname': hostvars[item]['ansible_hostname'], 'ip': hostvars[item]['openstack']['private_v4'] } ] }}"
+    private_records: >
```
tomassedovic: As far as I can tell, this (and a lot of other changes in this PR) the same code as before, just formatted differently.

These pollute the actual changes and make the review much harder to do thoroughly. Could you please create a separate PR for formatting changes or at least put them in a separate commit?

tomassedovic:
I've only had a brief look so far.

Since we're removing the GPL'd openstack.py inventory, the INVENTORY-LICENSE.txt file as well as the GPL notice in README should be removed.

I like that we're doing that but this PR really does three separate things and that honestly makes it quite difficult to test and review.

Can we split the static inventory, bastion node and floating IPs to separate pull requests?

tomassedovic:
Running this (I've just copied the sample inventory, updated the image/network/RHN credentials) with the default values for everything prints out this error:

```
TASK [static-inventory : set_fact for openstack inventory bastion (infra) nodes] **************************************************************************************************************
Tuesday 11 July 2017  10:41:45 +0000 (0:00:00.115)       0:00:58.474 ********** 

TASK [static-inventory : Add cluster nodes w/o floating IPs to inventory] *********************************************************************************************************************
Tuesday 11 July 2017  10:41:45 +0000 (0:00:00.031)       0:00:58.505 ********** 
fatal: [localhost]: FAILED! => {"failed": true, "msg": "The conditional check 'not item in registered_nodes_floating' failed. The error was: error while evaluating conditional (not item in registered_nodes_floating): Unable to look up a name or access an attribute in template string ({% if not item in registered_nodes_floating %} True {% else %} False {% endif %}).\nMake sure your variable name does not contain invalid characters like '-': argument of type 'StrictUndefined' is not iterable\n\nThe error appears to have been in '/home/centos/openshift-ansible-contrib/roles/static-inventory/tasks/openstack.yml': line 28, column 7, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n    - name: Add cluster nodes w/o floating IPs to inventory\n      ^ here\n"}

```

playbooks/provisioning/openstack/post-provision-openstack.yml
bogdando: added for debug, perhaps it's not needed at all, right

bogdando:
>  Add cluster nodes w/o floating IPs to inventory

I reproduced it with a centos7 control node running in the same tenant network with the cluster nodes. Will investigate it.

bogdando:
The shade inventory part goes here https://github.com/openshift/openshift-ansible-contrib/pull/538

bogdando:
Do not merge yet

bogdando:
e2e test fails with `Error scheduling default router-1-deploy: pod (router-1-deploy) failed to fit in any node`

bogdando:
@tomassedovic e2e works for me

bogdando:
@tomassedovic rebased , it's ready for testing now

tomassedovic:
With the bastion, I'm seeing that it skips the "OpenShift Pre-Requisites (part 2)" play, which means no NetworkManager configuration happens.

I couldn't figure out why yet. Any idea?

```
PLAY [OpenShift Pre-Requisites (part 2)] ******************************************************************************************************************************************************
skipping: no hosts matched

PLAY RECAP ************************************************************************************************************************************************************************************
localhost                  : ok=40   changed=18   unreachable=0    failed=0   
```

roles/static_inventory/tasks/openstack.yml
```diff
@@ -40,8 +44,25 @@
       add_host:
         name: '{{ item.name }}'
         groups: '{{ item.metadata.group }}'
-        ansible_host: '{{ item.public_v4 }}'
+        ansible_host: "{% if use_bastion|bool %}{{ item.name }}{% else %}{{ item.private_v4 }}{% endif %}"
         ansible_fqdn: '{{ item.name }}'
+        ansible_user: '{{ ssh_user }}'
         ansible_private_key_file: '{{ private_ssh_key }}'
+        ansible_ssh_extra_args: '-F {{ ssh_config_path }}'
```
tomassedovic: Right, we can't omit it for all servers with floating IPs because infra and masters still need a floating IP even in the bastion scenario.

So yeah, the ssh_extra_args should be set only when we have bastion enabled.

roles/static_inventory/tasks/openstack.yml
```diff
@@ -40,8 +44,25 @@
       add_host:
         name: '{{ item.name }}'
         groups: '{{ item.metadata.group }}'
-        ansible_host: '{{ item.public_v4 }}'
+        ansible_host: "{% if use_bastion|bool %}{{ item.name }}{% else %}{{ item.private_v4 }}{% endif %}"
```
bogdando: indeed, thanks!

roles/static_inventory/tasks/openstack.yml
bogdando: note, masters are going to not have floating IPs as the patch suggests, please double check requirements? e2e works well for me w/o floating IPs for masters...

playbooks/provisioning/openstack/README.md
```diff
@@ -173,24 +173,38 @@ variables for the `inventory/group_vars/OSEv3.yml`, `all.yml`:
...
+### Configure static inventory and access via a bastion node
 
 Example inventory variables:
 
+    openstack_use_bastion: true
+    bastion_ingress_cidr: "{{openstack_subnet_prefix}}.0/24"
     openstack_private_ssh_key: ~/.ssh/openshift
     openstack_inventory: static
     openstack_inventory_path: ../../../../inventory
+    openstack_ssh_config_path: /tmp/ssh.config.openshift.ansible
```
bogdando: makes sense

roles/openstack-stack/templates/heat_stack.yaml.j2
```diff
@@ -540,7 +553,11 @@ resources:
     properties:
       count: {{ num_masters }}
       resource_def:
+{% if use_bastion|bool %}
+        type: server_nofloating.yaml
```
bogdando: it works for me with a single master w/o a floating IP. How could I check the case for which it fails?

tomassedovic:
Oh, I see. The post-provision-openstack playbook must be run separately for the changes to pick up. I didn't understand that correctly.

Re-running now.

bogdando:
@tomassedovic done

roles/static_inventory/tasks/openstack.yml
```diff
@@ -16,22 +16,26 @@
...
     - name: set_fact for openstack inventory nodes
       set_fact:
+        registered_bastion_nodes: "{{ (registered_nodes_output.stdout | from_json) | json_query(q) }}"
         registered_nodes_floating: "{{ (registered_nodes_output.stdout | from_json) | json_query(q2) }}"
       vars:
         q: "[] | [?metadata.group=='infra.{{stack_name}}']"
         q2: "[] | [?metadata.clusterid=='{{stack_name}}'] | [?public_v4!='']"
       when:
         - refresh_inventory|bool
+        - use_bastion|bool
```
bogdando: fixed in the upper level commit

tomassedovic:
Thanks!

I was a little surprised that this treats an infra node as a bastion -- I expected it to create a separate server. We should probably add that as an option (every time I've seen something used as a bastion, it was a standalone node).

It's also a little unexpected that accessing the routes works from the outside but not from within the bastion: `dig cakephp-mysql-example-test.apps.openshift.example.com` gives the infra floating IP when I access it from the ansible control node, but it has no entry when I run it from inside the bastion node.

Regarding the floating IPs for masters: to access the UI or API, you now need to log in to the bastion or proxy through it. But I expect that the deployers would want to have it accessible directly just like the app routes. So for HA masters, the load balancer would have a floating IP and for a single master, it would have a floating IP.

But we can discuss all that in separate patches.

bogdando:
@tomassedovic 

> It's also a little unexpected that accessing the routes works from the outside but not from within the bastion

good for us, https://github.com/openshift/openshift-ansible-contrib/pull/591 fixes that!

oybed:
@tomassedovic @bogdando are you guys familiar with our docker image that use as a control host for provisioning environments?
https://github.com/redhat-cop/casl-ansible/tree/master/docker/control-host-openstack

If not, I'd suggest giving that one a try as it's a very light weight and stable approach. 

The bastion host @bogdando is referring to may be the one part of our infra-ansible repo, and that's for a different (more generic) purpose that we just use to have a "remote environment" and we often run the above docker control host on top of those.

bogdando:
@oybed nice to know, we should explore that option as we often use a remote environment for ansible control node as well

bogdando:
@tomassedovic wrt

>  to access the UI or API, you now need to log in to the bastion or proxy through it.

there is a patch to proxy access for UI https://github.com/openshift/openshift-ansible-contrib/pull/557
other ports, like API may be added into the tunnel as well. All it needs is a one-time sudo action for the  admin/control host to install and run a tunneling service.

> So for HA masters, the load balancer would have a floating IP

Indeed, for HA masters API and UI may be directly accessed via the LB node.

PR #538, opened 2017-07-12
Generate static inventory with shade inventory

bogdando:
#### What does this PR do?

* Autogenerate inventory/hosts when 'inventory: static' (Default),
  with the shade-inventory tool.
* Drop unused anymore: openstack.py and associated GPL notes,
  an example static inventory, omit manual updates for the
  inventory DNS names in the deployment guide.
* Switch openstack.py formatted inventory hostvars
  to the shade-inventory format (omit openstack.* from hostvars).
* Populate node labels from inventory vars instead of the heat
  templates combined with inventory vars.
* Add app (k8s minions) nodes group for primary node labels.

e2e manual testing, with the `inventory: static`

cc: @tomassedovic  PTAL


tomassedovic:
I'm getting an error when it comes to provisioning the DNS (see below).

Looking at the code, this doesn't seem to have anything related to shade now. Looks like something's missing and we're no longer creating the inventory.

Also, `inventory: static` doesn't seem to be used in anywhere in the code.

```
PLAY [Refresh Server inventory] ***************************************************************************************************************************************

PLAY [Wait for the the nodes to come up] ******************************************************************************************************************************
skipping: no hosts matched

PLAY [cluster_hosts] **************************************************************************************************************************************************
skipping: no hosts matched

PLAY [Assign hostnames] ***********************************************************************************************************************************************
skipping: no hosts matched

PLAY [Subscribe DNS Host to allow for configuration below] ************************************************************************************************************
skipping: no hosts matched

PLAY [Determine which DNS server(s) to use for our generated records] *************************************************************************************************

TASK [dns-server-detect : fail] ***************************************************************************************************************************************
Friday 14 July 2017  10:13:28 +0000 (0:00:00.259)       0:03:40.753 *********** 
fatal: [localhost]: FAILED! => {"failed": true, "msg": "The conditional check 'hostvars[groups['dns'][0]] is undefined' failed. The error was: error while evaluating conditional (hostvars[groups['dns'][0]] is undefined): 'dict object' has no attribute 'dns'\n\nThe error appears to have been in '/home/centos/openshift-ansible-contrib/roles/dns-server-detect/tasks/main.yml': line 2, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n---\n- fail:\n  ^ here\n"}

PLAY RECAP ************************************************************************************************************************************************************
localhost                  : ok=27   changed=10   unreachable=0    failed=1   
```

playbooks/provisioning/openstack/sample-inventory/group_vars/OSEv3.yml
```diff
@@ -5,7 +5,7 @@ openshift_master_default_subdomain: "apps.{{ env_id }}.{{ public_dns_domain }}"
 
 # NOTE(shadower): do not remove this line, otherwise the default node labels
 # won't be set up.
-openshift_node_labels: "{{ openstack.metadata.node_labels }}"
+openshift_node_labels: "{{ metadata.node_labels }}"
```
tomassedovic: My openshift installation failed with `{{ metadata.node_labels }}: 'metadata' is undefined`.

We're not setting `metadata` anywhere and openshift-ansible/.../byo uses the static inventory so it doesn't see the shade output.

I think we need to set metadata in the generated `hosts` file. Or maybe even just set the node labels directly in the inventory to each group.

tomassedovic: The second option would be something like this:

```
[infra:vars]
openshift_node_labels="{'region': 'infra'}"

[app:vars]
openshift_node_labels="{'region': 'primary'}"
```

bogdando: There is already `nodes` group that takes the app labels, nothing more to add I suppose

PR #539, opened 2017-07-12
Add defaults values for some openstack vars

tomassedovic:
#### What does this PR do?

Ansible shows errors when the `rhsm_register` and
`openstack_flat_secgrp` values are not present in the inventory even
though they have sensible default values.

This makes them both default to false when they're not specified.

Remove `openstack_flat_secgrp` and `rhsm_register` values from your inventory's  `group_vars/all.yml` an run the openstack provisioning playbook (on centos -- since we're not subscribing to red hat).

The deployment should succeed as if both vars were there and set to `false`.

N\A

cc: @bogdando PTAL

### DIFF
--- a/playbooks/provisioning/openstack/post-provision-openstack.yml
+++ b/playbooks/provisioning/openstack/post-provision-openstack.yml
@@ -14,7 +14,7 @@
   become: true
   roles:
   - role: subscription-manager
-    when: hostvars.localhost.rhsm_register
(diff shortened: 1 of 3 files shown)


bogdando:
Please also comment the flat sg defaults in the all.yml. We should keep only mandatory variables  there, so users could deploy with their existing all.yml w/o following alternative options

PR #540, opened 2017-07-12
Add a role to generate a static inventory

bogdando:
#### What does this PR do?

* Add the static-inventory role that configures the `inventory/hosts` file (not used yet).

Not used yet

cc: @tomassedovic PTAL


roles/static_inventory/tasks/openstack.yml
```diff
@@ -0,0 +1,68 @@
...
+      when:
+        - refresh_inventory|bool
+
+    - name: set_fact for openstack inventory bastion (infra) nodes
+      set_fact:
+        registered_bastion_nodes: "{{ (registered_nodes_output.stdout | from_json) | json_query(q) }}"
+        registered_nodes_floating: "{{ (registered_nodes_output.stdout | from_json) | json_query(q2) }}"
+      vars:
+        q: "[] | [?metadata.group=='infra.{{stack_name}}']"
+        q2: "[] | [?metadata.group=='*{{stack_name}}'] | [?public_v4!='']"
```
bogdando: this has to be fixed, WIP

roles/static_inventory/defaults/main.yml
```diff
@@ -0,0 +1,21 @@
+---
+# Either to checkpoint the dynamic inventory into a static one
+refresh_inventory: True
+inventory: static
+inventory_path: ~/openstack-inventory
```
tomassedovic: Everything below this line should IMO be in a separate PR that adds the bastion node.

roles/static_inventory/tasks/main.yml
```diff
@@ -0,0 +1,10 @@
+---
+- name: Generate in-memory inventory
+  include: openstack.yml
+
+- name: Checkpoint in-memory data into a static inventory
+  include: checkpoint.yml
+
+- name: Generate SSH config for accessing hosts via bastion
```
tomassedovic: This should be added in a PR that adds the bastion node, not here.

roles/static_inventory/tasks/openstack.yml
```diff
@@ -0,0 +1,68 @@
...
+
+    - name: set fact for openstack inventory cluster nodes
+      set_fact:
+        registered_nodes: "{{ (registered_nodes_output.stdout | from_json) | json_query(q) }}"
+      vars:
+        q: "[] | [?metadata.clusterid=='{{stack_name}}']"
+      when:
+        - refresh_inventory|bool
+
+    - name: set_fact for openstack inventory bastion (infra) nodes
```
tomassedovic: This should be in the bastion PR, not here.

roles/static_inventory/tasks/sshconfig.yml
```diff
@@ -0,0 +1,13 @@
+---
```
tomassedovic: This whole file should be in the bastion PR

roles/static_inventory/templates/openstack_ssh_config.j2
```diff
@@ -0,0 +1,21 @@
+Host *
```
tomassedovic: This whole file should be in the bastion PR

roles/static_inventory/defaults/main.yml
bogdando: done

roles/static_inventory/tasks/openstack.yml
```diff
@@ -0,0 +1,51 @@
...
+      with_items: "{{ registered_nodes }}"
+      when: not item in registered_nodes_floating
+      add_host:
+        name: '{{ item.name }}'
+        groups: '{{ item.metadata.group }}'
+        ansible_host: '{{ item.name }}'
+        ansible_fqdn: '{{ item.name }}'
+        ansible_user: '{{ ssh_user }}'
+        ansible_private_key_file: '{{ private_ssh_key }}'
+        ansible_ssh_extra_args: '-F {{ ssh_config_path }}'
```
tomassedovic: This should be optional. The deployment currently fails when `ssh_config_path` is not set.

roles/static_inventory/tasks/openstack.yml
```diff
@@ -0,0 +1,51 @@
...
+      with_items: "{{ registered_nodes_floating }}"
+      when: item in registered_nodes_floating
+      add_host:
+        name: '{{ item.name }}'
+        groups: '{{ item.metadata.group }}'
+        ansible_host: '{{ item.name }}'
+        ansible_fqdn: '{{ item.name }}'
+        ansible_user: '{{ ssh_user }}'
+        ansible_private_key_file: '{{ private_ssh_key }}'
+        ansible_ssh_extra_args: '-F {{ ssh_config_path }}'
```
tomassedovic: This should be optional. The deployment currently fails when `ssh_config_path` is not set.

roles/static_inventory/tasks/checkpoint.yml
```diff
@@ -0,0 +1,11 @@
+---
+- name: create static inventory dir
+  file:
+    path: "{{ inventory_path }}"
```
tomassedovic: Can we replace this with `"{{ inventory_dir }}"` instead? That way, we'll just add `hosts` to the existing inventory the user already set up. I think that will be less confusing than creating a new directory somewhere else.

roles/static_inventory/tasks/openstack.yml
```diff
@@ -0,0 +1,51 @@
...
+
+    - name: Add cluster nodes with floating IPs to inventory
+      with_items: "{{ registered_nodes_floating }}"
+      when: item in registered_nodes_floating
+      add_host:
+        name: '{{ item.name }}'
+        groups: '{{ item.metadata.group }}'
+        ansible_host: '{{ item.name }}'
+        ansible_fqdn: '{{ item.name }}'
+        ansible_user: '{{ ssh_user }}'
```
tomassedovic: Same here, please replace `ssh_user` with `ansible_user`.

roles/static_inventory/tasks/openstack.yml
```diff
@@ -0,0 +1,51 @@
...
+
+    - name: Add cluster nodes w/o floating IPs to inventory
+      with_items: "{{ registered_nodes }}"
+      when: not item in registered_nodes_floating
+      add_host:
+        name: '{{ item.name }}'
+        groups: '{{ item.metadata.group }}'
+        ansible_host: '{{ item.name }}'
+        ansible_fqdn: '{{ item.name }}'
+        ansible_user: '{{ ssh_user }}'
```
tomassedovic: This is failing for me (undefined variable). According to the docs, this used to be `ansible_ssh_user` and should be just `ansible_user` now:

http://docs.ansible.com/ansible/intro_inventory.html

roles/static_inventory/tasks/openstack.yml
```diff
@@ -0,0 +1,51 @@
...
+        ansible_private_key_file: '{{ private_ssh_key }}'
```
tomassedovic: If we use `ansible_ssh_private_key_file` instead of `private_ssh_key` here, the user has more control over it (they can e.g. pass it using the `--private-key` option to ansible-playbook instead of setting it in the inventory).

It should also be optional in case they don't want to store this in the inventory or if they're using ssh-agent.

roles/static_inventory/tasks/openstack.yml
```diff
@@ -0,0 +1,51 @@
...
+        ansible_ssh_extra_args: '-F {{ ssh_config_path }}'
+        private_v4: '{{ item.private_v4 }}'
+
+    - name: Add cluster nodes with floating IPs to inventory
+      with_items: "{{ registered_nodes_floating }}"
+      when: item in registered_nodes_floating
+      add_host:
+        name: '{{ item.name }}'
+        groups: '{{ item.metadata.group }}'
+        ansible_host: '{{ item.name }}'
```
tomassedovic: This can't be `item.name` because the DNS is not set up and populate yet. So accessing the nodes via their hostname fails.

I think it should be set to the floating IP address instead.

roles/static_inventory/tasks/openstack.yml
bogdando: I'd prefer to not make it too complex with many options for users to pick up.

roles/static_inventory/tasks/openstack.yml
bogdando: Well, it doesn't fails when having a static SSH config. That's why I didn't want to split the features into a separtate steps. I hoped to use FQDN names would be more comfortable for users. Given that it works with the static SSH config, no DNS needed.

roles/static_inventory/tasks/checkpoint.yml
bogdando: Do you mean omit the `create static inventory dir` ?

tomassedovic: Yeah. Plus replacing inventory_path with inventory_dir because that's a special variable that points to the current inventory.

bogdando: This would require manual steps, like creating a *right* directory (or the guide will stop working for users with custom inventory paths) and strictly following the guide's described manual steps, which only brings more chaos. I think ansible must take care of it instead, no manual/external configuration steps.

roles/static_inventory/tasks/openstack.yml
tomassedovic: But it should work without the static ssh config, too. One of the reasons I wanted to split the bastion code from the static inventory was so we can see things like these. The SSH config isn't necessary so we shouldn't force it on our users.

roles/static_inventory/tasks/checkpoint.yml
tomassedovic: I don't understand -- what manual steps does it require? If you use the defaults, it'll be just `./inventory` just like now, but if you say `ansible-playbook -i some-other-inventory` that will just work, too (generating `some-other-inventory/hosts`).

Using `inventory_dir` means that ansible will take care of everything and no configuration is necessary.

roles/static_inventory/tasks/openstack.yml
bogdando: It looks not convenient to access inventory nodes via IPs. And nothing can be done if there is no floating IP for a node, yet we have bastion and SSH config...

bogdando: So I'll put private_v4 then for nodes w/o floating IP with a caveat note that control node must belong to the private IP

tomassedovic: The bastion node should be able to access the private IPs (otherwise what's the point of it) and Ansible still prints out the item.name when running. From the point of view of the user running the playbooks nothing changes.

tomassedovic: missed the second comment, sorry. Yeah, please do that. Or if that doesn't work, we can use ssh config in that scenario. But at least with the floating IPs, they should be the ansible_host.

playbooks/provisioning/openstack/README.md
```diff
@@ -118,6 +118,14 @@ The `required_packages` variable also provides a list of the additional
 prerequisite packages to be installed before to deploy an OpenShift cluster.
 Those are ignored though, if the `manage_packages: False`.
 
+The `inventory` controls either a static inventory will be created after the
+cluster nodes provisioned on OpenStack cloud. Note, the fully dynamic inventory
+is yet to be supported, so the static inventory will be created anyway.
+
+The `inventory_dir` points the directory to host the generated static inventory.
```
tomassedovic: You can remove this paragraph. Ansible sets `inventory_dir` automaticaly:

http://docs.ansible.com/ansible/playbooks_variables.html

It's not something the user needs to set up. It will just always point to the inventory directory.

roles/static_inventory/defaults/main.yml
```diff
@@ -0,0 +1,8 @@
+---
+# Either to checkpoint the dynamic inventory into a static one
+refresh_inventory: True
+inventory: static
+inventory_dir: ~/openstack-inventory
```
tomassedovic: You can remove this line now.

bogdando: I'm not sure, that would work as expected with the follow up [PR](https://github.com/openshift/openshift-ansible-contrib/pull/550/files?short_path=9141c3c#diff-9141c3c37ff1e871d38171e178e46e72) . We need this var to control the inventory file placement

tomassedovic: Oh, right. I forgot this is just the role. Right, so we may want to remove `openstack_inventory_dir` from that PR and leave this unchanged.

Let me try that and see.

tomassedovic: Okay, so this now does re-use the existing inventory, but because `inventory_dir` is a special Ansible variable, it overrides the custom vars which means that it's no matter what we set `inventory_dir` to, it will always point to the current inventory.

So I think we should change this back to `inventory_path` (sorry! :-( ) and then in #550 remove (or change the default of) `openstack_inventory_dir`.

tomassedovic:
@bogdando is this ready for merging?

PR #541, opened 2017-07-12
Switch the sample inventory to CentOS

tomassedovic:
#### What does this PR do?
This changes the image name and deployment types to use centos instead
of rhel and sets `rhsm_register` to false.

With these changes, the inventory should be immediately deployable
using the default values (assuming the image, network and flavor names
match).

Ideally, the upstream CI will just end up using this inventory with
little to no changes, too at some point.

Copy the updated sample-inventory, make sure the image, network and flavor names match and test it end to end without any additional changes. The deployment should succeed with Origin running on CentOS servers.

nona

cc: @bogdando PTAL

### DIFF
--- a/playbooks/provisioning/openstack/sample-inventory/group_vars/OSEv3.yml
+++ b/playbooks/provisioning/openstack/sample-inventory/group_vars/OSEv3.yml
@@ -1,5 +1,6 @@
 ---
 openshift_deployment_type: origin
+openshift_release: 1.5.1
 #openshift_deployment_type: openshift-enterprise
 #openshift_release: v3.5
 openshift_master_default_subdomain: "apps.{{ env_id }}.{{ public_dns_domain }}"
--- a/playbooks/provisioning/openstack/sample-inventory/group_vars/OSEv3.yml
+++ b/playbooks/provisioning/openstack/sample-inventory/group_vars/OSEv3.yml
@@ -1,6 +1,7 @@
 ---
-openshift_deployment_type: openshift-enterprise
-openshift_release: v3.5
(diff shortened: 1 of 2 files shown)


playbooks/provisioning/openstack/sample-inventory/group_vars/OSEv3.yml
```diff
@@ -1,6 +1,7 @@
 ---
-openshift_deployment_type: openshift-enterprise
-openshift_release: v3.5
+openshift_deployment_type: origin
+#openshift_deployment_type: openshift-enterprise
+#openshift_release: v3.5
```
bogdando: please add origin_release: 1.5.1

playbooks/provisioning/openstack/sample-inventory/group_vars/all.yml
```diff
@@ -4,7 +4,7 @@ public_dns_domain: "example.com"
 public_dns_nameservers: []
 
 openstack_ssh_public_key: "openshift"
-openstack_default_image_name: "rhel73"
+openstack_default_image_name: "centos7"
 openstack_default_flavor: "m1.medium"
```
bogdando: nit: it seems we may need also, or some subset of it, for CI cases:
``openshift_disable_check: disk_availability,memory_availability,package_availability,docker_image_availability,docker_storage,package_version``

tomassedovic: Yeah, though I don't want to disable these checks in the sample inventory. This is something people trying it out should be made aware of.

The ci script adds this configuration explicitly.

PR #542, opened 2017-07-12
Set up NetworkManager automatically

tomassedovic:
#### What does this PR do?

This removes the extra step of running the
`openshift-ansible/playbooks/byo/openshift-node/network_manager.yml`
before installing openshift. In addition, the playbook relies on a
host group that the provisioning doesn't provide (oo_all_hosts).

Instead, we set up NetworkManager on CentOS nodes automatically. And
we restart it on RHEL (which is necessary for the nodes to pick up the
new DNS we configured the subnet with).

This makes the provisioning easier and more resilient.

Run an end-to-end deployment on RHEL+Enterprise and Centos+Origin without running the network_manager playbooks.

none

cc: @Tlacenka  @bogdando PTAL

### DIFF
--- a/playbooks/provisioning/openstack/post-provision-openstack.yml
+++ b/playbooks/provisioning/openstack/post-provision-openstack.yml
@@ -59,12 +59,20 @@
 
 - name: OpenShift Pre-Requisites
   hosts: OSEv3
-  gather_facts: False
+  gather_facts: true
   become: true
-  tasks:
+  pre_tasks:
   - name: "Include DNS configuration to ensure proper name resolution"
     lineinfile:
       state: present
       dest: /etc/sysconfig/network
       regexp: "IP4_NAMESERVERS={{ hostvars['localhost'].private_dns_server }}"
       line: "IP4_NAMESERVERS={{ hostvars['localhost'].private_dns_server }}"
+  roles:
+  # CentOS isn't using NetworkManager by default. Make sure it's installed:
+  - role: node-network-manager
+    when: ansible_distribution == "CentOS"
+  tasks:
+  - name: Restart NetworkManager to pick up the DNS changes
+    service: name=NetworkManager state=restarted
+    when: ansible_distribution == 'Red Hat Enterprise Linux'
--- a/playbooks/provisioning/openstack/post-provision-openstack.yml
+++ b/playbooks/provisioning/openstack/post-provision-openstack.yml
@@ -69,10 +69,4 @@
       regexp: "IP4_NAMESERVERS={{ hostvars['localhost'].private_dns_server }}"
       line: "IP4_NAMESERVERS={{ hostvars['localhost'].private_dns_server }}"
   roles:
-  # CentOS isn't using NetworkManager by default. Make sure it's installed:
-  - role: node-network-manager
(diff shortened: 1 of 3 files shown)


playbooks/provisioning/openstack/post-provision-openstack.yml
```diff
@@ -59,12 +59,20 @@
...
+  pre_tasks:
   - name: "Include DNS configuration to ensure proper name resolution"
     lineinfile:
       state: present
       dest: /etc/sysconfig/network
       regexp: "IP4_NAMESERVERS={{ hostvars['localhost'].private_dns_server }}"
       line: "IP4_NAMESERVERS={{ hostvars['localhost'].private_dns_server }}"
+  roles:
+  # CentOS isn't using NetworkManager by default. Make sure it's installed:
+  - role: node-network-manager
```
bogdando: where is this role comes from?

playbooks/provisioning/openstack/post-provision-openstack.yml
```diff
@@ -59,12 +59,20 @@
...
+    when: ansible_distribution == "CentOS"
+  tasks:
+  - name: Restart NetworkManager to pick up the DNS changes
```
bogdando: if the role has a handler, we could include it for any OS type and make the service notified

tomassedovic:
Setting this to [WIP] because I didn't get to test this thoroughly yet.

playbooks/provisioning/openstack/post-provision-openstack.yml
tomassedovic: I wrote it so from this pull request. It's basically the same thing as the network_manager.yml plabook does in openshift-ansible but turned into a role.

Theoretically, we could move the role there and make sure both playbooks use it, but it's tiny so I'm not sure it matters.

playbooks/provisioning/openstack/post-provision-openstack.yml
tomassedovic: The role doesn't have a handler yet, but we could add one.

I didn't want to run the role for RHEL, but I guess there's no harm in it. Should I do that?

bogdando: I think so, it simplifies the code

tomassedovic:
Alright, the end-end for centos/origin worked for me. Testing rhel/enterprise now.

tomassedovic:
Got a successful end to end for rhel & enterprise, too. Removing the WIP tag, this is ready for reviews and merging now.

bogdando:
Could we reuse the `openshift-ansible/playbooks/byo/openshift-node/network_manager.yml` instead of adding more roles to maintain?

tomassedovic:
Not as it stands right now. It's a playbook not a role so we can't import it with ansible-galaxy like we do the DNS stuff. Plus we really need "state=restarted" not just "started" for RHEL.

We could move the `node-network-manager` role to openshift-ansible and update our code and their `network_manager.yml` to use it (if they're happy with that).

But even then, I'd prefer we merge this now and replace it with ansible-galaxy when that role is merged.

What do you think?

PR #546, opened 2017-07-13
Replace greaterthan and equalto in openstack-stack

tomassedovic:
#### What does this PR do?

The `greaterthan` and `equalto` Jinja filters were added in 2.8 which is notably not packaged in
CentOS and RHEL. This removes them in favour of the `==` and `>` operators
which are available in Jinja 2.7.

Make sure your jinja2 version is 2.7 (e.g. by running `pip uninstall Jinja2 && pip install --user Jinja2=2.7.2`) and run the provisioning part of the deployment (full end to end isn't strictly speaking necessary).

cc: @bogdando

### DIFF
--- a/roles/openstack-stack/templates/heat_stack.yaml.j2
+++ b/roles/openstack-stack/templates/heat_stack.yaml.j2
@@ -432,7 +432,7 @@ resources:
           port_range_min: 53
           port_range_max: 53
           remote_ip_prefix: "{{ openstack_subnet_prefix }}.0/24"
-{% if num_masters is greaterthan 1 %}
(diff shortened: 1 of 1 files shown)


tomassedovic:
@bogdando this is the jinja version bit we talked about. I misremembered it but it boils down to:

a) we have four lines of code that require jinja 2.8
b) centos and rhel both package jinja 2.7

So it's much easier to change it in our side and to be honest, I prefer `==` and `>` to `equalto` and `greaterthan`.

What do you think?

bogdando:
would be nice to fix this version for the requirements of the control node as well

PR #365, opened 2017-05-16
[WIP] Removed emptydir & unified variables per provider

e-minguez:

PR #547, opened 2017-07-13
Move OCP variables to one place

pschiffe:
#### What does this PR do?
Moves OCP variables to one place.

Verify that the deployment finished successfully.

PR #506

cc: @bdurrow @e-minguez @cooktheryan PTAL


e-minguez:
👏 that's what I wanted to do [here](https://github.com/openshift/openshift-ansible-contrib/pull/365/files) but I failed every time. It would be nice to have the same approach on every other provider 🤔

cooktheryan:
@e-minguez its like we almost just had this discussion :D

pschiffe:
@e-minguez yeah, ansible sometimes just sucks.. I think that this solution is stupid, but it's probably the only one which works..

bdurrow:
I think that this opens the possibility replace having the variables spelled out in the static-inventory template also.  I'm not sure a simple pattern matching would be enough but I imagine there is a robust YAML->JSON conversion method.

pschiffe:
If you find something reasonable, just let me know. I don't know any better way how to do it now, and I just don't want to throw time at it as current state is fine.

PR #550, opened 2017-07-14
Static inventory autogeneration

bogdando:
#### What does this PR do?
* At the provisioning stage, allow users to auto-generate a static
  inventory w/o manual steps needed. The alternative to
  go fully dynamic TBD.
* Move openshift pre-install playbook to the post provision playbook,
  where the second part of the pre install tasks is already placed.

Depends-on https://github.com/openshift/openshift-ansible-contrib/pull/538

e2e with `inventory: static`

cc: @tomassedovic PTAL

### DIFF
--- a/playbooks/provisioning/openstack/provision-openstack.yml
+++ b/playbooks/provisioning/openstack/provision-openstack.yml
@@ -8,6 +8,10 @@
     - include: pre_tasks.yml
   roles:
     - role: openstack-stack
+    - role: static_inventory
+      when: openstack_inventory|default('static') == 'static'
+      inventory_path: "{{ openstack_inventory_path|default(inventory_dir) }}"
+      private_ssh_key: "{{ openstack_private_ssh_key|default('~/.ssh/id_rsa') }}"
 
 - name: Refresh Server inventory
   hosts: localhost
@@ -16,20 +20,5 @@
   gather_facts: False
   tasks:
     - meta: refresh_inventory
-
-- hosts: cluster_hosts
-  name: Wait for the the nodes to come up
-  become: False
-  gather_facts: False
-  tasks:
-    - wait_for_connection:
-
-- hosts: cluster_hosts
-  gather_facts: True
-  tasks:
-    - name: Debug hostvar
-      debug:
-        msg: "{{ hostvars[inventory_hostname] }}"
-        verbosity: 2
 
 - include: post-provision-openstack.yml
--- a/playbooks/provisioning/openstack/README.md
+++ b/playbooks/provisioning/openstack/README.md
@@ -93,8 +93,9 @@ steps, and the Neutron subnet for the Heat stack is updated to point to that
 server in the end. So the provisioned servers will start using it natively
 as a default nameserver that comes from the NetworkManager and cloud-init.
 
-`openstack_ssh_key` is a Nova keypair -- you can see your keypairs with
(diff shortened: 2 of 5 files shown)


bogdando:
e2e fails with `openstack.metadata.node_labels` dependencies (the format of the openstack.py based inventory):
```
     Play:     Gather and set facts for node hosts
     Task:     openshift_facts
```

bogdando:
It seems we need some dynamic inventory wrapper script around shade inventory, just like openstack.py does, but much more simple. It should just put the `private_v4`, `public_v4` and `metadata` dict that shade creates in the root path, into the `openstack.` path. Something like this.

tomassedovic:
Yeah, I suspected that would be the case. Alternatively, we could move the node label setting out of heat and into our playbooks.

bogdando:
AFAICT, the etcd group must not be empty. I can't tell how that have been working before, but testing shows for me that the empty etcd group ends up with no etcd deployed at all.

bogdando:
@tomassedovic e2e works for me

tomassedovic:
Triggered a "CI" run on my fork:

https://travis-ci.org/tomassedovic/openshift-ansible-contrib/jobs/255565791

tomassedovic:
It failed because of an empty etcd group. That's addressed in #559 so I've created a new run with that PR applied:

https://travis-ci.org/tomassedovic/openshift-ansible-contrib/jobs/255571124

playbooks/provisioning/openstack/provision-openstack.yml
```diff
@@ -8,6 +8,10 @@
     - include: pre_tasks.yml
   roles:
     - role: openstack-stack
+    - role: static_inventory
+      when: openstack_inventory|default('static') == 'static'
+      inventory_path: "{{ openstack_inventory_path|default('../../../../inventory') }}"
```
tomassedovic: This will only work with the default inventory at `~/inventory` and needs to be modified every time that location is changed. Your deployers will likely maintain multiple inventories and will be passing them to ansible-playbook via `-i path/to/inventory`.

If you change the line to this:

    inventory_path: "{{ openstack_inventory_path|default(inventory_dir) }}"

It will work both with the current `~/inventory` situation from README and with any value passed to `-i`.

playbooks/provisioning/openstack/provision-openstack.yml
```diff
@@ -8,6 +8,10 @@
...
+      private_ssh_key: "{{ openstack_private_ssh_key|default('~/.ssh/openshift') }}"
```
tomassedovic: Because the README now removes the `--private-key` call from `ansible-playbook`, the default key should be `~/.ssh/id_rsa`.

tomassedovic:
Oh also: this PR seems to have some of the changes #538 has. Do we need both or does this one have everything, now?

playbooks/provisioning/openstack/provision-openstack.yml
bogdando: it is specified in the inventory, it may be any of the values

bogdando:
Let's first merge https://github.com/openshift/openshift-ansible-contrib/pull/538 then I'll rebase that one

tomassedovic:
Running the "CI" here: https://travis-ci.org/tomassedovic/openshift-ansible-contrib/jobs/255674668

PR #552, opened 2017-07-14
Retry tasks in the subscription manager role

Tlacenka:
#### What does this PR do?
Some subscription manager role tasks (such as subscription or attaching pool) are now retried by ansible, so that even during slow internet connection, the chance of failure is decreased.

1) To make sure that subscription manager tasks fails and is retried, enter wrong credentials in your inventory used for provisioning.
2) Run provisioning using rhel images.
3) Ansible will retry to subscribe using wrong credentials and fail at the end.

-------------


bogdando:
nit: I wonder if the `block:` wrapper could work as well?

pschiffe:
@bogdando `block` doesn't support loops, at least yet.

Issue #558, opened 2017-07-18
Etcd inventory group must be not empty


bogdando:
# BUG REQUEST INFO:

**Environment**: static inventory
- **Cloud provider:** openstack

An empty etcd group ends up deploying no etcd nodes, when testing the static inventory [PR](https://github.com/openshift/openshift-ansible-contrib/pull/550)
I can't tell how that empty etcd group ever worked otherwise, with the current master. A mystery.

**Anything else do we need to know**:

The fix seems to include masters group into the etcd group

PR #559, opened 2017-07-18
Include masters into etcd group, when it is empty

bogdando:
#### What does this PR do?
Include masters into etcd group, when it's empty

e2e manual testing

Closes https://github.com/openshift/openshift-ansible-contrib/issues/558

cc: @tomassedovic PTAL

### DIFF
--- a/roles/static_inventory/templates/inventory.j2
+++ b/roles/static_inventory/templates/inventory.j2
@@ -44,6 +44,7 @@ masters.{{ stack_name }}
 
 [etcd:children]
 etcd.{{ stack_name }}
+{% if 'etcd' not in groups or groups['etcd']|length == 0 %}masters.{{ stack_name }}{% endif %}
 
 [nodes:children]
 masters


roles/static_inventory/templates/inventory.j2
```diff
@@ -44,6 +44,7 @@ masters.{{ stack_name }}
 
 [etcd:children]
 etcd.{{ stack_name }}
+{% if groups['etcd']|length == 0 %}masters.{{ stack_name }}{% endif %}
```
tomassedovic: In my testing, the `groups` didn't even have `etcd` so this step failed on undefined variable.

Changing it to this seems to work:

    {% if 'etcd' not in groups or groups['etcd']|length == 0 %}masters.{{ stack_name }}{% endif %}

bogdando: the `[etcd:children]` placeholder, makes it always to be in groups, isn't it?

tomassedovic: I would think so, but it's not the case. This is the error I was getting without it:

```
TASK [static_inventory : create inventory from template] ***********************
Thursday 20 July 2017  09:14:42 +0000 (0:00:00.026)       0:01:37.329 ********* 
fatal: [localhost]: FAILED! => {"changed": false, "failed": true, "msg": "AnsibleUndefinedVariable: 'dict object' has no attribute 'etcd'"}
```

https://travis-ci.org/tomassedovic/openshift-ansible-contrib/jobs/255581702#L894

PR #561, opened 2017-07-19
During provisioning, make unnecessary packages optional under a switch

Tlacenka:
#### What does this PR do?
The `openshift-prep` role installs packages that are not necessary for the deployment, such as `bash-completion`n and `vim-enhanced`.
These are now optional under `install_debug_packages` switch, which is set to `false` by default.

1) Run provisioning with both `manage_packages` and `install_debug_packages` set to `true`.
- Result: all packages will be installed and updated, even the debug ones.

2) Run provisioning with `manage_packages` set to `true`, `install_debug_packages` set to `false`.
- Result: only required packages will be installed and updated.

--


tomassedovic:
Thanks! Looks good to me, I'll test it later & merge.

One question: how did you decide which packages are required and which aren't? Did you test it or just pick out the obvious ones?

Tlacenka:
Without having much of a deeper understanding of what happens during provisioning, these are the obvious choices. The rest looks like it could be used for configuring networking or downloading additional packages/images.

tomassedovic:
Yep, no worries. Just wondered. We can take a deeper look at this later (like, I'm guessing that more of these are not really necessary but that needs to be tested)

tomassedovic:
Triggered a "CI run" here:

https://travis-ci.org/tomassedovic/openshift-ansible-contrib/jobs/255273651

Will merge once it passes.

tomassedovic:
Oops, the "CI" didn't actually run the end to end test. How embarrassing!

Take two: https://travis-ci.org/tomassedovic/openshift-ansible-contrib/jobs/255280243

PR #563, opened 2017-07-19
Set ansible_become for the OSEv3 group

tomassedovic:
#### What does this PR do?

Because openshift-ansible requires root on the cluster nodes, but it
doesn't explicitly set it in the playbooks (like we do), let's set it
in our inventory instead of requiring to pass `--become` to
`ansible-playbook`.

That will simplify the installation steps as well as let us include
the provisioning and openshift-ansible playbooks in a single playbook.

Run end to end without passing `--become` to the openshift-ansible byo playbook (i.e. like it says in the readme now).

cc: @bogdando @Tlacenka  PTAL

### DIFF
--- a/playbooks/provisioning/openstack/README.md
+++ b/playbooks/provisioning/openstack/README.md
@@ -181,7 +181,7 @@ file, this is how you stat the provisioning process:
 
 Once it succeeds, you can install openshift by running:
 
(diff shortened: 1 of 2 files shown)


Tlacenka:
It works for me :) And I agree that it is useful for including `openshift-ansible` playbooks.

PR #571, opened 2017-07-21
Upscaling OpenShift application nodes

Tlacenka:
#### What does this PR do?
It scales up number of application nodes by a value predefined by the user.
When value is not defined, the current number of app nodes is increased by 1.
When there is no deployment to scale up, the value becomes the target number of app nodes.

Note: Always remove `hosts` file from your inventory when deleting a deployment.

**Test it on an existing deployment**:
1) Run `oc get nodes -l autoscaling=app` on your master node.
- If the result is empty, it means that you haven't completed a deployment where this label gets set. Therefore, you have to delete your current deployment and run a new end to end deployment that sets the autoscaling label first (which is done automatically, so no extra steps during the development are required).
- If the result displays current application nodes, you're all set.
2) Now that you have met the prerequisites, run the `openshift-ansible-contrib/playbooks/provisioning/openstack/scale-up.yaml` playbook.
- In order to scale up your deployment by a custom amount, add `-e increment_by=<number greater or equal to 1>` when executing `ansible-playbook`.
- You can customise the path to your `openshift-ansible` directory by adding `-e openshift_ansible_dir=<path to openshift-ansible>`. The value must be an absolute path and cannot contain '/' at the end.
3) Check that the deployment and playbook pre/post-verifications were successful. Try assigning a workload to the newly created nodes.

**Test it without an existing deployment**:
1) If you have a deployment, delete it.
2) Run the `openshift-ansible-contrib/playbooks/provisioning/openstack/scale-up.yaml` playbook.
- In order to create a custom amount of nodes (1 by default), add `-e increment_by=<number greater or equal to 1>` when executing `ansible-playbook`.
- You can customise the path to your `openshift-ansible` directory by adding `-e openshift_ansible_dir=<path to openshift-ansible>`. The value must be an absolute path and cannot contain '/' at the end.
3) Check that the deployment and playbook pre/post-verifications were successful. Try assigning a workload to the newly created nodes.

--

cc: @tomassedovic @bogdando @tzumainn  PTAL


playbooks/scale-up.yaml
```diff
@@ -0,0 +1,43 @@
...
+- include: ../../openshift-ansible/playbooks/byo/config.yml
+
+# Verify new number of nodes
+- hosts: masters
+  tasks:
+  - name: Get number of nodes
+    shell: oc get nodes | grep app-node -c
+    register: oc_new_result
+  - name: Check that value has been updated
+    when: 'nodes | int != oc_new_result.stdout | int'
```
tomassedovic: I would slightly prefer using the `assert` module here because we're asserting that the scaling actually happened. But it's not a big deal.

playbooks/scale-up.yaml
```diff
@@ -0,0 +1,43 @@
...
+- hosts: masters
+  tasks:
+  - name: Get number of nodes
+    shell: oc get nodes | grep app-node -c
+    register: oc_old_result
+  - name: Set new number of nodes
+    set_fact:
+      # Number of nodes in the new configuration (incremented by 1 by default)
+      nodes: "{{ oc_old_result.stdout | int + 1 }}"
+  - name: Check that new value is higher
```
tomassedovic: Can we also verify that the `oc get nodes` value matches the existing `openstack_num_node`?

Basically making sure that whatever's set in the vars actually represents the reality.

playbooks/scale-up.yaml
```diff
@@ -0,0 +1,43 @@
...
+  tasks:
+  - name: Set increased number of nodes for provisioning and installation
+    set_fact:
+      openstack_num_nodes: "{{ nodes }}"
+
+# Run provision.yaml with higher number of nodes to create a new app-node VM
+- include: provisioning/openstack/provision.yaml
+
+# Run config.yml to update openshift installation (add the new node(s))
+- include: ../../openshift-ansible/playbooks/byo/config.yml
```
tomassedovic: This is good for now, but eventually, we'll probably want to control this via an ansible variable. Could you add a TODO for that?

playbooks/scale-up.yaml
bogdando: or just use the openstack_num_node value. Inventory vars may be only source of trueth

Tlacenka:
@bogdando When I prompt for `openstack_num_nodes`, instead of the current number of app nodes in OpenShift, it returns the value of said variable in my inventory, which is not updated after a successful deployment, at least not in my experience.

bogdando:
@Tlacenka do you mean that the initial openstack_num_nodes value might change because of the autoscaling events or the like? Then setting a fact to evaluate it totally makes sens to me!

playbooks/scale-up.yaml
bogdando: I'm confused now. If the number of nodes may change runtime, it makes no sense to verify that the `oc get nodes` value matches the existing `openstack_num_node`? Otherwise, if it never changes run-time, it doesn't make sense to run `oc get nodes`, as it will be  always equal to the `openstack_num_node`...

Tlacenka:
@bogdando  No, that's not what I mean.

I now tried printing `openstack_num_nodes` value.

Note: `openstack server list` shows me 3 app nodes, even the broken `oc get nodes` shows me 5 entries (even if it had counted all other types of nodes, it would have only been 4) so that means that the nodes are there.

But, using the following playbook:

```
---
- hosts: masters
  tasks:
  - debug:
      var: openstack_num_nodes
```
and running it like so:
```
ansible-playbook --user openshift --private-key ~/.ssh/openshift -i <inventory path> <file containing the playbook>
```

I get this output:
```
ok: [master-0.openshift.example.com] => {
    "openstack_num_nodes": 1
}
```

This value is the value set in my `all.yml` file in my inventory, so if I used this value before rerunning provisioning + OpenShift installation, I would have to count on the user having the right value in his inventory file. Therefore, I concluded that using `oc get nodes` is the only way of getting the actual number of nodes.

Correct me if I'm wrong.

bogdando:
I thinks that the `openstack_num_nodes` should be the only source of truth then. It should be fixed to match real app nodes numbers, and `oc get nodes` should be omitted.

bogdando:
`openstack server list` does not know about app nodes AFAICT. The node labels we want for app nodes only makes sense, right @tomassedovic ? could you please comment on that topic?

Tlacenka:
when I deployed it now with it checking both the number `oc get nodes` returns and current `openstack_num_nodes` value on `masters` host, only `oc get nodes` returned the right value...so not sure if I'm doing it wrong or if it's the desired behaviour

tomassedovic:
@Tlacenka can you paste the output of `oc get nodes`, `openstack server list` and the `openstack_num_nodes` variable? Just so we have something concrete to look at.

After the initial deployment, openstack_num_nodes should be equal to the number of app nodes (so without infra and masters). If it's not, that's super weird and we ought to figure out why.

So what @bogdando's trying to say is to make `openstack_num_nodes` the value that always defines the number of nodes, right? So scaling would always involve changing the inventory (say by increasing openstack_num_nodes) and then running the provisioning & openshift-ansible playbooks, right?

I think that's worth exploring, but then we really need to get to the bottom of why the values seem to diverge.

bogdando:
> So what @bogdando's trying to say is to make openstack_num_nodes the value that always defines the number of nodes, right? So scaling would always involve changing the inventory (say by increasing openstack_num_nodes) and then running the provisioning & openshift-ansible playbooks, right?

Correct. Thank you for clarifying this @tomassedovic !

> I think that's worth exploring, but then we really need to get to the bottom of why the values seem to diverge.

It think it diverge exactly because of the reasons you've mentioned: oc get nodes returns info not only for app nodes, but masters and infras. And openstack server list adds dns nodes there. While all we need to scale up/down *the app nodes* (and to look for) is only the openstack_num_nodes! It is a nice to update the description for the latter, to outline that in fact it is a count for app nodes only.

tomassedovic:
That's true, but the playbook adds a grep: `oc get nodes | grep app-node -c` and this should return the same value as `openstack_num_nodes` unless the cluster was changed outside of ansible.

Tlacenka:
> @Tlacenka can you paste the output of oc get nodes, openstack server list and the openstack_num_nodes variable? Just so we have something concrete to look at.

- `openstack server list` output:

```
+----------------------+----------------------+--------+----------------------+-------------------------+
| ID                   | Name                 | Status | Networks             | Image Name              |
+----------------------+----------------------+--------+----------------------+-------------------------+
| f3eb6212-284a-4880-8 | app-node-1.openshift | ACTIVE | openshift-ansible-   | rhel73                  |
| 0c1-d7cf9b1a6fa1     | .example.com         |        | openshift.example    |                         |
|                      |                      |        | .com-                |                         |
|                      |                      |        | net=192.168.99.6,    |                         |
|                      |                      |        | 38.145.33.243        |                         |
| 63a118e7-e728-4627   | app-node-2.openshift | ACTIVE | openshift-ansible-   | rhel73                  |
| -a81e-14cbbf93f55d   | .example.com         |        | openshift.example    |                         |
|                      |                      |        | .com-                |                         |
|                      |                      |        | net=192.168.99.5,    |                         |
|                      |                      |        | 38.145.33.102        |                         |
| 44a0b14d-f54e-4371-9 | app-node-3.openshift | ACTIVE | openshift-ansible-   | rhel73                  |
| 4d9-d3fbebc15612     | .example.com         |        | openshift.example    |                         |
|                      |                      |        | .com-                |                         |
|                      |                      |        | net=192.168.99.17,   |                         |
|                      |                      |        | 38.145.34.84         |                         |
| 6e8b5cee-32dc-4a82   | infranode-0.openshif | ACTIVE | openshift-ansible-   | rhel73                  |
| -a8bd-d8e8635da617   | t.example.com        |        | openshift.example    |                         |
|                      |                      |        | .com-                |                         |
|                      |                      |        | net=192.168.99.4,    |                         |
|                      |                      |        | 38.145.34.68         |                         |
| 5353abc2-e999-4126-8 | dns-0.openshift.exam | ACTIVE | openshift-ansible-   | rhel73                  |
| 164-0e111440bde5     | ple.com              |        | openshift.example    |                         |
|                      |                      |        | .com-                |                         |
|                      |                      |        | net=192.168.99.8,    |                         |
|                      |                      |        | 38.145.33.190        |                         |
| 69d28494-646e-43e7   | master-0.openshift.e | ACTIVE | openshift-ansible-   | rhel73                  |
| -80ed-7b146d608e87   | xample.com           |        | openshift.example    |                         |
|                      |                      |        | .com-                |                         |
|                      |                      |        | net=192.168.99.7,    |                         |
|                      |                      |        | 38.145.34.55         |                         |
| 03cafd40-91ec-4f00   | app-node-0.openshift | ACTIVE | openshift-ansible-   | rhel73                  |
| -b62b-274422e15c1d   | .example.com         |        | openshift.example    |                         |
|                      |                      |        | .com-                |                         |
|                      |                      |        | net=192.168.99.9,    |                         |
|                      |                      |        | 38.145.34.30         |                         |
| 5f2b2cfa-16a8-412a-  | workaround           | ACTIVE | private=192.168.0.6, | CentOS-7-x86_64-Generic |
| a5a3-1ddfb5ba0b21    |                      |        | 38.145.33.157        | Cloud-1703              |
+----------------------+----------------------+--------+----------------------+-------------------------+
```

- `oc get nodes` output:

```
NAME                                STATUS                     AGE
app-node-0.openshift.example.com    Ready                      1d
app-node-1.openshift.example.com    Ready                      1d
app-node-2.openshift.example.com    Ready                      1d
app-node-3.openshift.example.com    Ready                      1d
infranode-0.openshift.example.com   Ready                      1d
master-0.openshift.example.com      Ready,SchedulingDisabled   1d
```
-  `oc get nodes | grep app-node -c` output is `4`.

- when running the following playbook

```
---
- hosts: masters
  tasks:
  - debug:
      var: openstack_num_nodes

- hosts: localhost
  tasks:
  - debug:
      var: openstack_num_nodes
```

the output is:

```
PLAY [masters] ******************************************************************************************

TASK [Gathering Facts] **********************************************************************************
ok: [master-0.openshift.example.com]

TASK [debug] ********************************************************************************************
ok: [master-0.openshift.example.com] => {
    "openstack_num_nodes": 1
}

PLAY [localhost] ****************************************************************************************

TASK [Gathering Facts] **********************************************************************************
ok: [localhost]

TASK [debug] ********************************************************************************************
ok: [localhost] => {
    "openstack_num_nodes": 1
}
```

the only way I was able to update the variable for the provisioning and installation was to set said value to the defined one right before running these playbooks (line 20 in the playbook file).

As of now, the whole playbook runs successfully up to the point where the equivalence of `nodes` (new value defined by used) and `openstack_num_nodes` var is checked. Turns out that `openstack_num_nodes` returned to its previous value by now. `nodes` and the result of `oc get nodes | grep ...` return the same value for me.

bogdando:
> oc get nodes | grep app-node -c and this should return the same value as openstack_num_nodes
@tomassedovic @Tlacenka that works for me, although I do not understand the case why we want to call `oc` instead of thinking of the inventory vars as the only source of truth. Anyway, please add an assertion to stop the playbook with an error, if those was found unexpectedly *not equal*

Tlacenka:
@bogdando 
It's already there, the current playbook ends with an error for me. The said assert starts at line 45 in the `scale-up.yaml` file.

playbooks/scale-up.yaml
```diff
@@ -0,0 +1,50 @@
...
+  - name: Get number of nodes
+    shell: oc get nodes | grep app-node -c
+    register: oc_old_result
+  - name: Set new number of nodes
+    set_fact:
+      # Number of nodes in the new configuration (incremented by 1 by default)
+      nodes: "{{ oc_old_result.stdout | int + 1 }}"
+  - name: Check that new value is higher
+    assert:
+      that: 'nodes | int > oc_old_result.stdout | int'
```
bogdando: got it, this protects the future playbooks execution from unexpected scale down while running scale-up play, right? And otherwise, the scale-down protects users from doing unexpected scale-up. LGTM now, thanks!

Tlacenka:
@bogdando So have you tried  upscaling your development?

If it didn't fail on comparing `nodes` and `openstack_num_nodes`, it's different from my behaviour and it should be explored.

If it fails for you as well, we need to find out what's happening before this PR can be merged and potentially remove the check in order for it to succeed before merging.

bogdando:
@Tlacenka not yet, I will try it out shortly

Tlacenka:
So, I came to a conclusion that the process needs to be split in two, therefore I added a bash script that runs both playbooks. Manual is updated above.

playbooks/upscaling_pre-tasks.yaml
```diff
@@ -0,0 +1,26 @@
+---
+# Check that new number of nodes is higher
+- hosts: masters
+  tasks:
+  - name: Get number of nodes
+    shell: oc get nodes | grep app-node -c
```
tomassedovic: I'm surprised this doesn't require root (i.e. `become: true`)..

playbooks/upscaling_pre-tasks.yaml
```diff
@@ -0,0 +1,26 @@
...
+      msg: >
+        FAIL: New number of nodes must be higher than the current one
+        ({{ oc_old_result.stdout | int }}).
+
+# Update openstack_num_nodes value in the inventory
+- hosts: localhost
+  tasks:
+  - name: Replace openstack_num_nodes value
+    lineinfile:
+      path: ../../{{ inv_directory }}/group_vars/all.yml
```
tomassedovic: Please replace this with `{{ inventory_dir }}/group_vars/all.yml`. That will always point to the directory that's passed to `ansible-playbook` with `-i`:

http://docs.ansible.com/ansible/latest/playbooks_variables.html

playbooks/upscaling_run.sh
```diff
@@ -0,0 +1,79 @@
...
+  echo "Parameter -i is required. For more information, use -h|--help"
+  exit 1
+fi
+
+# Run upscaling_pre-tasks.yaml to:
+# - run preverification based on entered number of nodes
+# - update openstack_num_nodes variable in inventory
+ansible-playbook $ANSIBLE_PARAMS -i "$INVENTORY" \
+openshift-ansible-contrib/playbooks/upscaling_pre-tasks.yaml \
+-e "$EXTERNAL_VAR" -e "inv_directory=${INVENTORY%/}" $VERBOSE
```
tomassedovic: Please remove the `inv_directory` var in favour of the built-in `inventory_dir`.

playbooks/upscaling_pre-tasks.yaml
```diff
@@ -0,0 +1,26 @@
...
+      regexp: '^openstack_num_nodes:'
+      line: 'openstack_num_nodes: {{ hostvars[groups["masters"][0]]["nodes"] }} '
```
tomassedovic: Not a huge deal, but I think removing the set_fact above and using `{{ oc_old_result.stdout | int + 1 }}` here directly would be a little cleaner. Just a thought, feel free to keep what you have if you like it better.

playbooks/upscaling_pre-tasks.yaml
Tlacenka: yea, me too but it seems to work just fine

playbooks/upscaling_pre-tasks.yaml
Tlacenka: the up-scaling increments current number of nodes 1 ONLY if user does not override it by defining a constant higher than that...so if I removed it and only incremented it by 1 per deployment, it would take several hours for someone to scale it up by...say 10, wouldn't it?

tomassedovic:
In the readme, please also document that it requires `ansible_become: true` in `group_vars/OSEv3.yml`.

I forgot to update my inventory and while the normal deployment works fine, the scaling really needs that value explicitly.

tomassedovic:
So in my testing, the nodes got created, but when I used an existing app and tried to scale its pods, they while they got scheduled to the newly created node, they never ended up running there.

Did you try to schedule a workload on the new nodes?

tomassedovic:
Oh, one more thing, this post-scaling check failed:

```
TASK [Check that the actual number of nodes matches defined value] ****************************************************************************************************************************
Tuesday 01 August 2017  12:56:55 +0000 (0:00:00.364)       0:16:13.840 ******** 
fatal: [master-0.openshift.example.com]: FAILED! => {"failed": true, "msg": "the field 'args' has an invalid value, which appears to include a variable that is undefined. The error was: 'node
s' is undefined\n\nThe error appears to have been in '/home/centos/openshift-ansible-contrib/playbooks/upscaling_scale-up.yaml': line 16, column 5, but may\nbe elsewhere in the file depending
 on the exact syntax problem.\n\nThe offending line appears to be:\n\n    register: oc_new_result\n  - name: Check that the actual number of nodes matches defined value\n    ^ here\n"}
```

I think the `upscaling_scale-up.yaml` playbook should say `{{ oc_new_result }}` instead of `{{ nodes }}`.

playbooks/upscaling_pre-tasks.yaml
Tlacenka: it actually needs to be written like this: `'{{ inventory_dir ~ "/group_vars/all.yml" }}'` (might be useful to save some time in case you need it in the future)

Tlacenka:
@tomassedovic 
I finished running the deployment (successfully) just now and passing -e nodes=N to the second playbook (`upscaling_scale-up.yaml`) seems to be enough for me to use `nodes` variable there. It is strange that it doesn't work for you.

I tried running a pod on the newly created node - it shows the pod on the app node but doesn't show it on master node, which is also strange.

tomassedovic:
Oh, I think I get why `nodes` was a problem for me: I didn't pass `-e nodes=N` but instead relied on the default "increment by one" behaviour. Which means that the `nodes` var was never set for me.

Which means the shell script should either disallow that or make it work.

playbooks/provisioning/openstack/README.md
```diff
@@ -221,6 +221,28 @@ Once it succeeds, you can install openshift by running:
     ansible-playbook openshift-ansible/playbooks/byo/config.yml
 
 
+## Scale Deployment up/down
+
+### Scaling up
+
+1. Make sure that `openshift-ansible-contrib/playbokos/provisioning/openstack/upscaling_run.sh` is executable. If not, run `chmod +x <file>`
```
tomassedovic: Typo: playbokos -> playbooks

playbooks/provisioning/openstack/README.md
```diff
@@ -221,6 +221,28 @@ Once it succeeds, you can install openshift by running:
...
+2. Check that `ansible_become: true` is set in `<inventory>/group_vars/OSEv3.yml`
+3. Go to a directory that contains both `inventory` and `openshift-ansible*` directories.
+4. Run the script:
+
+    ./openshift-ansible-contrib/playbooks/upscaling_run.sh -i <path to inventory>
+    [-e nodes=<new number of nodes>] [-h] [-v]
```
tomassedovic: Github's rendering seems to eat the angle brackets for some reason even though this should be a code block because markdown:

https://github.com/Tlacenka/openshift-ansible-contrib/blob/88460aa65f62e56dca0e04558059f5153e81d9ef/playbooks/provisioning/openstack/README.md

Try wrapping it in blocks of triple backticks like so:

    ```
    ./openshift-ansible-contrib/playbooks/upscaling_run.sh -i <path to inventory> [-e nodes=<new number of nodes>] [-h] [-v]
    ```

playbooks/provisioning/openstack/upscaling_scale-up.yaml
```diff
@@ -0,0 +1,27 @@
...
+  tasks:
+  - name: Get number of nodes
+    shell: oc get nodes | grep app-node -c
+    register: oc_new_result
+  - name: Check that the actual number of nodes matches defined value
+    assert:
+      that: 'nodes | int == oc_new_result.stdout | int'
+      msg: >
+        FAIL: Number of application nodes has not been increased accordingly
+        (it should be {{ nodes }} but it is {{ oc_new_result.stdout | int }}).
```
tomassedovic: `{{ nodes }}` isn't set if you just want to increment the scale count by one (and therefore don't explicitly pass `-e nodes=...`.

bogdando:
@tomassedovic 
> there are 2 usecases: one is manual scaling -- where we shouldn't really update the inventory or do anything weird like that

For autoscaling, it makes sense to add another inventory var, like `auto_scale: False` (defaults) that makes provision and deploy playbooks to completely ignore nodes counts and always evaluate it runtime. You never know how your cluster grows or shrinks. It may be doing so, while you're in the middle of a playbook execution.

For manual scaling, as we all agreed, it seems, let's not update inventory files stored on disk, and basically may be a part of a user repository. Unless we own the whole pipeline, like generating inventory, submitting changes to it on code review etc. Let's just document for users doing manual scaling up/down to update inventory groupvars manually.

Tlacenka:
@bogdando 
> Enabling it makes provision and deploy playbooks to completely ignore nodes counts and always evaluate it runtime. 

I found out that the values are being extracted at [this point](https://github.com/openshift/openshift-ansible-contrib/blob/master/playbooks/provisioning/openstack/stack_params.yaml#L17) in `openshift-ansible-contrib`. How should it evaluate the counts other than from the variable?

bogdando:
As "evaluate it runtime" I mean only using the outputs of the `oc` commands

Tlacenka:
if you mean `oc get nodes`, the hostnames customisation effectively disables that as now the user can change `app-node` to literally anything

bogdando:
@Tlacenka `oc get nodes --show-labels` should always give the correct numbers for app nodes identified by tags. @tomassedovic thoughts?

tomassedovic:
@Tlacenka end to end CI: https://travis-ci.org/openshift/openshift-ansible-contrib/builds/267986163

playbooks/provisioning/openstack/README.md
```diff
@@ -342,6 +342,22 @@ In order to access UI, the ssh-tunnel service will be created and started on the
...
+One can scale up the number of application nodes by executing the ansible playbook
+`openshift-ansible-contrib/playbooks/provisioning/openstack/scale-up.yaml`.
+This process can be done even if there is currently no deployment available.
+The `increment_by` variable is used to specify by how much the deployment should
+be scaled up (if none exists, it serves as a target number of application nodes).
+
+Usage:
+
+```
+ansible-playbook --user openshift --private-key ~/.ssh/openshift -i <path to inventory> openshift-ansible-contrib/playbooks/provisioning/openstack/scale-up.yaml` [-e increment_by=<number>]
```
bogdando: nit: this may be shortened to just `ansible-playbook openshift-ansible-contrib/playbooks/provisioning/openstack/scale-up.yaml` [-e increment_by=<number>]`

playbooks/provisioning/openstack/scale-up.yaml
```diff
@@ -0,0 +1,68 @@
...
+    - hostvars[groups["masters"][0]]["oc_old_result"] is defined
+    register: result
+
+  - name: Update openstack_num_nodes variable (if there was no deployment)
+    set_fact:
+      openstack_num_nodes: "{{ increment_by }}"
+    when: result | skipped
+
+# Run provision.yaml with higher number of nodes to create a new app-node VM
+- include: provision.yaml
```
bogdando: please add a note to docs that this playbook works only w/o a bastion node (`openstack_use_bastion: False`). Otherwise, the static SSH config requires the post-provision step to be executed as a separate ansible-playbook call. For now, let's just add a note. We could fix this later for the bastion use case.

tomassedovic:
@Tlacenka how far did you test this? After scaling up, I tried to run the good old cake php app and the image push failed. Did you see anything like it?

```
OK (1 test, 1 assertion)
Pushing image 172.30.81.196:5000/test/cakephp-mysql-example:latest ...
Warning: Push failed, retrying in 5s ...
Warning: Push failed, retrying in 5s ...
Warning: Push failed, retrying in 5s ...
Warning: Push failed, retrying in 5s ...
Warning: Push failed, retrying in 5s ...
Warning: Push failed, retrying in 5s ...
Warning: Push failed, retrying in 5s ...
Registry server Address: 
Registry server User Name: serviceaccount
Registry server Email: serviceaccount@example.org
Registry server Password: <<non-empty>>
error: build error: Failed to push image: After retrying 6 times, Push image still failed
```

Tlacenka:
@tomassedovic since the openshift-ansible problem isn't afaik fixed yet and I did not use the last working commit cause...I am dumb, I only tested the deployment (post-deployment part failed at pushing image as usual...so isn't the error you are seeing related to that as well?)

tomassedovic:
@Tlacenka yeah, but the error is different and I was using a working branch of openshift-ansible.

tomassedovic:
@Tlacenka FYI if you pull the latest openshift-ansible, it should work agai.

Tlacenka:
@tomassedovic so, I tried the deployment again with the right `openshift-ansible` (hopefully).

- When I tried scaling up without an existing deployment, it was successful and the `cakephp` app was created and running.
- However, when I tried the same with an existing deployment, it failed the same way yours did. No clue why.

Tlacenka:
note: still in progress, I need to do some testing with the `openshift-ansible` scaleup playbook

tomassedovic:
I'll run the e2e CI and we can merge it afterwards.

tomassedovic:
@Tlacenka can you please merge master again? We've had a CI breakage 2 days ago and this pull request doesn't seem to have the fix in.

playbooks/provisioning/openstack/scale-up.yaml
```diff
@@ -0,0 +1,75 @@
...
+      that: 'increment_by | int >= 1'
+      msg: >
+        FAIL: The value of increment_by must be at least 1
+        (but it is {{ increment_by | int }}).
+  - name: Update openstack_num_nodes variable
+    set_fact:
+      openstack_num_nodes: "{{ oc_old_num_nodes | int + increment_by | int }}"
+
+# Run provision.yaml with higher number of nodes to create a new app-node VM
+- include: provision.yaml
```
bogdando: did that miss the post provision playbook to update DNS records for new nodes?

tomassedovic: The comment about bastion is in the readme:

https://github.com/openshift/openshift-ansible-contrib/pull/571/files#diff-9141c3c37ff1e871d38171e178e46e72R590

And `provision.yaml` should include everything -- it's the top-level playbook we use to run deployments. Or am I missing something?

bogdando: oh, sorry then!

PR #574, opened 2017-07-21
README: Added note about infra-ansible installation

Tlacenka:
#### What does this PR do?
Currently, when installing `infra-ansible` and some version of it already exists, this step is skipped. In order to install the newer version, the older one must be removed first. This should be mentioned in the provisioning README.

--

--

### DIFF
--- a/playbooks/provisioning/openstack/README.md
+++ b/playbooks/provisioning/openstack/README.md
@@ -39,8 +39,13 @@ Alternatively you can install directly from github:
     ansible-galaxy install git+https://github.com/redhat-cop/infra-ansible,master \
       -p openshift-ansible-contrib/roles
 
-Note, this assumes we're in the directory that contains the clonned
-openshift-ansible-contrib repo in its root path.
+Notes:
+* this assumes we're in the directory that contains the clonned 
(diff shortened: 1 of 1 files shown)


playbooks/provisioning/openstack/README.md
```diff
@@ -39,8 +39,13 @@ Alternatively you can install directly from github:
     ansible-galaxy install git+https://github.com/redhat-cop/infra-ansible,master \
       -p openshift-ansible-contrib/roles
 
-Note, this assumes we're in the directory that contains the clonned
-openshift-ansible-contrib repo in its root path.
+Notes:
+* this assumes we're in the directory that contains the clonned 
+openshift-ansible-contrib repo in its root path
+* when trying to install a different version, the previous one must be removed first
```
bogdando: please split into sentences

bogdando:
Interesting, if an argument to do an update exists for the `ansible-galaxy install`

tomassedovic:
Interestingly, `ansible-galaxy install` has a `-f` option:

  -f, --force           Force overwriting an existing role

The wording suggests it should update the roles but it doesn't.

PR #575, opened 2017-07-23
Rhv details

cwilkers:
#### What does this PR do?
  * Clean-up of playbooks
  * Added validation and uninstall playbooks
  * Load balancer http/s configs

See Ref Arch (when committed)

N/A

cc: @cooktheryan

### DIFF
--- a/reference-architecture/rhv-ansible/ocp-vars.yaml
+++ b/reference-architecture/rhv-ansible/ocp-vars.yaml
@@ -58,12 +58,19 @@ openshift_master_cluster_public_hostname: openshift.example.com
 # OpenShift Identity Providers
 # htpasswd shown here, other options documented at
 # https://docs.openshift.com/container-platform/3.5/install_config/configuring_authentication.html
-openshift_master_identity_providers: [{'name': 'htpasswd_auth', 'login': 'true', 'challenge': 'true', 'kind': 'HTPasswdPasswordIdentityProvider', 'filename': '/etc/origin/master/htpasswd'}]
+openshift_master_identity_providers:
+  - name: htpasswd_auth
+    login: true
+    challenge: true
+    kind: HTPasswdPasswordIdentityProvider
+    filename: /etc/origin/master/htpasswd
 # Defining htpasswd users
-#openshift_master_htpasswd_users={'user1': '<pre-hashed password>', 'user2': '<pre-hashed password>'}
+#openshift_master_htpasswd_users:
+#  - user1: <pre-hashed password>
+#  - user2: <pre-hashed password>
 # Example with admin:changeme
-openshift_master_htpasswd_users={'admin' : '$apr1$zAhyA9Ko$rBxBOwAwwtRuuaw8OtCwH0'}
-
+openshift_master_htpasswd_users:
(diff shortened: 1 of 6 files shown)


reference-architecture/rhv-ansible/ocp-vars.yaml
```diff
@@ -54,6 +55,25 @@ load_balancer_hostname: openshift-lb.example.com
...
+  - name: htpasswd_auth
+    login: true
+    challenge: true
+    kind: HTPasswdPasswordIdentityProvider
+    filename: /etc/origin/master/htpasswd
+# Defining htpasswd users
+#openshift_master_htpasswd_users:
+#  - user1: <pre-hashed password>
+#  - user2: <pre-hashed password>
+# Example with admin:changeme
```
e-minguez: I'll add instructions how to create those pre-hashed passwords

cwilkers: @e-minguez Good idea, though upstream didn't bother including instructions :-) That can go in the RefArch, yes?

reference-architecture/rhv-ansible/playbooks/openshift-install.yaml
```diff
@@ -60,4 +58,35 @@
         - "role={{ openshift_node_labels.role }}"
     openshift_registry_selector: "role=infra"
     openshift_router_selector: "role=infra"
+    # Load balancer config
```
e-minguez: Are we using the same loadbalancer for api than for apps?

reference-architecture/rhv-ansible/ocp-vars.yaml
e-minguez: I think if it is just a command add it just in case :)

reference-architecture/rhv-ansible/ocp-vars.yaml
```diff
@@ -45,6 +45,7 @@ root_ssh_key:
 deployment_type: openshift-enterprise
 openshift_vers: v3_5
 containerized: false
+console_port: 8443
```
cwilkers: This only got added because the verification roles require it, and fail without it.

reference-architecture/rhv-ansible/playbooks/openshift-install.yaml
cwilkers: Yes, this requires some extra config, but it's do-able.

reference-architecture/rhv-ansible/ocp-vars.yaml
cwilkers: Re-reading, I'd misunderstood... I think a default value would require a role structure. Can you do that for a playbook?

cwilkers:
@e-minguez I just added a couple minor changes from the stakeholders' discussion this morning.

reference-architecture/rhv-ansible/playbooks/openshift-install.yaml
e-minguez: But then it is a SPOF for both API and apps... I think we should discuss if we want a single HAproxy or an HA deployment instead and split API and apps

PR #588, opened 2017-07-26
Added SDN option

e-minguez:
#### What does this PR do?
Allow user choose OpenShift SDN in ARM template

Deploy OCP using the ref. arch. and select any of the provided and supported SDN options and check the value on the /etc/origin/{master|node}/{master|node}-config.yaml file.

Nope

cc: @glennswest  @cooktheryan  PTAL


glennswest:
Looks good

PR #596, opened 2017-07-27
Access UI via a bastion node

bogdando:
#### What does this PR do?

When using a bastion and a single master, use the lb-secgrp
to access UI port allowed from the ingress bastion node cidr.
For HA (masters>1), UI still should be accessed via
the LB node's ingress cidr, omitting the bastion.

e2e test via accessing openshift UI dashboard, with bastion enabled, following the README guide

cc: @tomassedovic @oybed PTAL


bogdando:
@tomassedovic PTAL

tomassedovic:
This had a conflict with the private networks patch. I've resolved it.

tomassedovic:
Started an end to end CI run here:

https://travis-ci.org/openshift/openshift-ansible-contrib/builds/264802297

will review in the meantime

roles/static_inventory/tasks/sshtun.yml
```diff
@@ -0,0 +1,15 @@
+---
+- name: Create ssh tunnel systemd service
+  template:
+    src: ssh-tunnel.service.j2
+    dest: /etc/systemd/system/ssh-tunnel.service
+    mode: 0755
```
tomassedovic: None of my systemd service files need to be executable. Is this really needed?

roles/static_inventory/tasks/sshtun.yml
```diff
@@ -0,0 +1,15 @@
...
+
+- name: reload the systemctl daemon after file update
+  shell: systemctl daemon-reload
```
tomassedovic: This can be a `command` instead. Shell is for when you need pipes or io redirection.

tomassedovic:
end-end ci passed. So yeah let me know if you want to merge this as is or add another commit.

roles/static_inventory/tasks/sshtun.yml
bogdando: For some strange reason, all files in my /etc/systemd/system have +x set

bogdando: done

roles/static_inventory/tasks/sshtun.yml
bogdando: done

roles/static_inventory/tasks/sshtun.yml
tomassedovic: Huh, strange. This seems to suggest that they should be 644 instead:

https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/7/html/System_Administrators_Guide/sect-Managing_Services_with_systemd-Unit_Files.html#proc-Managing_Services_with_systemd-Unit_File_Create

Would you mind trying to see if that works? Could be something weird on your system. But we can also do that as a separate patch if you prefer.

tomassedovic: Oh, you've fixed it in the meantime, sorry! I get confused by the github ui sometimes. Will merge now.

PR #598, opened 2017-07-27
Support scaling up for masters and nodes

### LINKED ISSUE #567: GCP - Ability to scale cluster
FEATURE REQUEST
Every time the script is run it will try to destroy the whole cluster and rebuild it from scratch.
I think it would be very helpful to be able to scale up and down the cluster without having to rebuild it from scratch.

pschiffe:
#### What does this PR do?
Enable scaling up of OCP cluster. To use it, update your `config.yaml` file with new number of required nodes (must be bigger than before, scaling down is not supported), and run `gcloud.sh --scaleup` command.

Try to scale up existing OCP deployment.

Original PR #506 
Resolves #567 

cc: @bdurrow @cooktheryan PTAL


Fale:
Will those scaleups playbook also scaledown if nodes number is lower than current?

pschiffe:
@Fale no, I don't think that scaling down is supported with openshift automatically. You can, of course, evacuate the node and shut it down, but I can't tell whether we are going to support that on ref. arch. level.

I should probably add a check for this to prevent accidental loss of data.

Issue #589, opened 2017-07-26
Private_v4 hostvar may take wrong values for multi-network VMs


bogdando:
# BUG REQUEST INFO:

**Environment**: provisioning a static inventory
- **Cloud provider:** Openstack

- **OS (`printf "$(uname -srm)\n$(cat /etc/os-release)\n"`):** n/a

- **Version of Ansible:** (`ansible --version`): n/a

- **Version of Jinja:** (`pip freeze | grep -i jinja`) n/a

- **Version of Shade:** (`pip freeze | grep -i shade`) 1.21.0

**Openstack-ansible-contrib version (commit) (`git rev-parse --short HEAD`):** ed4eb09

**Copy of used inventory files and custom variables (please omit your secrets!):** n/a

**Command used to invoke ansible**: provision playbook

**Output of ansible run**: n/a

**Anything else do we need to know**:

In openstack dynamic inventory, multiple private interfaces are stored as a hash, which has no ordering guarantees, thus, returns a random value for the private_v4, it seems.
This should be mitigated with a custom json_query to be executed against fetched dynamic inventory to get the private_v4 of the required network from the networks hash.

PR #604, opened 2017-07-28
Support multiple private networks for static inventory

bogdando:
#### What does this PR do?
Support multiple private networks for static inventory / SSH configs generation.
Add openstack_private_network_name to filter by a wanted private network.

Deploy with multiple private networks (update heat template for Nova servers), choose one as an admin/control net and define it as the `openstack_private_network_name`. Finish e2e, it should pass. It should not break things when deploying with a single private network as well.

Closes https://github.com/openshift/openshift-ansible-contrib/issues/589

cc: @tomassedovic @oybed  PTAL


bogdando:
I noted that filtered dynamic inventory repeats nodes for several regions. So added the region contstraint as well. Not sure if this impacts the private_v4 jmespath manipulations for this PR though. But it could, so I'd prefer to submit that as a single unit of work please.

bogdando:
For some strange reason, the networks dict from shade-inventory contains no values. It's looks like a bug. It works as expected with shade 1.21.0 (networks contains values), though it doesn't with 1.22.2 (empty networks). Shall I better off rely on the addresses hash, it shows more consistent results across versions of shade? :-)

bogdando:
@tomassedovic ready for review/testing. works for me.

tomassedovic:
I'm running it now. Sorry for the delay.

tomassedovic:
@bogdando how do I deploy with multiple private networks? Do I need to change the heat templates?

bogdando:
@tomassedovic yes, currently there is no other way to deploy Nova servers with multiple networks. A trivial change TBD though.

roles/static_inventory/defaults/main.yml
```diff
@@ -19,3 +19,6 @@ private_ssh_key: ~/.ssh/openshift
 
 # The patch to store the generated config to access bastion/hosts
 ssh_config_path: /tmp/ssh.config.ansible
+
+openstack_region: regionOne
```
tomassedovic: The users already specify the region in their openstack credentials (alongside the username, tenant, etc.) so we should just read the value from there. Having to enter the same value twice in different places will only lead to confusion.

See: http://docs.ansible.com/ansible/latest/os_client_config_module.html

I've tested the above with `openstack.clouds[0].region_name` and it seems to always work so let's use that instead.

bogdando: I'm not sure what is `openstack.clouds[0].region_name`

bogdando:
@tomassedovic 
> Why are we filtering on the region anyway? Is that really necessary?

https://github.com/openshift/openshift-ansible-contrib/pull/604#issuecomment-318695269

roles/static_inventory/defaults/main.yml
tomassedovic: The os_client_config module sets it up. If you run the playbook below, it will print out the current region name.

And yes, please set the default to that value (note that it can be an empty string, too and that's okay). And actually, I don't think this makes any sense to be configurable in our playbooks so you can remove the `openstack_region` var.

```
---
- hosts: localhost
  become: false
  gather_facts: false
  tasks:
  - os_client_config:
  - debug: var=openstack.clouds[0].region_name
```

tomassedovic: To be clear, using os_client_config will return whatever you have in the `OS_REGION_NAME` environment variable or `region` in your clouds.yaml.

bogdando: Got it. Should I also invoke the os_client_config module someplace?

bogdando:
@tomassedovic updated, removed the region filters, those are not needed / unrelated anyway

tomassedovic:
I've triggered the end to end CI run here:

https://travis-ci.org/openshift/openshift-ansible-contrib/builds/264786207

will look at the code in the meantime

tomassedovic:
CI passed \o/

tomassedovic:
Thanks and sorry for waiting! Looks good.

PR #623, opened 2017-08-07
[WIP] Support external/pre-provisioned authoritative cluster DNS

bogdando:
#### What does this PR do?

* Allow sending updates to multiple (based on the `full_dns_domain` fact) private and public views via 
  mixed external/in-stack DNS servers (i.e.: provisioned for internal use and external for external use).
* Support dynamic updates to multiple clusters views served by multiple DNS servers (see the 
  `external_nsupdate_keys` dict for details). Auto-evaluate used nsupdate keys for the instack DNS 
  server, when using a mixed instack/external DNS servers.
* Also allow to not provision in-stack DNS servers at all and just go with fully external DNS servers.

Fully external DNS server:

* Deploy s DNS server (f.e. by [this guide](https://github.com/openshift/openshift-ansible-contrib/tree/master/reference-architecture/osp-dns)) and populate an inventory file (e.g. `inventory/external_dns.yaml`) with its nsupdate data
* Run provision openstack with app nodes count of 1 and dns nodes count of 0, and add 
  `-e@inventory/external_dns.yaml` for the ansible-playbook commands used
* Finish e2e (it shall pass)
* Run provision openstack with app nodes count of 2 and dns nodes count of 0 and add 
  `-e@inventory/external_dns.yaml` for the ansible-playbook commands used
* Finish e2e (it shall pass)
* Check if all app nodes' FQDNs are resolvable via your external DNS

Mixed instack/external DNS:

* Deploy s DNS server (f.e. by [this guide](https://github.com/openshift/openshift-ansible-contrib/tree/master/reference-architecture/osp-dns)) and populate an inventory file (e.g. `inventory/external_dns.yaml`) with its nsupdate data as a public view, for a private view add this:
```
...
private-openshift.example.com:
  instack: true
  public_access: false # use true if a control node does not belong to the admin (private) cluster network
```
* Run provision openstack with app nodes count of 1 and dns nodes count of 1, and add
  `-e@inventory/external_dns.yaml` for the ansible-playbook commands used
* Finish e2e (it shall pass)
* Run provision openstack with app nodes count of 2 and dns nodes count of 1 and add
  `-e@inventory/external_dns.yaml` for the ansible-playbook commands used
* Finish e2e (it shall pass)
* Check if all app nodes' public FQDNs are resolvable via your external DNS and the private DNS server 
  is used for a private view.

cc: @tomassedovic @oybed PTAL


bogdando:
What works: (pre-provisioned DNS server from a heat stack) it updates its records with unchanged secrets for consequent provision openstack playbook runs.

What doesn't work yet: nslookup for cluster domain does not resolve via the configured DNS server's floating IP (NXDOMAIN). So the example [external_nsupdate_keys](https://github.com/bogdando/ocp-inventory/blob/master/external_dns.yaml) I use for testing is wrong, it seems. Need help with this @tomassedovic @oybed please.

bogdando:
You can also update and use this [example playbook](https://github.com/bogdando/ocp-inventory/blob/master/external_dns_test.yaml) to test the dynamic updates.

playbooks/provisioning/openstack/post-provision-openstack.yml
```diff
@@ -45,6 +45,14 @@
   become: False
   roles:
     - dns-server-detect
+  tasks:
```
oybed: This is already handled by the `dns-server-detect` and if it isn't, the `dns-server-detect` role needs to be enhanced as this is the purpose of that role.

playbooks/provisioning/openstack/post-provision-openstack.yml
```diff
@@ -90,3 +99,18 @@
         line: "IP4_NAMESERVERS={{ hostvars['localhost'].private_dns_server }}"
   roles:
     - node-network-manager
+
+- name: Show configured DNS nsupdate keys for future use as an external one
```
oybed: I like the idea of showing the keys, but unfortunately I think this message may just be "lost" in the noise. My suggestion would be to just have another role that goes with the `dns-server` role (in `infra-ansible` ) that allows us to obtain/gather the keys and display them.

playbooks/provisioning/openstack/post-provision-openstack.yml
bogdando: the dns-server-detect role always sets the `private_dns_server` to `external_nsupdate_keys.private.server`, while for this case we want it to take the 1st dns server private IP (from the Heat stack)

playbooks/provisioning/openstack/post-provision-openstack.yml
bogdando: The end of the post-provision playbook is the current separator between the provision and deployment, it should be no lost in the noise hopefully. I'll explore the option with a separate role in infra-ansible tho. Thanks!

playbooks/provisioning/openstack/post-provision-openstack.yml
bogdando: done

bogdando:
When I scale the app nodes 1->2, cluster nodes can resolve the new node FQDN after the nsupdates done via the external server. But the new app nodes, even though has the mirror configuration to the app-node-0, can't resolve cluster nodes names. I have to debug this.

oybed:
@bogdando Ahh, yes, you have hit a bit of a bug in the dns implementation in `infra-ansible`. Basically the ACLs needs to be updated on the bind server, but this doesn't happen with just adding a new private record through the use of the `dns` role (and nsupdate underneath). The solution may be to open up the ACLs a bit more as right now they are entered as `/32` entries. If we knew the subnet size, we could just add the entire subnet as a private portion of the setup. I'll take a look to see if we can come up with a solution to automatically detect the subnet size and pass that to the `dns-server` role.

bogdando:
@oybed thanks! we have the openshift cluster subnet defined in the inventory vars. But yeah, we need a generic solution for the dns modules

roles/dns-server-detect/tasks/main.yml
```diff
@@ -28,8 +29,9 @@
     public_dns_server: "{{ external_nsupdate_keys['public']['server'] }}"
   when:
     - external_nsupdate_keys['public'] is defined
+    - hostvars[groups['dns'][0]] is undefined
```
oybed: I believe this will cause breakage when we want to use a (internally) provisioned DNS server for internal records + an external for external records.

oybed:
@bogdando Looked into the missing ACL a bit more; the code should be idempotent, so you should be able to add new instances and just re-run the dns-server role to get the ACL updated - i.e.: these lines:
https://github.com/openshift/openshift-ansible-contrib/blob/master/playbooks/provisioning/openstack/post-provision-openstack.yml#L56-L57

I've also submitted an enhancement PR against the dns-server to ensure that any minor changes will restart the server to ensure the configuration is running correctly:
https://github.com/redhat-cop/infra-ansible/pull/63

roles/dns-server-detect/tasks/main.yml
bogdando: yeah, I suggested that the dns group override external nsupdate keys' server entries. This has to be done if we want to support pre-provisioned by a Heat stack DNS servers for future runs. I look forward for better solutions for that use case...

bogdando:
@oybed I hoped to not invoke the dns-server role w/o need, like [this](https://github.com/openshift/openshift-ansible-contrib/pull/623/files#diff-e8bd699c2d8a77d744c68be9bc8da7ceR58). It takes time, and is not idempotent IIRC, as it regens the secrets on each run

playbooks/provisioning/openstack/post-provision-openstack.yml
bogdando: Note, [this change](https://github.com/redhat-cop/infra-ansible/pull/63/files#diff-7e608771f779ebd52f40d3abf6fb8157R3) doesn't cover the specific case I encoded here. We need a tricky logic for putting the public_v4 as the private server entry, can't be used if dumped just as is...

oybed:
@bogdando  [this](https://github.com/openshift/openshift-ansible-contrib/pull/623/files#diff-e8bd699c2d8a77d744c68be9bc8da7ceR58) won't work when using both a provisioned and external DNS (i.e.: provisioned for internal use and external for external use). Also the role is idempotent - it only generates secrets if not found (and to force a new key you just have to delete the proper key before a re-run). .

playbooks/provisioning/openstack/post-provision-openstack.yml
```diff
@@ -55,6 +55,7 @@
   roles:
     - role: dns-views
     - role: infra-ansible/roles/dns-server
+      #when: not external_nsupdate_keys is defined
```
bogdando: this will go away once idempotency of the role verified (it shall not regen keys for consequent runs)

playbooks/provisioning/openstack/post-provision-openstack.yml
bogdando: done, moved to the dns server detect role, @oybed

bogdando:
So the `private/public_dns_server` is auto-detected OK, but duplicating `nsupdate_server_private/public` logic spoils this up. I think the latter must be just removed. @tomassedovic @oybed perchance, do you know how come we have those duplicating vars?

bogdando:
With the https://github.com/openshift/openshift-ansible-contrib/pull/623/commits/c24d4f65af16b2728e20e80a380496c64d7cd9cf it works for me. We'll have a call with @oybed to discuss the details as it is not clear why these complex changes are needed and what is already works out of box/what does not.

bogdando:
Yeah, so now it works. The only thing is that debug message does not expand the updated fact, it keeps showing the original inventory value I passed in:
```
    "external_nsupdate_keys": {
        "private-openshift.cluster.lc": {
            "instack": true, 
            "public_access": true
        }, 
        "public-openshift.cluster.lc": {
            "instack": true
        }
    }
```
While in fact, dynamic updates are sent with the autocompleted nsupdate data for my pre-provisioned instack DNS server:
```
TASK [dns-server-detect : set_fact] **********************************************************************************************************************************************
Tuesday 22 August 2017  18:49:17 +0200 (0:00:00.047)       0:02:22.367 ******** 
ok: [localhost] => {"ansible_facts": {"external_nsupdate_keys": {"private": {"instack": true, "public_access": true}, "public": {"instack": true, "key_algorithm": "HMAC-SHA256", "key_secret": "iiVksjYRCiIeazZudkKOlzgmyhUjmIzIic/67Uk6Xbs=", "server": "38.145.34.165"}}}, "changed": false}

TASK [dns-server-detect : set_fact] **********************************************************************************************************************************************
Tuesday 22 August 2017  18:49:17 +0200 (0:00:00.048)       0:02:22.415 ******** 
ok: [localhost] => {"ansible_facts": {"external_nsupdate_keys": {"private": {"instack": true, "key_algorithm": "HMAC-SHA256", "key_secret": "mm7iy41AN054T3FXqgpOGHFE7qZRWEMzR8Pq3Efp5vg=", "public_access": true, "server": "38.145.34.165"}, "public": {"instack": true}}}, "changed": false}
```
which is odd, but not a fatal thing to fix.

bogdando:
@oybed @tomassedovic the 'allow custom public key names' commit is needed for OSP 10 ref arch compatibility. Here is an example output for the mixed instack/OSP10 external DNS case: https://pastebin.com/QpMrm4EC .  And an example [inventory vars](https://github.com/bogdando/ocp-inventory/commit/2434c68da3a1e8592ee161ab33ccd75d46b8e45d) used for that deploy.

So this works now \o/
I updated the how-to test scenarios in the caption. So try this out please.

bogdando:
Only dynamic updates work as expected. And e2e fails as the cluster hosts get a public IP for the private nameserver. I use `public_access: true` for my external control node, so the `private_dns_nameserver` gets the public IP for sending updates to the private view. This is expected, but doesn't fit internal use for cluster hosts. They need a private IP.

bogdando:
e2e passes for me

bogdando:
I'll revise the changes once again

bogdando:
reworked as https://github.com/openshift/openshift-ansible-contrib/pull/690

PR #632, opened 2017-08-11
Minor fixes to osp-cli scripts

e-minguez:
#### What does this PR do?
Minor fixes

cc: @rlopez133 @cooktheryan PTAL

### DIFF
--- a/reference-architecture/osp-cli/ch5.8.5_configure_docker_storage_ansible.sh
+++ b/reference-architecture/osp-cli/ch5.8.5_configure_docker_storage_ansible.sh
@@ -3,6 +3,6 @@ cat <<EOF >docker-storage-setup
 DEVS=/dev/vdb
 VG=docker-vg
 EOF
-ansible nodes -i inventory -m copy -a "src=docker-storage-setup dest=/etc/sysconfig/docker-storage-setup force=yes"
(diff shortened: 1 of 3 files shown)


reference-architecture/osp-cli/ch5.8.5_configure_docker_storage_ansible.sh
```diff
@@ -3,6 +3,6 @@ cat <<EOF >docker-storage-setup
 DEVS=/dev/vdb
 VG=docker-vg
 EOF
-ansible nodes -i inventory -m copy -a "src=docker-storage-setup dest=/etc/sysconfig/docker-storage-setup force=yes"
+ansible app-nodes,infra-nodes -i inventory -m copy -a "src=docker-storage-setup dest=/etc/sysconfig/docker-storage-setup force=yes"
```
e-minguez: As masters are also nodes but they don't have a dedicated disk for docker images, the docker-storage-setup file is only copied to the hosts they have the dedicated disk.

reference-architecture/osp-cli/ch5.8.5_configure_docker_storage_ansible.sh
```diff
@@ -3,6 +3,6 @@ cat <<EOF >docker-storage-setup
...
 ansible nodes -i inventory -a "/usr/bin/docker-storage-setup"
```
e-minguez: Masters will end up with a loopback file for docker images but as they are not going to run containers, it's ok.

PR #637, opened 2017-08-14
Specify different image names for roles

Tlacenka:
#### What does this PR do?
User can now use different images for different roles (servers).

1) Print out available Glance images ( `openstack image list`).
2) In `<path to inventory>/group_vars/all.yml`, uncomment line(s) with variables for specifying image names (`openstack_<role name>_image_name`) and set the value to one of the other available images.
3) Run the provisioning.
4) Display servers with `openstack server list` and notice that specified roles have changed image names to those set in `all.yml`. The rest of the servers use the default image (defined in `openstack_default_image_name`).

--


tomassedovic:
I've resolved the conflict with the merged multiple networks patch.

tomassedovic:
Triggered an end to end CI run here:

https://travis-ci.org/openshift/openshift-ansible-contrib/builds/264818071

I'll review the code in the meantime.

tomassedovic:
I've tested it with different images locally and the end to end passed. Merging

PR #638, opened 2017-08-14
Specify different flavors for roles

Tlacenka:
#### What does this PR do?
User can now use different flavors for different roles (servers).

1) Print out available Nova flavors (`openstack flavor list`). In `<path to inventory>/group_vars/all.yml`, uncomment line(s) with variables for specifying flavors (`openstack_<role name>_flavor`) and set the value to one of the other available flavors.
2) Run the provisioning.
3) Display servers with `openstack server list` and note ID of both a server with a changed flavor and a server with a default flavor.
4) With `openstack server show <server ID>`, display server's flavor and notice that specified roles have changed image names to those set in `all.yml`. The rest of the servers use the default flavor (defined in `openstack_default_flavor`).

--


tomassedovic:
I've resolved the conflict, running end to end CI here:

https://travis-ci.org/openshift/openshift-ansible-contrib/builds/264831781

I'll review the code in the meantime

tomassedovic:
I've tested this locally and the code looks good, too, but the CI didn't run to completion (it didn't fail either and it's unlikely caused by this patch but I want to re-run it just in case anyway).

So here's end-end number two: https://travis-ci.org/openshift/openshift-ansible-contrib/builds/264846904

PR #639, opened 2017-08-14
Add end to end openstack testing to travis

tomassedovic:
#### What does this PR do?

This is using an RDO Cloud account dedicated to CI.

No manual testing. Just check the Travis CI output. There should be the python 2.7 and 3.5 tox jobs, but in addition, a new end to end job that runs the openstack provisioning, too.

The run takes about 40 minutes on average and should only be executed when we change the code in roles or the playbooks/provisioning section.

To see an example of a successful run, go here:

https://travis-ci.org/tomassedovic/openshift-ansible-contrib/jobs/264409141

n/a

cc:  @bogdando @cooktheryan @celebdor PTAL


tomassedovic:
Huh, it seems like travis is not exporting the creds env vars even though that's working in my fork...

Could be comment above `secure`, I guess? /me tries to remove it.

tomassedovic:
Nope, that wasn't it :-(

ci/openstack/provision.sh
```diff
@@ -0,0 +1,80 @@
...
+mv "$INVENTORY"/ansible.cfg .
+
+sed -i "s/^openstack_ssh_public_key.*/openstack_ssh_public_key: $KEYPAIR_NAME/" "$INVENTORY"/group_vars/all.yml
+sed -i 's/^openstack_external_network_name.*/openstack_external_network_name: "38.145.32.0\/22"/' "$INVENTORY"/group_vars/all.yml
+sed -i 's/^openstack_default_image_name.*/openstack_default_image_name: "CentOS-7-x86_64-GenericCloud-1703"/' "$INVENTORY"/group_vars/all.yml
+sed -i 's/^openstack_num_nodes.*/openstack_num_nodes: 1/' "$INVENTORY"/group_vars/all.yml
+sed -i "s/^env_id.*/env_id: $ENV_ID/" "$INVENTORY"/group_vars/all.yml
+
+PUBLIC_IP="$(curl --silent https://api.ipify.org)"
+echo "node_ingress_cidr: $PUBLIC_IP/32" >> "$INVENTORY"/group_vars/all.yml
```
bogdando: nit: add ssh_ingress_cidr please

ci/openstack/provision.sh
```diff
@@ -0,0 +1,80 @@
...
+echo "manage_packages: False" >> "$INVENTORY"/group_vars/all.yml
+echo "ephemeral_volumes: True" >> "$INVENTORY"/group_vars/all.yml
+
+# TODO(shadower): this just started to break. If this fix works, let's remove
+# it from the sample inventory proper.
+sed -i 's/openshift_release/#openshift_release/' "$INVENTORY"/group_vars/OSEv3.yml
```
bogdando: note, these can be passed the command line -e args as well. Simplifies the code, removes seds :)

tomassedovic: Sigh, of course they can. I'm an idiot. Thanks!

ci/openstack/vars.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
+
+if [ "${RUN_OPENSTACK_CI:-}" == true ]
+then
+    # TODO(shadower): Can we only run this when the project admin asked for it?
```
bogdando: see an example script to fetch data from github API https://github.com/kubernetes-incubator/kubespray/blob/master/scripts/premoderator.sh

tomassedovic:
@cooktheryan @bogdando any idea why the Travis doesn't set the CI env vars here even though they work fine in my fork?

See:
https://travis-ci.org/tomassedovic/openshift-ansible-contrib/jobs/264434544#L426
vs:
https://travis-ci.org/openshift/openshift-ansible-contrib/jobs/264434799#L438

(the fork has OS_USERNAME etc., but they're unset in the travis here)

tomassedovic:
Oh, there are some issues with the secrets not being passed between PRs coming from the local repo vs. the remote ones. So I've pushed the same branch here to see if that helps.

tomassedovic:
nope :-(

tomassedovic:
I'll try to add them to the repository settings.

tomassedovic:
Okay, that did it:

https://travis-ci.org/openshift/openshift-ansible-contrib/jobs/264444380

I'll remove the secret env section from .travis.yml and fix @bogdando's comments shortly.

ci/openstack/vars.sh
tomassedovic: Yeah but I'm not sure how to integrate this here. Travis doesn't let you run builds manually (as far as I could figure out).

The only thing I can think of is: the first build would always fail (because there are no comments), but then someone could say "ci check this" and manually restart the travis build.

That seems cumbersome to me (but maybe is worth it?)

tomassedovic:
It passes! \o/. See: https://travis-ci.org/openshift/openshift-ansible-contrib/builds/264640870

I think we can merge this (ping @cooktheryan -- wanna make sure you're okay with this).

But there's a caveat: the creds are never exposed to pull requests that aren't coming from this repo. So the CI is unavailable in the normal fork/push co my fork/create PR workflow. This is a security feature of Travis and afaik, there's no workaround. See the CI run for the same branch coming from a fork:

https://travis-ci.org/openshift/openshift-ansible-contrib/builds/264640839

So to actually run the openstack CI on a pull request, we have to push the branch to this repo, which is a bit of a pain. We can live with that for now, but it seems to me that we really should just move this to aos-ci jenkins. The good news is, we now have all the scripts we need are in this PR.

tomassedovic:
Oh I've just realised: travis will run the tests again when we merge a change into master. Should we maybe skip the end-end CI when on master?

Also, more reason for us to move this under the jenkins instance eventually.

ci/openstack/vars.sh
bogdando: Unless we have a better solution...

bogdando:
> So to actually run the openstack CI on a pull request, we have to push the branch to this repo, which is a bit of a pain. We can live with that for now.

+1, that works. It's better than nothing. Later we could rework this.

tomassedovic:
Alright, we should be good now. Both travis runs should succeed.

The travis-ci/pr will not have the openstack creds, but that should not fail it -- it will just skip the end to end run.

The travis-ci/push job should do a full end-end run.

tomassedovic:
Okay, they both passed -- I think we're good to go!

@cooktheryan what do you think?

ci/openstack/install.sh
```diff
@@ -0,0 +1,25 @@
...
+fi
+
+if [ "${CI_OVER_CAPACITY:-}" == "true" ]; then
+    echo the CI is over capacity, skipping the end-end test.
+    exit 1
+fi
+
+git clone https://github.com/openshift/openshift-ansible ../openshift-ansible
+cd ../openshift-ansible
+git checkout openshift-ansible-3.6.153-1
```
bogdando: nit: the openshift-ansible version needs to be parametrized

ci/openstack/install.sh
```diff
@@ -0,0 +1,25 @@
...
+cd ../openshift-ansible-contrib
+
+pip install ansible shade dnspython python-openstackclient python-heatclient
+
+curl -L -o oc.tgz https://github.com/openshift/origin/releases/download/v1.5.1/openshift-origin-client-tools-v1.5.1-7b451fc-linux-64bit.tar.gz
```
bogdando: ditto

ci/openstack/provision.sh
```diff
@@ -0,0 +1,89 @@
...
+ephemeral_volumes: True
+EOF
+
+
+# TODO(shadower): this just started to break. We will need to unset this from
+# the sample inventory.
+sed -i 's/openshift_release/#openshift_release/' "$INVENTORY"/group_vars/OSEv3.yml
+
+cat << EOF >> "$INVENTORY"/group_vars/OSEv3.yml
+openshift_master_identity_providers:
```
bogdando: these may be as well specified as extra-vars, the latter takes precedence over the group vars

ci/openstack/install.sh
tomassedovic: Parametrised by what? An environment variable in travis?

For the record, I just wanted to make sure we always use the same version of openshift-ansible so we're not randomly broken by it.

But thinking about it more, maybe we actually always want to use the latest version so we do get notified about breakage and have a chance to file issues/fixes.

bogdando: Yeah, I was thinking of an env var for travis to get the needed version , or use the latest by default

PR #642, opened 2017-08-15
Upgrade templates to support OCP 3.6

glennswest:
#### What does this PR do?
This PR moves the version of OCP to 3.6. This includes changes to the boot volume size to pass
minimum size checks built into 3.6, pre-installs Docker, enables and starts it. 

Both the azuredeploy.json and the allinone.json can be instanced. Note that the resources for the 
azuredeploy.json are significant, and small vm's will not pass the 3.6 mandatory memory or core sizes.

None

cc: @cooktheryan @e-minguez  PTAL


reference-architecture/azure-ansible/azuredeploy.json
```diff
@@ -286,23 +286,23 @@
...
+      "defaultValue" : true,
+      "metadata" : {
         "description" : "Enable OCP metrics"
       }
     },
     "logging" : {
-      "type": "bool",
-      "defaultValue": true,
-      "metadata" :{
+      "type" : "bool",
```
cooktheryan: can you check the spaces here vs the previous code?

It looks like a lot of the "changes" aren't really changes

reference-architecture/azure-ansible/infranode.json
```diff
@@ -118,7 +118,8 @@
               "uri" : "[concat('http://', parameters('sa'), '.blob.core.windows.net/openshiftvmachines/', parameters('vmName'), '.vhd')]"
             },
             "caching" : "ReadWrite",
-            "createOption" : "FromImage"
+            "createOption" : "FromImage",
+            "diskSizeGB" : "128"
```
cooktheryan: Do we have any control on making the disk size smaller ? 128 is rather large

reference-architecture/azure-ansible/master.json
```diff
@@ -130,7 +130,8 @@
               "uri" : "[concat('http://', parameters('sa'), '.blob.core.windows.net/openshiftvmachines/', parameters('vmName'), '.vhd')]"
             },
             "caching" : "ReadWrite",
-            "createOption" : "FromImage"
+            "createOption" : "FromImage",
+            "diskSizeGB" : "128"
```
cooktheryan: same as above

reference-architecture/azure-ansible/azuredeploy.json
glennswest: Actually alot of these are triggered by comma changes and spaces.
The comma changes are mandatory, the spaces changes are caused by a tool chain
difference. I will suggest to edu to use the same tool for editing, and we can avoid 
some of the white space. But things its also lack of smarts on git for json format, as its
not a material change.

reference-architecture/azure-ansible/infranode.json
glennswest: There multiple partitions here, and now a check for 60 Gig on root free. 128 is kind of the good place to be, as depending on storage type, your charged for 128 anyway.. Also it avoids downstream issues later.

PR #643, opened 2017-08-15
Set custom hostnames for servers

Tlacenka:
#### What does this PR do?
User can now set custom hostname prefix for roles (servers).

1) Remove `hosts` file in your inventory directory if it still exists.
2) In `<path to inventory>/group_vars/al.yml`, uncomment line(s) with variable for setting custom hostnames (`openstack_<role>_hostname`) and set their value to something else to your liking (but which is also a valid hostname).
3) Run the provisioning.
4) Display your servers with `openstack server list`. Note that servers with customised hostnames have prefixes according to variables in your inventory.

--


tomassedovic:
I've triggered an end to end CI run here:

https://travis-ci.org/openshift/openshift-ansible-contrib/builds/264859762

will do a review in the meantime

tomassedovic:
Oops, I fumbled that one, let's try this again:

https://travis-ci.org/openshift/openshift-ansible-contrib/builds/264864436

tomassedovic:
Ok, the end-end ci passed, but I didn't get to do the review yet. Will do it tomorrow.

tomassedovic:
I agree. Just to be super safe, I'm running a full end to end test with all the hostnames modified to make sure there is no subtle breakage with that. Once that's done, we can merge this.

PR #644, opened 2017-08-15
Configure different Docker volume sizes for different roles

Tlacenka:
#### What does this PR do?
User can now specify different Docker volume size for the servers.
Additionally
- I changed `app_volume_size` to `node_volume_size` as `node` is used to describe an application node in multiple places.
- I added `lb_volume_size` instead of a hard-coded value in `heat_stack.yaml.j2`.

1) In `<path to inventory>/group_vars/all.yml`, uncomment line(s) with variables for specifying Docker volume size (`docker_<role>_volume_size`) and change its value to a valid volume size.
2) Run the provisioning.
3) Display all servers with `openstack server list`. Choose one server with a default volume size and one with a changed size, not their ID.
4) Display server's volume size by running `openstack server show <server ID>`, note the volume ID (last line) and then run `openstack volume show <volume ID>`. Check that the right values were used.

--


playbooks/provisioning/openstack/README.md
```diff
@@ -120,6 +120,13 @@ cluster node groups, like app or infra nodes. For example: `{'region': 'infra'}`
 The `openstack_nodes_to_remove` allows you to specify the numerical indexes
 of App nodes that should be removed; for example, ['0', '2'],
 
+The `docker_volume_size` is the default Docker volume size the servers will use.
```
bogdando: please describe how these play with the ephemeral volumes variable

tomassedovic:
end to end CI: https://travis-ci.org/openshift/openshift-ansible-contrib/builds/265209739

(I was kind of hoping Travis would add the checks here, but oh well :-( )

Tlacenka:
@tomassedovic I looked at the CI and it said it failed...but when I then I checked the log...it didn't show any fails or unreachables in the play recap...what happened?

tomassedovic:
@Tlacenka yep, I'm looking at it now. It seems the CI's unable to find the `oc` binary (which was literally never a problem before).

I'll try to fix this and re-run this.

tomassedovic:
Let's try this again:

https://travis-ci.org/openshift/openshift-ansible-contrib/builds/265499471

tomassedovic:
Okay this last test with my fixes passed: https://travis-ci.org/openshift/openshift-ansible-contrib/builds/265533486

We're good to merge this. Sorry and thanks for your patience!

PR #646, opened 2017-08-15
Migrate bin/cluster and associated playbooks from openshift-ansible

sdodson:
These playbooks were never officially supported and need to be removed from openshift-ansible. I'm not certain that they should live in this repo either as I imagine they conflict in functionality to some extent with code that already lives in this repo.

./bin/cluster --help


abutcher:
These playbooks reference relative common playbooks from openshift-ansible which will no longer be available.

sdodson:
Probably need to update them to reference them from the installed path like the example you pointed me at yesterday.

PR #647, opened 2017-08-16
Update openshift_release in the sample inventory

tomassedovic:
#### What does this PR do?
Openshift Enterprise is now at 3.6. And we should not specify the
origin release as it's more rolling and the older versions may no longer
be available (this has broken our CI recently).

Do an end to end deployment of origin and enterprise using the sample inventory. Nothing should break.

n/a

cc: @Tlacenka  @bogdando @tzumainn  PTAL


tomassedovic:
@bogdando sure, will do. As far as I can tell, it uses the latest release of origin (which used to be 1.5.1 but is 3.6.0). But it looks like the configured origin repos only provide the latest version, so when origin makes a new release, the old one is not available anymore.

tomassedovic:
done

tomassedovic:
Please ignore the failed push check -- I accidentally pushed a branch into this repo which triggered a check that I then cancelled.

Tlacenka:
@tomassedovic I noticed that in the provisioning [README](https://github.com/openshift/openshift-ansible-contrib/blob/master/playbooks/provisioning/openstack/README.md), it still requires origin release. I think it would be nice to update the documentation of `OSEv3.yml`.

tomassedovic:
Good point, thanks!

Tlacenka:
@tomassedovic 
I might be missing something but when trying to do the end to end development, it fails in the OpenShift installation part. I am using `rhel73` image for the nodes.

```
Message:  You requested openshift_release 3.6, which is not matched by
               the latest OpenShift RPM we detected as atomic-openshift-3.5.5.31
               on host master-0.openshift.example.com.
               We will only install the latest RPMs, so please ensure you are getting the release
               you expect. You may need to adjust your Ansible inventory, modify the repositories
               available on the host, or run the appropriate OpenShift upgrade playbook.
```

PR #648, opened 2017-08-16
Disable the openshift_version default group var

bogdando:
#### What does this PR do?
The given default value no longer works with the error msg:
"You requested openshift_release 1.5.1, which is not matched by the latest OpenShift RPM we detected as origin-3.6.0..."

e2e

cc: @tomassedovic PTAL


bogdando:
overlaps to https://github.com/openshift/openshift-ansible-contrib/pull/647

bogdando:
overlaps to #647

PR #679, opened 2017-08-22
Fix node label customisation

### LINKED ISSUE #674: Provision fails when openshift_cluster_node_labels does not define at least one label for each group
## Bug Report

When getting values of `openshift_cluster_node_labels` in [heat_stack.yaml.j2](https://github.com/openshift/openshift-ansible-contrib/blob/master/roles/openstack-stack/templates/heat_stack.yaml.j2#L688) during provisioning, if `openshift_cluster_node_labels.infra` (or `.app`) is not defined, provisioning ends with an error (that `iteritems()` cannot be called for an undefined object).

This happens when someone overrides the  default `openshift_cluster_node_labels` value defined [here](https://github.com/openshift/openshift-ansible-contrib/blob/master/roles/common/defaults/main.yml) with their own values but does not define at least one label for both `app` and `infra` group.

Tlacenka:
#### What does this PR do?
This PR fixes problems with label customisation.

- In order for the heat templates to work, it adds an empty dictionary to both infra and app groups if there isn't any label defined for them.
- The `openshift_cluster_node_labels` definition is moved to `all.yml` due to its usage in the `openstack-stack` role during provisioning.
- Documentation for customising node labels is expanded.

1) In `all.yml`, add custom definition to `openshift_cluster_node_labels` which contains only one of the available groups (app or infra).
2) Run provisioning with `-vvv` option and take a look at the output of the tasks right before the `openstack-stack` role (in particular, `Set openshift_cluster_node_labels for the app group if not defined`). It should show the content of `openshift_cluster_node_labels` variable that has both `app` and `infra` keys in it (along with your custom label).
3) Check that the end to end deployment is successful.
4) Connect to the OpenShift master node and display all nodes, including their labels with `oc get nodes --show-labels`. Your custom labels should be printed out along with the predefined labels.

Closes #674 


Tlacenka:
yey, it passed, and so did the build @bogdando started ( https://travis-ci.org/openshift/openshift-ansible-contrib/builds/267579894 )

playbooks/provisioning/openstack/pre_tasks.yml
```diff
@@ -31,3 +31,19 @@
...
   when:
   - openshift_master_default_subdomain is undefined
+
+# Check that openshift_cluster_node_labels has regions defined for all groups
+# NOTE(kpilatov): if node labels are to be enabled for more groups,
+#                 this check needs to be modified as well
+- name: Set openshift_cluster_node_labels if undefined (should not happen)
+  set_fact:
+    openshift_cluster_node_labels: {'app': {'region': 'primary'}, 'infra': {'region': 'infra'}}
+  when: openshift_cluster_node_labels is not defined
```
bogdando: An elegant solution to catch up changes for the all.yml for folks who're used to not update it. I like it!

tomassedovic:
@bogdando did you test the features here? I was under the impression that @Tlacenka wanted to finish her local tests before merging this.

tomassedovic:
@bogdando oh my bad, I think that was the scaling up PR. This is fine.

bogdando:
yeah, that one passed CI

PR #681, opened 2017-08-22
Use overlay2 in GCP ref. arch.

pschiffe:
#### What does this PR do?
Use overlay2 in GCP ref. arch. Pre-install docker in gold image. Enable overwriting of DNS records, so some redeploy use-cases are faster.

Verify that OCP deploys on GCP correctly, and docker uses overlay storage on separate disk.

cc: @cooktheryan  PTAL

### DIFF
--- a/reference-architecture/gcp/deployment-manager/core.jinja
+++ b/reference-architecture/gcp/deployment-manager/core.jinja
@@ -72,12 +72,21 @@ resources:
             #cloud-config
             write_files:
             - content: |
-                DEVS='/dev/disk/by-id/google-docker'
-                VG='docker-vol'
-                DATA_SIZE='95%VG'
-                EXTRA_DOCKER_STORAGE_OPTIONS='--storage-opt dm.basesize=3G'
-              path: /etc/sysconfig/docker-storage-setup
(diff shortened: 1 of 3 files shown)


reference-architecture/gcp/deployment-manager/core.jinja
```diff
@@ -72,12 +72,21 @@ resources:
             #cloud-config
             write_files:
             - content: |
-                DEVS='/dev/disk/by-id/google-docker'
-                VG='docker-vol'
-                DATA_SIZE='95%VG'
-                EXTRA_DOCKER_STORAGE_OPTIONS='--storage-opt dm.basesize=3G'
-              path: /etc/sysconfig/docker-storage-setup
+                DOCKER_STORAGE_OPTIONS='--storage-driver overlay2'
```
cooktheryan: Are you providing a device or due to the health checks are you just increasing /var ?

pschiffe:
I always provided separate disk for docker in GCP, so there's no
change. Just docker-storage-setup cannot create filesystem and mount
it, so I'm skipping it and configuring the docker directly.

peter

Issue #686, opened 2017-08-23
Openstack provisioning broken with latest openshift-ansible


tomassedovic:
As shown in this end to end CI run:

https://travis-ci.org/openshift/openshift-ansible-contrib/jobs/267210691

This is broken with the current openshift-ansible master (commit b107f677a3563f6db9a1ef17fe10518d321d495c).

I've verified that this commit does not have the issue: 1febf0b20de622d358dc890417dc4cda7c3f71a9

# BUG REQUEST INFO:

**Environment**:
- **Cloud provider:** RDO Newton (openstack)

- **OS (`printf "$(uname -srm)\n$(cat /etc/os-release)\n"`):** centos7

- **Version of Ansible:** (`ansible --version`): ansible2.3

- **Version of Jinja:** (`pip freeze | grep -i jinja`) n/a

- **Version of Shade:** (`pip freeze | grep -i shade`) n/a

**Openstack-ansible-contrib version (commit) (`git rev-parse --short HEAD`):** 027210b

tomassedovic:
I have as strong suspicion this is the culprit:

https://github.com/openshift/openshift-ansible/commit/00afac6ef3235ab66e0a0c02cc863cb577e127f2

tomassedovic:
Yep, I've just tested it locally, d7d97964f5e3b5d2df9985e7f9d7b74a974458cb is the commit that broke us.

PR #687, opened 2017-08-23
Pin openshift ansible in ci

tomassedovic:
#### What does this PR do?
This pins a specific openshift-ansible version in the openstack end to end CI that sholudn't be broken for us.

The end to end ci should pass. No manual testing necessary.

Issue #686 

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ python:
 env:
   global:
     - CI_CONCURRENT_JOBS=1
-    - OPENSHIFT_ANSIBLE_COMMIT=master
+    - OPENSHIFT_ANSIBLE_COMMIT=1febf0b20de622d358dc890417dc4cda7c3f71a9
   matrix:
     - RUN_OPENSTACK_CI=false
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -15,7 +15,7 @@ python:
 env:
   global:
     - CI_CONCURRENT_JOBS=1
-    - OPENSHIFT_ANSIBLE_COMMIT=master
+    - OPENSHIFT_ANSIBLE_COMMIT=1febf0b20de622d358dc890417dc4cda7c3f71a9
```
bogdando: please use a tag?

tomassedovic: The CI's running now so I'll let it run its course. We'd have to re-run it with the tag afterwards (to make sure that fixes it) so if you don't mind, I'd just leave this one be and I'll make sure to use tags here in the future.

tomassedovic:
@bogdando the CI passed, can we merge this to unblock everyone or do you want me to switch to using tags?

Tlacenka:
LGTM :)

PR #688, opened 2017-08-23
Added checks for configured images and flavors

Tlacenka:
#### What does this PR do?
Now that users can specify images and flavors for each role, we need a check that confirms that all configured images and flavors are available. Which is exactly what this PR does.

**Invalid setting**:
1) In `all.yml`, try setting image/flavor for some roles to something invalid.
2) Run provisioning and check that it fails with `Image <your image name> is not available.`

**Valid setting**:
1) Run provisioning with valid images and flavors and check that it is successful.

--


tomassedovic:
End to End CI running here:

https://travis-ci.org/openshift/openshift-ansible-contrib/builds/268310791

PR #690, opened 2017-08-24
Support external/pre-provisioned authoritative cluster DNS

bogdando:
#### What does this PR do?
* Document how to use fully external DNS servers w/o provisioning
  dns servers group with Heat.

* Document how to use a mixed servers setup for dynamic records
  updates mathing public or private views.

* Allow custom public key names for OSP10 dns service compatibility.
  The osp-dns configures the named service with the fixed key_name
  'update-key'. Add optional key_name for the external_nsupdate_keys
  public section to allow custom key names.

Fully external DNS server:

* Deploy s DNS server (f.e. by [this guide](https://github.com/openshift/openshift-ansible-contrib/tree/master/reference-architecture/osp-dns)) and populate an inventory file (e.g. `inventory/external_dns.yaml`) with its nsupdate data
* Run provision openstack with app nodes count of 1 and dns nodes count of 0, and add 
  `-e@inventory/external_dns.yaml` for the ansible-playbook commands used
* Finish e2e (it shall pass)
* Run provision openstack with app nodes count of 2 and dns nodes count of 0 and add 
  `-e@inventory/external_dns.yaml` for the ansible-playbook commands used
* Finish e2e (it shall pass)
* Check if all app nodes' FQDNs are resolvable via your external DNS

Mixed instack/external DNS:

* Deploy s DNS server (f.e. by [this guide](https://github.com/openshift/openshift-ansible-contrib/tree/master/reference-architecture/osp-dns)) and populate an inventory file (e.g. 
  `inventory/external_dns.yaml`) `external_nsupdate_keys.public` with the required server, key,
  algorithm data.
* Run provision openstack with app nodes count of 1 and dns nodes count of 1, and add
  `-e@inventory/external_dns.yaml` for the ansible-playbook commands used
* Finish e2e (it shall pass)
* Run provision openstack with app nodes count of 2 and dns nodes count of 1 and add
  `-e@inventory/external_dns.yaml` for the ansible-playbook commands used
* Finish e2e (it shall pass)
* Check if:
  * app nodes' public FQDNs are resolvable via your external DNS (run nslookup or dig against the public 
     DNS server public IP)
  * the private DNS server is used for a private view (run nslookup or dig from a cluster node to verify 
     this).

cc: @tomassedovic @oybed PTAL

### DIFF
--- a/roles/dns-records/tasks/main.yml
+++ b/roles/dns-records/tasks/main.yml
@@ -14,6 +14,7 @@
     nsupdate_server_private: "{{ external_nsupdate_keys['private']['server'] }}"
     nsupdate_key_secret_private: "{{ external_nsupdate_keys['private']['key_secret'] }}"
     nsupdate_key_algorithm_private: "{{ external_nsupdate_keys['private']['key_algorithm'] }}"
+    nsupdate_private_key_name: "{{ external_nsupdate_keys['private']['key_name']|default('private-' + full_dns_domain) }}"
   when:
     - external_nsupdate_keys is defined
     - external_nsupdate_keys['private'] is defined
@@ -32,7 +33,7 @@
       - view: "private"
         zone: "{{ full_dns_domain }}"
         server: "{{ nsupdate_server_private }}"
-        key_name: "{{ ( 'private-' + full_dns_domain ) }}"
+        key_name: "{{ nsupdate_private_key_name|default('private-' + full_dns_domain) }}"
         key_secret: "{{ nsupdate_key_secret_private }}"
         key_algorithm: "{{ nsupdate_key_algorithm_private | lower }}"
         entries: "{{ private_records }}"
@@ -54,6 +55,7 @@
     nsupdate_server_public: "{{ external_nsupdate_keys['public']['server'] }}"
     nsupdate_key_secret_public: "{{ external_nsupdate_keys['public']['key_secret'] }}"
     nsupdate_key_algorithm_public: "{{ external_nsupdate_keys['public']['key_algorithm'] }}"
+    nsupdate_public_key_name: "{{ external_nsupdate_keys['public']['key_name']|default('public-' + full_dns_domain) }}"
   when:
     - external_nsupdate_keys is defined
     - external_nsupdate_keys['public'] is defined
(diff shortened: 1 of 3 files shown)


roles/dns-records/tasks/main.yml
```diff
@@ -54,6 +54,7 @@
     nsupdate_server_public: "{{ external_nsupdate_keys['public']['server'] }}"
     nsupdate_key_secret_public: "{{ external_nsupdate_keys['public']['key_secret'] }}"
     nsupdate_key_algorithm_public: "{{ external_nsupdate_keys['public']['key_algorithm'] }}"
+    nsupdate_public_key_name: "{{ external_nsupdate_keys['public']['key_name']|default('public-' + full_dns_domain) }}"
```
oybed: This is a good change, and I think it could be beneficial to have it for the "private" section above as well. Do you mind making a similar change above?

roles/dns-records/tasks/main.yml
```diff
@@ -72,7 +73,7 @@
       - view: "public"
         zone: "{{ full_dns_domain }}"
         server: "{{ nsupdate_server_public }}"
-        key_name: "{{ ( 'public-' + full_dns_domain ) }}"
+        key_name: "{{ nsupdate_public_key_name|default('public-' + full_dns_domain) }}"
```
oybed: same as above

roles/dns-records/tasks/main.yml
bogdando: will do!

tomassedovic:
One question to both of you @bogdando and @oybed while we're thinking about external DNS.

Do you think it would be worthwhile to add support for servers that don't have multiple views? The one I'm testing this against doesn't so I had to create two of them. We would have to add e.g. a hostname suffix to the private entries and configure openshift-ansible accordingly, or something.

FWIW this is what the Red Hat reference architectures do (different names for public/private resolution).

glennswest:
Yes 
The azure has a internal name server 
And servers have internal and external names 

In customer I see also multiple domains as well for applications 

To the point there needs to be a microservice dns to facilitate

roles/dns-records/tasks/main.yml
bogdando: done

bogdando:
@tomassedovic addressed the @oybed comments.

wrt a single view DNS servers and different FQDNs for priv/pub access, I believe we should switch the default DNS stack to Designate as soon as we can do that. I see no need to go (and maintain) offroad solutions but those that OpenStack DNSaaS proposes. Let's please create a separate feature request as well.

tomassedovic:
Yeah sorry if I was unclear. I meant it as a separate RFE/PR. We can merge this one when the end/end CI passes.

oybed:
@tomassedovic supporting an internal and external set of hostnames is possible and it can all be driven by the playbooks. However, as-is, and what I have done in these cases where the DNS server doesn't support views, is to stand up 2 DNS servers (potentially with different ACLs) and hence one will serve the internal aspect and the other the external. With that approach, the playbooks will work as-is without any special handling/changes needed.

tomassedovic:
@oybed yeah that (2 servers) is what I've done so far. And yeah, good point on this being possible through the playbooks alone. I forgot about that.

Issue #621, opened 2017-08-06
Openstack provider creates Open DNS service


ggillies:
**Environment**: Openshift Origin 1.5 on CentOS 7
- **Cloud provider:** Openstack

- **OS (`printf "$(uname -srm)\n$(cat /etc/os-release)\n"`):** Linux 3.10.0-514.10.2.el7.x86_64 x86_64
NAME="CentOS Linux"
VERSION="7 (Core)"
ID="centos"
ID_LIKE="rhel fedora"
VERSION_ID="7"
PRETTY_NAME="CentOS Linux 7 (Core)"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:centos:centos:7"
HOME_URL="https://www.centos.org/"
BUG_REPORT_URL="https://bugs.centos.org/"

CENTOS_MANTISBT_PROJECT="CentOS-7"
CENTOS_MANTISBT_PROJECT_VERSION="7"
REDHAT_SUPPORT_PRODUCT="centos"
REDHAT_SUPPORT_PRODUCT_VERSION="7"

- **Version of Ansible:** (`ansible --version`): ansible 2.3.1.0
  config file = /home/ggillies/code/upshift-scale/ansible.cfg
  configured module search path = [u'library', u'ceph-ansible/library']
  python version = 2.7.13 (default, Jun 26 2017, 10:20:05) [GCC 7.1.1 20170622 (Red Hat 7.1.1-3)]

- **Version of Jinja:** (`pip freeze | grep -i jinja`) Jinja2==2.9.6

- **Version of Shade:** (`pip freeze | grep -i shade`) shade==1.22.2

- **Openstack-ansible-contrib version (commit) (`git rev-parse --short HEAD`):** dc9a301

After successfully using the Openstack provider playbooks to deploy Openshift on Openstack, I noticed that the dns service deployed (dns-0 VM) is deployed as an Open DNS service. My deployment is publicly facing (on the internet) and this is a security vulnerability as people can use Open DNS servers for attacks on other services (as well as attacking my own infrastructure). This DNS service should be restricted to only provide the records it serves.

bogdando:
Please take a look the [security notes](https://github.com/openshift/openshift-ansible-contrib/blob/master/playbooks/provisioning/openstack/README.md#security-notes) . Specifying the `node_ingress_cidr` effectively hides the provisioned DNS servers from a public/open access

tomassedovic:
[...]

PR #694, opened 2017-08-25
Document global DNS security options

bogdando:
#### What does this PR do?
Depends on https://github.com/redhat-cop/infra-ansible/pull/69

* Document global DNS security options, specifically a recursion switches.
* Do not create a view if externally managed.
* Allow to specify the recursion settings for public/private views defined by the dns-view role.

note: dnsses is yet to be supported

Defaults are left as is and need no to be tested.
A secure setup recommended in the guide, should be tested:

* Define inventory vars:
  ```
  named_config_recursion: 'no'
  named_config_views:
    public:
      - recursion: 'no'
      ... (remaining data for the infra-ansible casl dns role) ...
  ```
* Provision a non-recursive public DNS, by [casl dns* roles](https://github.com/redhat-cop/infra-ansible/tree/master/roles/dns-server), using the inventory vars defined above. See this [example playbook](https://github.com/bogdando/ocp-inventory/blob/master/playbooks/deploy-dns-casl.yaml) based on the [osp 10 ref arch DNS service](https://github.com/openshift/openshift-ansible-contrib/tree/master/reference-architecture/osp-dns) 'hacked' to use the infra-ansible casl dns role.
* Provision and deploy a mixed DNS servers setup, having an in-stack private DNS allowing recursion.
* e2e shall pass
* Verify there is no recursion available for the external DNS:
```
dig foo.com @<IP>
WARNING: recursion requested but not available
```
* Repeat the steps with an in-stack DNS server used for priv/pub views, and define `named_public_recursion: 'no'`.

Related https://github.com/openshift/openshift-ansible-contrib/issues/621

cc: @tomassedovic @oybed PTAL

### DIFF
--- a/playbooks/provisioning/openstack/README.md
+++ b/playbooks/provisioning/openstack/README.md
@@ -85,6 +85,8 @@ default hostname (usually the role name) is used.
 The `public_dns_nameservers` is a list of DNS servers accessible from all
 the created Nova servers. These will be serving as your DNS forwarders for
 external FQDNs that do not belong to the cluster's DNS domain and its subdomains.
+If you're unsure what to put in here, you can try the google or opendns servers,
+but note that some organizations may be blocking them.
 
 The `openshift_use_dnsmasq` controls either dnsmasq is deployed or not.
 By default, dnsmasq is deployed and comes as the hosts' /etc/resolv.conf file
--- a/playbooks/provisioning/openstack/README.md
+++ b/playbooks/provisioning/openstack/README.md
@@ -200,6 +200,18 @@ be the case for development environments. When turned off, the servers will
 be provisioned omitting the ``yum update`` command. This brings security
 implications though, and is not recommended for production deployments.
 
+##### DNS servers security options
+
+Aside from `node_ingress_cidr` restricting public access to in-stack DNS
+servers, there are following (bind/named specific) DNS security
+options available:
+
+    named_public_recursion: 'no'
+    named_private_recursion: 'yes'
+
+External DNS servers, which is not included in the 'dns' hosts group,
+are not managed. It is up to you to configure such ones.
+
 ### Configure the OpenShift parameters
 
 Finally, you need to update the DNS entry in
--- a/playbooks/provisioning/openstack/sample-inventory/group_vars/all.yml
+++ b/playbooks/provisioning/openstack/sample-inventory/group_vars/all.yml
@@ -92,6 +92,10 @@ rhsm_register: False
 #    key_algorithm: 'hmac-md5'
 #    server: '192.168.1.2'
 
+# # Customize DNS server security options
+#named_public_recursion: 'no'
+#named_private_recursion: 'yes'
+
 
 # NOTE(shadower): Do not change this value. The Ansible user is currently
 # hardcoded to `openshift`.
--- a/roles/dns-views/defaults/main.yml
+++ b/roles/dns-views/defaults/main.yml
@@ -1,0 +1,4 @@
+---
+external_nsupdate_keys: {}
+named_private_recursion: 'yes'
+named_public_recursion: 'no'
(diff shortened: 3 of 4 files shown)


bogdando:
dnssec options will not take effect yet, the dns-server role needs to be adjusted for required dnssec configuration steps, keys generating signing zones et al. But what is unexpected, I can see the dig tool showing `ra rd` flags and executes recursions against my test server, even though I told it to not recurse.

oybed:
@bogdando keep in mind that the views also have recursion config - did you ensure those are disabled? In general, this needs better documentation, but you need to set:
```
named_config_views:
  - recursion: "no"
```

oybed:
@bogdando as discussed, one enhancement we can make is to make the views default to the global setting rather than just default to enabled. I'll submit another PR to `infra-ansible` for this.

bogdando:
@tomassedovic  it's ready for review and testing

tomassedovic:
This looks great, thanks!

We need to sort out the default values (they're currently not working), but when I set them up explicitly:

```
named_public_recursion: 'no'
named_private_recursion: 'yes'
```

I've encountered two issues.

1. My openstack requires setting explicit forwarders (via `public_dns_nameservers`) and we currently only set those for the public views, not the private ones. That means the nodes could not install packages or do any external name resolutions.

Here's the views file on the deployed DNS:

```
$ ansible --become --private-key ~/.ssh/openshift -i devbox-inventory/ dns -m command -a 'cat /etc/named/named.conf.view'
dns-0.openshift.example.com | SUCCESS | rc=0 >>
view "private" {
        match-clients { "private"; };

        recursion yes;

        zone "openshift.example.com" IN {
                type master;
                file "static/private-openshift.example.com.db";
                allow-update { key private-openshift.example.com; };
        };


        zone "." {
                type forward;
                forwarders {
                        127.0.0.1;
                };
        };

};

view "public" {
        match-clients { "public"; };

        recursion no;

        zone "openshift.example.com" IN {
                type master;
                file "static/public-openshift.example.com.db";
                allow-update { key public-openshift.example.com; };
        };


        include "/etc/named.rfc1912.zones";

        zone "." {
                type forward;
                forwarders {
                        10.44.6.1;
                        10.44.6.2;
                        };
        };

};
```

That should be fixed by adding `forwarder: "{{ public_dns_nameservers }}"` to the private view here just like the public one does in our `dns-views` role:

https://github.com/openshift/openshift-ansible-contrib/blob/a9ca2ec9ec25e752dab2be543b91ddd122826057/roles/dns-views/tasks/main.yml#L9

Paging @oybed to make sure this actually makes sense.


2. When I did the above, the openshift-ansible installation succeeded, but running a build failed, because the build pod couldn't resolve names either.

I am not yet sure why this is and will investigate this further soon.

bogdando:
@tomassedovic I think you're right and we need to

> That should be fixed by adding forwarder: "{{ public_dns_nameservers }}" to the private view

I'll test with the empty `public_dns_nameservers`. which is a default value, as well

bogdando:
@tomassedovic basically, if we leave the default `public_dns_nameservers` empty, adding `forwarder: "{{ public_dns_nameservers }}"` to the private view would fix nothing

tomassedovic:
But there are situations (such as the RDO Cloud) where empty `public_dns_nameservers` work already. These wouldn't be broken if we added an empty forwarder list to the private view would they?

I think the defaults should stay the same, but we should document the forwarders situation better. I'd prefer we didn't use any open DNS as a default, because they may be blocked in certain orgs and it would be difficult to debug that failure.

But we can say something like "if you're unsure what to put in here, you can try the google or opendns servers, but note that some organizations may be blocking them).

bogdando:
@tomassedovic done

tomassedovic:
Looks good, thanks! I'll test this once I've handled the CI issues (hopefully later today).

PR #701, opened 2017-08-28
Allow using a provider network

### LINKED ISSUE #622: Support Openstack Provider using an existing provider network
Support Openstack Provider using an existing provider network

## **Problem Description**

In a typical deployment on Openshift on Openstack, when using the Openshift SDN, you essentially have two layers of encapsulation with Openvswitch (once at the Openshift SDN level, and one at the Openstack network level). I would like the Openstack provider playbooks to support using either a new, or an existing provider network for SDN traffic. This would solve this problem as we could use a provider network of type vlan, bypassing openstacks normal openvswitch encasulation

## **Implementation**

Describe your plan for implementation steps and potential
assignees/contributors.

## **Acceptance criteria**

Provide a definition of done and acceptance criteria.

## **Testing**

Describe how to test this, including manual steps, if any.

tomassedovic:
#### What does this PR do?
This adds a new option `openstack_provider_network_name` which will take
a name of an existing network and put the servers there. It will also
prevent creating floating IP addresses as the provider network's IPs
should already be accessible without any additional routing required.

Fixes #622

You need an openstack with a provider network set up. Alternatively, you can create a new Neutron network, create a server in the network, SSH in and run the deployment from there.

1. Set `openstack_provider_network_name` to the name of your provider network in `inventory/group_vars/all.yml`
2. Comment out the `openstack_external_network_name` and `openstack_private_network_name` options in `all.yml`
3. Run the `provision.yaml` playbook

* Verify that no new Neutron network or subnet were created.
* Verify that all the Nova servers were placed in the existing provider network.
* Verify that none of the servers have a floating IP address assigned

4. Either update the provider subnet's `dns_nameservers` option (by: `openstack subnet set --dns-nameserver`) or edit `/etc/resolv.conf` on each node to make sure the nodes use the provisioned DNS
5. Install openshift with the openshift-ansible playbooks
6.  Run a source-to-image workload and verify that the image gets created, pushed to the registry and the pod is accessible via its route (e.g.: `oc new-app --template=cakephp-mysql-example`)

#622 

cc: @ggillies @rlopez133  @bogdando PTAL

### DIFF
--- a/playbooks/provisioning/openstack/README.md
+++ b/playbooks/provisioning/openstack/README.md
@@ -244,6 +244,26 @@ be the case for development environments. When turned off, the servers will
 be provisioned omitting the ``yum update`` command. This brings security
 implications though, and is not recommended for production deployments.
 
+#### Provider Network
+
+Normally, the playbooks create a new Neutron network and subnet and attach
+floating IP addresses to each node. If you have a provider network set up, this
+is all unnecessary as you can just access servers that are placed in the
+provider network directly.
+
+To use a provider network, set its name in `openstack_provider_network_name` in
+`inventory/group_vars/all.yml`.
+
+You must also unset the `openstack_external_network_name` and
+`openstack_private_network_name` fields.
+
+**NOTE**: this will not update the nodes' DNS, so running openshift-ansible
+right after provisioning will fail (unless you're using an external DNS server
+your provider network knows about). You must make sure your nodes are able to
+resolve each other by name, e.g. by setting `dns_nameservers` on your subnet or
+editing each node's `/etc/resolv.conf`.
+
+
 ### Configure the OpenShift parameters
 
 Finally, you need to update the DNS entry in
--- a/playbooks/provisioning/openstack/README.md
+++ b/playbooks/provisioning/openstack/README.md
@@ -254,15 +254,13 @@ provider network directly.
 To use a provider network, set its name in `openstack_provider_network_name` in
 `inventory/group_vars/all.yml`.
 
-You must also unset the `openstack_external_network_name` and
-`openstack_private_network_name` fields.
+If you set the provider network name, the `openstack_external_network_name` and
+`openstack_private_network_name` fields will be ignored.
 
 **NOTE**: this will not update the nodes' DNS, so running openshift-ansible
 right after provisioning will fail (unless you're using an external DNS server
 your provider network knows about). You must make sure your nodes are able to
-resolve each other by name, e.g. by setting `dns_nameservers` on your subnet or
-editing each node's `/etc/resolv.conf`.
-
+resolve each other by name.
 
 ### Configure the OpenShift parameters
 
--- a/playbooks/provisioning/openstack/prerequisites.yml
+++ b/playbooks/provisioning/openstack/prerequisites.yml
@@ -87,20 +87,6 @@
       that: 'key_result.rc == 0'
       msg: "Keypair {{ openstack_ssh_public_key }} is not available"
 
-  - name: The provider and private networks can't be set at the same time
-    fail:
-      msg: You cannot set both `openstack_provider_network_name` and `openstack_private_network_name` at the same time
-    when:
-    - openstack_provider_network_name|default(None)
-    - openstack_private_network_name|default(None)
-
-  - name: The provider and external networks can't be set at the same time
-    fail:
-      msg: You cannot set both `openstack_provider_network_name` and `openstack_external_network_name` at the same time
-    when:
-    - openstack_provider_network_name|default(None)
-    - openstack_external_network_name|default(None)
-
 # Check that custom images and flavors exist
 - hosts: localhost
 
--- a/playbooks/provisioning/openstack/prerequisites.yml
+++ b/playbooks/provisioning/openstack/prerequisites.yml
@@ -65,10 +65,12 @@
     os_networks_facts:
       name: "{{ openstack_external_network_name }}"
     register: network_result
+    when: not openstack_provider_network_name|default(None)
   - name: Check that network is available
     assert:
       that: "network_result.ansible_facts.openstack_networks"
       msg: "Network {{ openstack_external_network_name }} is not available"
+    when: not openstack_provider_network_name|default(None)
 
   # Check keypair
   # TODO kpilatov: there is no Ansible module for getting OS keypairs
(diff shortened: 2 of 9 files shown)


tomassedovic:
I'm not quite sure what to do with DNS here. Normally, we update the Neutron subnet we create to use the appropriate DNS server (i.e. either the one we created or the external one). We used to edit `/etc/resolv.conf` on each node but it was not playing well with NetworkManager and the dnsmasq nodes openshift installs by default. By setting it on the subnet/network level, we can leave the nodes alone and everything works.

But I don't think we can even do the same thing with provider networks. They're generally shared between workloads and even tenants and we may not even have the permission to do anything to it. Not to mention that since it's a resource we didn't create, I don't want to mess with it (similar to not wanting to modify the Ansible host).

So what should we do here? If you use an external DNS and configure your provider network and the playbooks to use it, everything should just work. Is that sufficient or should we fall back to say editing `/etc/resolv.conf` on each node again?

@ggillies and @rlopez133 would very much appreciate your thoughts on the matter.

playbooks/provisioning/openstack/README.md
```diff
@@ -244,6 +244,26 @@ be the case for development environments. When turned off, the servers will
...
+
+Normally, the playbooks create a new Neutron network and subnet and attach
+floating IP addresses to each node. If you have a provider network set up, this
+is all unnecessary as you can just access servers that are placed in the
+provider network directly.
+
+To use a provider network, set its name in `openstack_provider_network_name` in
+`inventory/group_vars/all.yml`.
+
+You must also unset the `openstack_external_network_name` and
```
bogdando: ditto, let's make it ignored, with no additional changes required from users

playbooks/provisioning/openstack/prerequisites.yml
```diff
@@ -85,6 +87,20 @@
...
+    when:
+    - openstack_provider_network_name|default(None)
+    - openstack_private_network_name|default(None)
+
+  - name: The provider and external networks can't be set at the same time
+    fail:
+      msg: You cannot set both `openstack_provider_network_name` and `openstack_external_network_name` at the same time
+    when:
+    - openstack_provider_network_name|default(None)
+    - openstack_external_network_name|default(None)
```
bogdando: ditto. No need to fail, IMO, let it be ignored instead. Assuming users know that they want, when defining a provider network name.

roles/openstack-stack/tasks/subnet_update_dns_servers.yaml
```diff
@@ -6,3 +6,4 @@
     state: present
     use_default_subnetpool: yes
     dns_nameservers: "{{ [private_dns_server|default(public_dns_nameservers[0])]|union(public_dns_nameservers)|unique }}"
+  when: not openstack_provider_network_name|default(None)
```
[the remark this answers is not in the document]
tomassedovic: Oh right, I forgot this is within the openstack-stack module. Good catch!

roles/openstack-stack/templates/heat_stack.yaml.j2
```diff
@@ -73,6 +73,7 @@ outputs:
 
 resources:
 
+{% if not provider_network %}
   net:
```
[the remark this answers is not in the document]
tomassedovic: Oh! Good point.

playbooks/provisioning/openstack/sample-inventory/group_vars/all.yml
```diff
@@ -16,6 +16,11 @@ openstack_ssh_public_key: "openshift"
 openstack_external_network_name: "public"
 #openstack_private_network_name:  "openshift-ansible-{{ stack_name }}-net"
 
+## If you want to use a provider network, set its name here.
+## NOTE: you must remove the `openstack_external_network_name` and
```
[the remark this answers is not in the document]
tomassedovic: Okay, the problem is that these options conflict with each other. But yeah I guess we can just document that they're not going to be used when the provider network is set. Will do.

playbooks/provisioning/openstack/README.md
```diff
@@ -244,6 +244,26 @@ be the case for development environments. When turned off, the servers will
...
+`openstack_private_network_name` fields.
+
+**NOTE**: this will not update the nodes' DNS, so running openshift-ansible
+right after provisioning will fail (unless you're using an external DNS server
+your provider network knows about). You must make sure your nodes are able to
+resolve each other by name, e.g. by setting `dns_nameservers` on your subnet or
+editing each node's `/etc/resolv.conf`.
```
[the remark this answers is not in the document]
tomassedovic: You mean `/etc/resolv.conf` or the whole paragraph?

tomassedovic:
@bogdando this should address all your comments.

tomassedovic:
Don't test this. I messed something up during the initial rebase and the heat templating is failing.

tomassedovic:
Okay, the last commit fixed it. Ready for testing & reviewing.

playbooks/provisioning/openstack/stack_params.yaml
```diff
@@ -23,8 +23,14 @@ openstack_node_image: "{{ openstack_node_image_name | default(openstack_default_
 openstack_lb_image: "{{ openstack_lb_image_name | default(openstack_default_image_name) }}"
 openstack_etcd_image: "{{ openstack_etcd_image_name | default(openstack_default_image_name) }}"
 openstack_dns_image: "{{ openstack_dns_image_name | default(openstack_default_image_name) }}"
-openstack_private_network: "{{ openstack_private_network_name | default ('openshift-ansible-' + stack_name + '-net') }}"
-external_network: "{{ openstack_external_network_name }}"
+openstack_private_network: >-
+  {% if openstack_provider_network_name | default(None) -%}
+  {{ openstack_provider_network_name }}
```
bogdando: please add `|default('provider')`

tomassedovic: But the default case will be used (unless I'm missing something).

We're already in the `if` body so `openstack_provider_network_name` must be set to a value.

bogdando: oh, indeed

tomassedovic:
I'm going to run the end to end CI and wait for @ggillies and @rlopez133 to comment before merging

tomassedovic:
I've resolved the conflict in the readme and I'm going to merge this. @ggillies and @rlopez133 when you have time, please give this a look.

PR #704, opened 2017-08-29
[WIP] Provision lb node when multiple infra nodes

bogdando:
#### What does this PR do?
Provision a load balancer node, when having a multi-infra nodes setup.
The expected configuration for the LB node and the use case has to be elaborated and documented.
Currently, it gets no special configuration for infra nodes.

TBD

cc: @tomassedovic PTAL

### DIFF
--- a/roles/openstack-stack/templates/heat_stack.yaml.j2
+++ b/roles/openstack-stack/templates/heat_stack.yaml.j2
@@ -444,7 +444,7 @@ resources:
           remote_ip_prefix: "{{ openstack_subnet_prefix }}.0/24"
 {% endif %}
 
-{% if num_masters|int > 1 or ui_ssh_tunnel|bool %}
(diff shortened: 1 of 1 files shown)


bogdando:
What is the use case for a lb node and multi infra nodes? I can't find any docs or details on that. Which configuration the lb node is expected to get for multi infra nodes? What is the test case for it?

Shall the lb node's haproxy config provide a load balance for a docker registry, when deploying a multiple infra nodes?
I can see only a single logical frontend/backend definition stanza for master lb:

```
frontend  atomic-openshift-api
    bind *:8443
    default_backend atomic-openshift-api
    mode tcp
    option tcplog

backend atomic-openshift-api
    balance source
    mode tcp
    server      master0 192.168.99.12:8443 check
```

so in order to continue this PR, I need to know more for supported use cases of lb for registry and routers, and what else hosted at the infra nodes?..

bogdando:
It's time to resubmit on top of https://github.com/openshift/openshift-ansible/pull/6039

PR #706, opened 2017-08-30
Point openshift_master_cluster_public_hostname at master or lb if defined

tzumainn:
#### What does this PR do?
If openshift_master_cluster_public_hostname is defined, then create a DNS entry that points it at the master (single master) or a load balancer (multi-master)

Specify a fun value for openshift_master_cluster_public_hostname in OSEv3.yml, and re-run the provisioning playbook; make sure your DNS is update and ping the hostname. It should resolve to the same IP as your master/load balancer node

N/A

cc: @tomassedovic @bogdando @Tlacenka  PTAL

### DIFF
--- a/roles/dns-records/tasks/main.yml
+++ b/roles/dns-records/tasks/main.yml
@@ -50,6 +50,22 @@
   with_items: "{{ groups['infra_hosts'] }}"
   when: hostvars[item]['public_v4'] is defined
 
+- name: "Add public master cluster hostname records to the public A records"
+  set_fact:
+    public_records: "{{ public_records | default([]) + [ { 'type': 'A', 'hostname': hostvars[groups.masters[0]].openshift_master_cluster_public_hostname, 'ip': hostvars[groups.masters[0]].public_v4 } ] }}"
+  with_items: "{{ groups['cluster_hosts'] }}"
+  when:
+    - hostvars[groups.masters[0]].openshift_master_cluster_public_hostname is defined
+    - "{{ hostvars['localhost'].openstack_num_masters == 1 }}"
+
+- name: "Add public master cluster hostname records to the public A records"
+  set_fact:
+    public_records: "{{ public_records | default([]) + [ { 'type': 'A', 'hostname': hostvars[groups.masters[0]].openshift_master_cluster_public_hostname, 'ip': hostvars[groups.lb[0]].public_v4 } ] }}"
+  with_items: "{{ groups['cluster_hosts'] }}"
+  when:
+    - hostvars[groups.masters[0]].openshift_master_cluster_public_hostname is defined
+    - "{{ hostvars['localhost'].openstack_num_masters > 1 }}"
+
 - name: "Set the public DNS server details to use the external value (if provided)"
   set_fact:
     nsupdate_server_public: "{{ external_nsupdate_keys['public']['server'] }}"
--- a/roles/dns-records/tasks/main.yml
+++ b/roles/dns-records/tasks/main.yml
@@ -9,6 +9,20 @@
     private_records: "{{ private_records | default([]) + [ { 'type': 'A', 'hostname': '*.' + openshift_app_domain, 'ip': hostvars[item]['private_v4'] } ] }}"
   with_items: "{{ groups['infra_hosts'] }}"
 
+- name: "Add public master cluster hostname records to the private A records (single master)"
+  set_fact:
+    private_records: "{{ private_records | default([]) + [ { 'type': 'A', 'hostname': (hostvars[groups.masters[0]].openshift_master_cluster_public_hostname | replace(full_dns_domain, ''))[:-1], 'ip': hostvars[groups.masters[0]].private_v4 } ] }}"
+  when:
+    - hostvars[groups.masters[0]].openshift_master_cluster_public_hostname is defined
+    - openstack_num_masters == 1
+
+- name: "Add public master cluster hostname records to the private A records (multi-master)"
+  set_fact:
+    private_records: "{{ private_records | default([]) + [ { 'type': 'A', 'hostname': (hostvars[groups.masters[0]].openshift_master_cluster_public_hostname | replace(full_dns_domain, ''))[:-1], 'ip': hostvars[groups.lb[0]].private_v4 } ] }}"
+  when:
+    - hostvars[groups.masters[0]].openshift_master_cluster_public_hostname is defined
+    - openstack_num_masters > 1
+
 - name: "Set the private DNS server to use the external value (if provided)"
   set_fact:
     nsupdate_server_private: "{{ external_nsupdate_keys['private']['server'] }}"
--- a/roles/dns-records/tasks/main.yml
+++ b/roles/dns-records/tasks/main.yml
@@ -52,16 +52,14 @@
 
 - name: "Add public master cluster hostname records to the public A records (single master)"
   set_fact:
-    public_records: "{{ public_records | default([]) + [ { 'type': 'A', 'hostname': hostvars[groups.masters[0]].openshift_master_cluster_public_hostname, 'ip': hostvars[groups.masters[0]].public_v4 } ] }}"
-  with_items: "{{ groups['cluster_hosts'] }}"
+    public_records: "{{ public_records | default([]) + [ { 'type': 'A', 'hostname': (hostvars[groups.masters[0]].openshift_master_cluster_public_hostname | replace(full_dns_domain, ''))[:-1], 'ip': hostvars[groups.masters[0]].public_v4 } ] }}"
   when:
     - hostvars[groups.masters[0]].openshift_master_cluster_public_hostname is defined
     - openstack_num_masters == 1
 
 - name: "Add public master cluster hostname records to the public A records (multi-master)"
   set_fact:
-    public_records: "{{ public_records | default([]) + [ { 'type': 'A', 'hostname': hostvars[groups.masters[0]].openshift_master_cluster_public_hostname, 'ip': hostvars[groups.lb[0]].public_v4 } ] }}"
-  with_items: "{{ groups['cluster_hosts'] }}"
+    public_records: "{{ public_records | default([]) + [ { 'type': 'A', 'hostname': (hostvars[groups.masters[0]].openshift_master_cluster_public_hostname | replace(full_dns_domain, ''))[:-1], 'ip': hostvars[groups.lb[0]].public_v4 } ] }}"
   when:
     - hostvars[groups.masters[0]].openshift_master_cluster_public_hostname is defined
     - openstack_num_masters > 1
--- a/roles/dns-records/tasks/main.yml
+++ b/roles/dns-records/tasks/main.yml
@@ -56,7 +56,7 @@
   with_items: "{{ groups['cluster_hosts'] }}"
   when:
     - hostvars[groups.masters[0]].openshift_master_cluster_public_hostname is defined
-    - "{{ openstack_num_masters == 1 }}"
+    - openstack_num_masters == 1
 
 - name: "Add public master cluster hostname records to the public A records (multi-master)"
   set_fact:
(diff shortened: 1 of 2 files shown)


tzumainn:
This code definitely works for both single master and multi-master; however I'm pretty sure there's a more proper way to do this. Just let me know what it is!

roles/dns-records/tasks/main.yml
```diff
@@ -50,6 +50,22 @@
   with_items: "{{ groups['infra_hosts'] }}"
   when: hostvars[item]['public_v4'] is defined
 
+- name: "Add public master cluster hostname records to the public A records"
+  set_fact:
+    public_records: "{{ public_records | default([]) + [ { 'type': 'A', 'hostname': hostvars[groups.masters[0]].openshift_master_cluster_public_hostname, 'ip': hostvars[groups.masters[0]].public_v4 } ] }}"
```
bogdando: could it be just the `openshift_master_cluster_public_hostname` inventory var?

roles/dns-records/tasks/main.yml
```diff
@@ -50,6 +50,22 @@
...
+  with_items: "{{ groups['cluster_hosts'] }}"
+  when:
+    - hostvars[groups.masters[0]].openshift_master_cluster_public_hostname is defined
```
bogdando: ditto

roles/dns-records/tasks/main.yml
```diff
@@ -50,6 +50,22 @@
...
+    - "{{ hostvars['localhost'].openstack_num_masters == 1 }}"
```
bogdando: just `openstack_num_masters` should work as well as we define it in the all groups all.yml

roles/dns-records/tasks/main.yml
tomassedovic: `openshift_master_cluster_public_hostname` is specified in OSEv3 so I don't think it's exposed in this role directly. But yeah @tzumainn can you test if you could go away with the hostvars business here?

roles/dns-records/tasks/main.yml
```diff
@@ -50,6 +50,22 @@
...
+
+- name: "Add public master cluster hostname records to the public A records"
+  set_fact:
+    public_records: "{{ public_records | default([]) + [ { 'type': 'A', 'hostname': hostvars[groups.masters[0]].openshift_master_cluster_public_hostname, 'ip': hostvars[groups.lb[0]].public_v4 } ] }}"
+  with_items: "{{ groups['cluster_hosts'] }}"
+  when:
+    - hostvars[groups.masters[0]].openshift_master_cluster_public_hostname is defined
+    - "{{ hostvars['localhost'].openstack_num_masters > 1 }}"
```
bogdando: ditto, for both vars

roles/dns-records/tasks/main.yml
tzumainn: Hi! I'm using the value defined in OSEv3.yml as Tomas says, and it looks like just specifying openshift_master_cluster_public_hostname doesn't work in that case.

roles/dns-records/tasks/main.yml
tzumainn: Yep, that works - updated!

tzumainn:
Thanks for the review! I've updated openstack_num_masters, but the other variable is located in OSEv3.yml, and doesn't seem to like a different way of calling it. Let me know if you'd like me to look at an alternative!

roles/dns-records/tasks/main.yml
```diff
@@ -50,6 +50,22 @@
...
+    - hostvars[groups.masters[0]].openshift_master_cluster_public_hostname is defined
+    - "{{ openstack_num_masters == 1 }}"
+
+- name: "Add public master cluster hostname records to the public A records (multi-master)"
+  set_fact:
+    public_records: "{{ public_records | default([]) + [ { 'type': 'A', 'hostname': hostvars[groups.masters[0]].openshift_master_cluster_public_hostname, 'ip': hostvars[groups.lb[0]].public_v4 } ] }}"
+  with_items: "{{ groups['cluster_hosts'] }}"
+  when:
+    - hostvars[groups.masters[0]].openshift_master_cluster_public_hostname is defined
+    - "{{ openstack_num_masters > 1 }}"
```
bogdando: this may cause an ansible warning, it should work just as ` - openstack_num_masters > 1` AFAIK

tzumainn: ah, you're right! fixed

tzumainn:
@bogdando thanks for the added comment, it should be resolved now!

tomassedovic:
I've added end to end CI and I'm testing it locally.

tomassedovic:
@bogdando I'm confused: did you approve this without testing it? It doesn't work for me.

I have `openshift_master_cluster_public_hostname: "api.{{ env_id }}.{{ public_dns_domain }}"` in `OSEv3.yml` and after provisioning, the public entry point isn't resolving:

```
$ nslookup api.openshift.example.com 10.40.128.135
Server:         10.40.128.135
Address:        10.40.128.135#53

** server can't find api.openshift.example.com: NXDOMAIN
```
but the master hostname is as usual:

```
thomas@worklaptop:~/code/openshift
[devbox demo]
$ nslookup master-0.openshift.example.com 10.40.128.135
Server:         10.40.128.135
Address:        10.40.128.135#53

Name:   master-0.openshift.example.com
Address: 10.40.128.82
```

tomassedovic:
@tzumainn it seems to be because the hostname portion should only have the part without the DNS zone (domin), not the full thing.

Compare the hostname entry in the ansible output for the master hostname vs. the ui/api:

```
TASK [infra-ansible/roles/dns : Add DNS A records] ********************************************************************************************************************
Friday 01 September 2017  12:37:22 +0200 (0:00:00.014)       0:01:32.933 ****** 
changed: [localhost] => (item=({u'key_algorithm': u'hmac-sha256', u'zone': u'openshift.example.com', u'key_name': u'public-openshift.example.com', u'server': u'10.40.1
28.135', u'key_secret': u'GWrRGTmom/AAUpBekzViJEPmwml1tD9cOHP1Cjc4Cx0=', u'view': u'public'}, {'ip': u'10.40.128.82', 'hostname': u'master-0', 'type': u'A'}))
ok: [localhost] => (item=({u'key_algorithm': u'hmac-sha256', u'zone': u'openshift.example.com', u'key_name': u'public-openshift.example.com', u'server': u'10.40.128.13
5', u'key_secret': u'GWrRGTmom/AAUpBekzViJEPmwml1tD9cOHP1Cjc4Cx0=', u'view': u'public'}, {'ip': u'10.40.128.82', 'hostname': u'api.openshift.example.com', 'type': u'A'
}))
```

bogdando:
@tomassedovic yes, I approve for a code-review. It doesn't cover testing.

tomassedovic:
@bogdando okay, no worries. Will you please make that explicit from now on, though? I.e. if you didn't test it, say that you just looked at the code.

roles/dns-records/tasks/main.yml
```diff
@@ -50,6 +50,22 @@
   with_items: "{{ groups['infra_hosts'] }}"
   when: hostvars[item]['public_v4'] is defined
 
+- name: "Add public master cluster hostname records to the public A records (single master)"
+  set_fact:
+    public_records: "{{ public_records | default([]) + [ { 'type': 'A', 'hostname': hostvars[groups.masters[0]].openshift_master_cluster_public_hostname, 'ip': hostvars[groups.masters[0]].public_v4 } ] }}"
```
tomassedovic: Change the `hostname` contents to:

    (hostvars[groups.masters[0]].openshift_master_cluster_public_hostname | replace(full_dns_domain, ''))[:-1]

That will turn say `api.openshift.example.com` to `api`, which is what we need for nsupdate.

tzumainn:
@tomassedovic hi! I may be confused,  but looking at https://github.com/openshift/openshift-ansible-contrib/blob/master/playbooks/provisioning/openstack/sample-inventory/group_vars/OSEv3.yml#L9 , I had thought that the default would result in something more like `api` rather than `api.openshift.example.com`... ? If so, wouldn't it make sense for the user to input something similar?

tzumainn:
@tomassedovic There's already an example in sample-inventory/group_vars/OSEv3.yml, isn't there? Or did I end up confusing variable names?

tomassedovic:
@tzumainn yeah we do have an example in the sample inventory, sorry. I expected this change to add it and forgot that it was there already. Though I'd change the suggested value to something like `api.{{ env_id }}.{{ public_dns_domain }}`, then.

But yea, the contents are a fully qualified domain name, not just e.g. the "api" portion of it. See the openshift-ansible sample inventory:

https://github.com/openshift/openshift-ansible/blob/4a11ed0808f508d7e9d0953149377e2e181bdae9/inventory/byo/hosts.ose.example#L297

If you take a look into your `inventory/hosts` the `masters` and `lb` groups use the full names (they correspond to the hostnames of the VMs).

tzumainn:
@tomassedovic got it! made the change you requested, let me know if I forgot anything!

roles/dns-records/tasks/main.yml
```diff
@@ -50,6 +50,20 @@
...
+    - hostvars[groups.masters[0]].openshift_master_cluster_public_hostname is defined
+    - openstack_num_masters == 1
+
+- name: "Add public master cluster hostname records to the public A records (multi-master)"
+  set_fact:
+    public_records: "{{ public_records | default([]) + [ { 'type': 'A', 'hostname': (hostvars[groups.masters[0]].openshift_master_cluster_public_hostname | replace(full_dns_domain, ''))[:-1], 'ip': hostvars[groups.lb[0]].public_v4 } ] }}"
+  when:
+    - hostvars[groups.masters[0]].openshift_master_cluster_public_hostname is defined
+    - openstack_num_masters > 1
+
```
tomassedovic: So the CI failure (and my local testing) both failed and it's because we apparently need to set the records for the private views as well.

Adding this fixed the single master scenario for me so please put that in with the private records and do the same for the multimaster setup:

```
- name: "Add public master cluster hostname records to the private A records (single master)"
  set_fact:
    private_records: "{{ private_records | default([]) + [ { 'type': 'A', 'hostname': (hostvars[groups.masters[0]].openshift_master_cluster_public_hostname | replace(full_dns_domain, ''))[:-1], 'ip': hostvars[groups.masters[0]].private_v4 } ] }}"
  when:
    - hostvars[groups.masters[0]].openshift_master_cluster_public_hostname is defined
    - openstack_num_masters == 1
```

tzumainn:
@tomassedovic thanks for the suggestion! Tried it like you said, and tested end-to-end with success - thanks!

tomassedovic:
Thanks @tzumainn! This works great! I'll merge it once the e2e ci passes.

tomassedovic:
Hm, the ci's still failing. Let me try the default option locally (I've only tested setting the URL, not leaving it unset).

tomassedovic:
It worked on me locally. I wonder if the CI is broken for master. I've triggered a run on the master branch here:

https://travis-ci.org/openshift/openshift-ansible-contrib/builds/271628145

And I'm re-running the default locally -- with the latest openshift-ansible. Maybe we have another breakage.

tomassedovic:
Right, okay. I'm seeing the same error locally with the latest openshift-ansible. We'll pin the CI to a working version, re-run it here and file an issue upstream.

tomassedovic:
Conveniently, Heat seems to be not working now:

https://travis-ci.org/openshift/openshift-ansible-contrib/jobs/271628147

tomassedovic:
Running with the CI fix: https://travis-ci.org/openshift/openshift-ansible-contrib/jobs/272429317

tomassedovic:
@tzumainn can you merge or rebase the latest master please? I think that needs to be done before we're allowed to merge this.

PR #707, opened 2017-08-31
Support Cinder-backed Openshift registry

tomassedovic:
#### What does this PR do?
It lets you set up the openshift registry with an existing pre-configured Cinder volume.

1. Before provisioning run: `openstack volume create --size 10 openshift-registry` and note it's UUID.
2. Add `prepare_and_format_registry_volume: true` to `inventory/group_vars/all.yml`
3. Add your openstack credentials to `inventory/group_vars/OSEv3.yml`

```
openshift_cloudprovider_kind: openstack
openshift_cloudprovider_openstack_username: <username>
openshift_cloudprovider_openstack_password: <password>
openshift_cloudprovider_openstack_auth_url: <auth url>
openshift_cloudprovider_openstack_tenant_name: <tenant name>
```
4. Configure openshift-ansible to use the Cinder volume by adding this to `inventory/group_vars/OSEv3.yml`:

```
openshift_hosted_registry_storage_kind: openstack
openshift_hosted_registry_storage_access_modes: ['ReadWriteOnce']
openshift_hosted_registry_storage_openstack_filesystem: xfs
openshift_hosted_registry_storage_openstack_volumeID: <volume uuid>
openshift_hosted_registry_storage_volume_size: 10Gi
```

5. Run the end to end tests -- they should all pass.
6. Validate that the registry is attached to the infra node: `openstack volume list` -- the volume should be `in-use`, attached to the infra node usually as `/dev/vdb`
7. SSH to the infra node and verify that the volume is mounted: `mount | grep vdb` (this should show a couple of mount points)

n/a

cc: @tzumainn @Tlacenka  @bogdando PTAL


playbooks/provisioning/openstack/README.md
```diff
@@ -263,6 +263,69 @@ variables for the `inventory/group_vars/OSEv3.yml`, `all.yml`:
...
+The volume needs to have a file system created before you put it to
+use. We can do prepare it for you if you put this in inventory/group_vars/all.yml:
+
+    prepare_and_format_registry_volume: true
+
+**NOTE:** doing so **will destroy any data that's currently on the volume**!
+
+You can also run the registry setup playbook directly:
+
+   ansible-playbook -i inventory playbooks/provisioning/openstack/prepare-and-format-cinder-volume.yaml
```
bogdando: a whitespace is missing

playbooks/provisioning/openstack/README.md
```diff
@@ -263,6 +263,69 @@ variables for the `inventory/group_vars/OSEv3.yml`, `all.yml`:
...
+    ## Openstack credentials
+    #openshift_cloudprovider_kind=openstack
+    #openshift_cloudprovider_openstack_auth_url=http://openstack.example.com:35357/v2.0/
+    #openshift_cloudprovider_openstack_username=username
+    #openshift_cloudprovider_openstack_password=password
+    #openshift_cloudprovider_openstack_domain_id=domain_id
+    #openshift_cloudprovider_openstack_domain_name=domain_name
+    #openshift_cloudprovider_openstack_tenant_id=tenant_id
+    #openshift_cloudprovider_openstack_tenant_name=tenant_name
+    #openshift_cloudprovider_openstack_region=region
```
bogdando: could we reuse the exported env vars? `"{{ lookup('env', 'OS_FOO') }}"`

playbooks/provisioning/openstack/sample-inventory/group_vars/OSEv3.yml
```diff
@@ -10,6 +10,26 @@ openshift_master_cluster_public_hostname: "{{ groups.lb.0|default(groups.masters
...
+## Openstack credentials
+#openshift_cloudprovider_kind=openstack
+#openshift_cloudprovider_openstack_auth_url=http://openstack.example.com:35357/v2.0/
+#openshift_cloudprovider_openstack_username=username
+#openshift_cloudprovider_openstack_password=password
+#openshift_cloudprovider_openstack_domain_id=domain_id
+#openshift_cloudprovider_openstack_domain_name=domain_name
+#openshift_cloudprovider_openstack_tenant_id=tenant_id
+#openshift_cloudprovider_openstack_tenant_name=tenant_name
+#openshift_cloudprovider_openstack_region=region
```
bogdando: ditto

playbooks/provisioning/openstack/README.md
tzumainn: I'm guessing this is because openshift-ansible already specifies these variables, so this just follows along with that?

tzumainn: Dumb question: are these values used in the playbook changes made in the PR? Or are they simply required by openshift-ansible playbooks if attaching a pre-existing volume?

tomassedovic:
@bogdando I planned to do that in a separate PR but yea can do that here.

Just to clarify though, these variables are required by openshift-ansible and may be different from the ones you use to provision. So we can the OS_* lookups as an example, but it's not applicable in all cases.

playbooks/provisioning/openstack/README.md
tomassedovic: @tzumainn the latter: they're openshift-ansible specific and they're used when attaching an existing volume. And they may differ from the creds used in provisioning.

tzumainn:
I think that if the inventory variables are required by the openshift-ansible playbooks, it's worth mentioning them here; I could see that preventing a bunch of confusion!

tomassedovic:
@tzumainn @bogdando I've updated the docs and showed how to read the creds from the environment in the sample inventory. Let me know if this is better.

tzumainn:
Figured out why the ansible deployment wasn't working for me; it has nothing to do with this PR, but with this known issue: https://github.com/kubernetes/kubernetes/issues/50461. I hacked the openshift-ansible files to create an openstack cloud config file that contains the described workaround, and everything worked as expected.

One other note: if we're going to recommend using creds from the environment, I might add a note mentioning that the auth-url can't be versionless, as that causes the ansible deployment to fail, and the default devstack identity endpoint is versionless.

tomassedovic:
Oh, wow, okay! Thanks for figuring that out. I'll update the readme.

tomassedovic:
@tzumainn I've updated the readme and merged master. Do you have any other requests or suggestions? If not, could you please click on "Files chaned", then "Review changes" and add your approval?

I'll run the end to end CI in the meantime.

tomassedovic:
e2e ci passed \o/

@tzumainn you should have hopefully received commit rights to this repo. Could you please accept them and see if you can approve this again? We should be able to merge this after wards.

playbooks/provisioning/openstack/sample-inventory/group_vars/OSEv3.yml
```diff
@@ -10,6 +10,9 @@ openshift_master_cluster_public_hostname: "{{ groups.lb.0|default(groups.masters
 
 osm_default_node_selector: 'region=primary'
 
+openshift_hosted_router_wait: True
+aopenshift_hosted_registry_wait: True
```
tzumainn: I think this is a typo - "openshift_hosted_registry_wait"?

tomassedovic: Ah indeed!

PR #714, opened 2017-09-05
Fix openstack CI

tomassedovic:
#### What does this PR do?
This attempts to fix the end to end CI by adding manual checks to ensure the router and registry pods are actually up.

If this works, we may want to turn that into a post-install playbook.

No manual testing require, just verify the e2e ci passed.

n/a

cc: @tzumainn @Tlacenka  @bogdando PTAL

### DIFF
--- a/ci/openstack/install-openshift.sh
+++ b/ci/openstack/install-openshift.sh
@@ -45,3 +45,12 @@ for i in $(seq 15); do
         printf .
         sleep 60
 done
+
+echo oc get nodes --show-labels:
+ansible -i "$INVENTORY" masters --user openshift --private-key ~/.ssh/id_rsa -m command -a 'oc get nodes --show-labels'
+
+echo oc status -v:
+ansible -i "$INVENTORY" masters --user openshift --private-key ~/.ssh/id_rsa -m command -a 'oc status -v'
+
+echo oc get all:
+ansible -i "$INVENTORY" masters --user openshift --private-key ~/.ssh/id_rsa -m command -a 'oc get all'
--- a/ci/openstack/install-openshift.sh
+++ b/ci/openstack/install-openshift.sh
@@ -31,6 +31,7 @@ for i in $(seq 15); do
     else
         printf .
         sleep 60
+    fi
 done
 
 echo Waiting for the docker-registry to come up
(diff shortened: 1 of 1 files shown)


tomassedovic:
Yay, the CI passed. I'm going to do a very naughty thing and merge this myself, because this is blocking several open pull requests and there's no one who can approve this in my time + zone.

tomassedovic:
Ha! Or not. Turns out I can't approve my patches. Well played, github, well played.

tomassedovic:
@e-minguez @pschiffe any chance you can give this a look & approve? The openstack end to end CI is broken without it.

ci/openstack/install-openshift.sh
```diff
@@ -18,3 +18,41 @@ export INVENTORY="$PWD/playbooks/provisioning/openstack/sample-inventory"
...
 
 ansible-playbook --become --timeout 180 --user openshift --private-key ~/.ssh/id_rsa -i "$INVENTORY" ../openshift-ansible/playbooks/byo/config.yml -e @extra-vars.yaml
+
+
+echo Waiting for the router to come up
+for i in $(seq 15); do
+    if ansible -i "$INVENTORY" masters --user openshift --private-key ~/.ssh/id_rsa -m shell -a 'oc get pod | grep router | grep Running'; then
+        echo Router is running
+        break
+    elif ansible -i "$INVENTORY" masters --user openshift --private-key ~/.ssh/id_rsa -m shell -a 'oc get pod | grep router | grep Failed'; then
```
e-minguez: You can avoid use shell ansible module and grep as:

```
oc get pods -l deploymentconfig=router -n default
```

e-minguez: Silly comment from my side, otherwise LGTM

tomassedovic: No, that's great. One question though, doesn't this just print out the status? We need the command to fail if it's not Running or Failed. The last grep fails there which gets propagated to Ansible.

e-minguez: Yep, you are right, grep is needed (I think jsonpath supports true or false output as well https://github.com/kubernetes/kubernetes/pull/39063 but I've been playing for 10 minutes and couldn't get it :D)

PR #717, opened 2017-09-07
Adding 3.6 support to VMware platform

dav1x:
#### What does this PR do?
- Allow for multiple deployments of OCP with a unique cluster_id
- Cleanup vsphere by dynamic inventory instead of inventory file
- Enhancements to rhsm and rhsm-{repos, subscription}
- Add cloud provider config file but not the automated installation

cd ~; git clone -b vmw-3.6;cd ~/openshift-ansible-contrib/reference-architecture/vmware-ansible
./ocp-on-vmware.py --create_inventory; ./ocp-on-vmware.py --create_ocp_vars; ./ocp-on-vmware.py

N/A

cc: @cooktheryan @e-minguez @rlopez133  PTAL


roles/rhsm-subscription/tasks/main.yaml
```diff
@@ -53,12 +48,12 @@
         password: "{{ rhsm_password }}"
         state: present
         pool: "{{ rhsm_pool }}"
-      when: "'Current' not in subscribed.stdout and rhsm_user is defined and rhsm_user"
+      when: "'not registered' in subscribed.stdout and rhsm_user is defined and rhsm_user"
```
cooktheryan: Can we add an or in for 'Current' I know that @pschiffe needed something like this bc there are a few options in what the output is

PR #739, opened 2017-09-19
Integrate SSO into 3.6 Ref Arch

glennswest:
#### What does this PR do?
This PR adds support for SSO/Keycload for authenticating developers using OpenShift, and
optionally applications running in OpenShift. It requires no user interaction for deployment, 
automatically creating all keys and certificates, creating a Openshift Client in SSO, and automatically editing Openshift Master Config, and restarting needed services. 
Note: This is only for ocp 3.6 tree.

Deploy normally ocp 3.6 single vm or ha ref arch, and after completion of install, log into console.

New Functionality

cc: @e-minguez @cooktheryan


reference-architecture/azure-ansible/3.6/allinone.sh
```diff
@@ -431,6 +431,213 @@ cat <<EOF > /home/${AUSERNAME}/postinstall.yml
...
+         SSO_ADMIN_PASSWORD="{{sso_password}}"
+         SSO_TRUSTSTORE=truststore.jks
+         SSO_TRUSTSTORE_PASSWORD="{{idm_xpassword}}"
+
+  - name: Stage 10 - OCCREATE SECRET ADD
+    command: oc new-app sso71-postgresql --param-file {{idm_dir}}/sso.params -l app=sso71-postgresql -l application=sso -l template=sso71-https
+  - set_fact: sso_token_url="https://login.{{sso_domain}}/auth/realms/cloud/protocol/openid-connect/token"
+  - name: Pause for app create
+    pause:
+      minutes: 4
```
e-minguez: I'd say to wait until the pod is running instead just waiting a number of minutes

reference-architecture/azure-ansible/3.6/allinone.sh
```diff
@@ -431,6 +431,213 @@ cat <<EOF > /home/${AUSERNAME}/postinstall.yml
...
+      - "result.rc > 10"
+  - name: Delete Old Project
+    command: oc delete project "{{sso_project}}"
+    ignore_errors: yes
+    register: result
+    failed_when: 
+      - "result.rc > 10"
+  - name: Pause for cleanup of old install
+    pause:
+      minutes: 2
```
e-minguez: I'd say to wait until the project has been deleted instead a number of minutes

reference-architecture/azure-ansible/3.6/allinone.sh
```diff
@@ -431,6 +431,213 @@ cat <<EOF > /home/${AUSERNAME}/postinstall.yml
...
+  - set_fact: sso_projectid="{{sso_project}}"
+  - set_fact: idm_xpassword="Xp-{{sso_password}}"
+  - name: Create Openshift Project for SSO
+    command: oc new-project "{{sso_project}}"
+  - name: Create Service Account
+    command: "oc create serviceaccount {{sso_project}}-service-account"
```
e-minguez: Just in case, I'll add the "-n {{ sso_project }}" to ensure the service account is created in the proper project

reference-architecture/azure-ansible/3.6/allinone.sh
```diff
@@ -431,6 +431,213 @@ cat <<EOF > /home/${AUSERNAME}/postinstall.yml
...
+  - name: Add admin role to user
+    command: "oadm policy add-role-to-user admin {{sso_username}}"
```
e-minguez: oadm is deprecated, we should use "oc adm" instead

reference-architecture/azure-ansible/3.6/allinone.sh
```diff
@@ -431,6 +431,213 @@ cat <<EOF > /home/${AUSERNAME}/postinstall.yml
...
+      return_content: yes
+      status_code: 201
+      validate_certs: no
+    register: create
+  - debug: var=create.json.secret
+- hosts: masters
+  vars_files:
+    - ssovars.yml
+  tasks:
+  - name: Copy xpass.conf to masters
```
e-minguez: xpass.crt

reference-architecture/azure-ansible/3.6/allinone.sh
```diff
@@ -431,6 +431,213 @@ cat <<EOF > /home/${AUSERNAME}/postinstall.yml
...
+    copy:
+      src:  "{{idm_dir}}/xpaas.crt"
+      dest: /etc/origin/master/xpaas.crt
+      owner: root
+      mode: 0600
+  - name: Setup SSO Config
+    blockinfile:
+      backup: yes
+      dest: /etc/origin/master/master-config.yaml
+      insertafter: HTPasswdPasswordIdentityProvider
```
e-minguez: OCP allows having multiple identity providers. In this particular case it will be htpasswd and sso, is that on purpose? Otherwise I'll remove the htpasswd.

glennswest: Just in case, as its chicken and egg problem, if sso does deploy you lose the ability to login.

glennswest: Because all the rest of the providers can be used inside of sso. So sso gives you a gui for adding providers and controlling them.

reference-architecture/azure-ansible/3.6/allinone.sh
glennswest: The problem is the project may be deleted, but there is still stuff in the background that is happening.

glennswest: Actually you can rerun the deploy script, and it will add back the sso. We could also give a sso-update.yml add on to recover to re-edit the config.

reference-architecture/azure-ansible/3.6/allinone.sh
glennswest: Actually I found a bug in the java side, I was doing a rest call to see if alive, and found it hosed to sso process and it would never work, so the sleep followed by the retry logic fixes this.

glennswest:
@e-minguez 
On "If the master-config.yaml is different than the hosts file for the OCP installation, if you run the installer again (or upgrade), it can remove the changes you made, so I'd say to modify the hosts file to include sso config if deployed." 

The problem you have, is if you really are doing a re-install, likely the sso container and related are gone. 

The appropriate fix is to re-run the sso install, "ansible-playbook setup-sso.yml", or
restore a backup of the config (If it was accidently deleted, or copy-past the last section of the setup-sso.yml assuming the app is still running. 

Reinstalls usually wipe any "old" containers, so dont thing anything would survive. and setup-sso would have to be used.

e-minguez:
I'm approving the PR but I think the two authentication methods note should be added to the ref. arch. and still, I don't like sleeps :P

PR #745, opened 2017-09-19
load balancer formatting fix

tzumainn:
#### What does this PR do?
Currently the heat stack template has incorrect formatting when a load balancer is needed. This PR fixes that issue.

Specify 2 master nodes in all.yml so that a load balancer is provisioned. Run the openstack provisioning playbook without this fix; there should be a formatting error when ansible tries to create the heat stack. Re-run with this fix, and provisioning should finish without errors.

N/A

cc: @tomassedovic @bogdando  PTAL

### DIFF
--- a/roles/openstack-stack/templates/heat_stack.yaml.j2
+++ b/roles/openstack-stack/templates/heat_stack.yaml.j2
@@ -577,14 +577,14 @@ resources:
           secgrp:
             - { get_resource: lb-secgrp }
             - { get_resource: common-secgrp }
-    {% if not provider_network %}
+{% if not provider_network %}
           floating_network: {{ external_network }}
-    {% endif %}
+{% endif %}
           volume_size: {{ lb_volume_size }}
-    {% if not provider_network %}
+{% if not provider_network %}
     depends_on:
       - interface
-    {% endif %}
+{% endif %}
 {% endif %}
 
   masters:


mbruzek:
I believe this change is valid. I made the change by hand and was able to get a little further in my cluster install. +1

tzumainn:
@bogdando Are there other places that have the same indentation? Because heat_stack.yaml.j2 uses no indents even in nested if statements - for example: https://github.com/tzumainn/openshift-ansible-contrib/blob/7abba3f9601fba07d0dc58b58bed648b9dfb4f86/roles/openstack-stack/templates/heat_stack.yaml.j2#L774-L823

The version that gives me formatting errors is the one on master. Do you not get errors, even when you specify multiple masters?

tomassedovic:
Oh my bad! I must have messed up the formatting in the provisioning network patch.

tomassedovic:
Hm, so the CI seems to install Ansible 2.4 and that broke us somehow. We'll have to fix that (I guess by requiriing 2.3 for now and by making sure we work with both going forward).

tomassedovic:
@tzumainn we've just merged the CI fix:

https://github.com/openshift/openshift-ansible-contrib/pull/750

If you could merge with the latest master branch, I'll re-run the CI and we should be good to go.

tzumainn:
@tomassedovic thanks - done!

PR #752, opened 2017-09-22
Fix public master cluster DNS record when using bastion

bogdando:
#### What does this PR do?
When using a bastion and a single master, add the bastion node's public IP the public master's IP for the DNS record.

e2e with a single master and bastion enabled shall pass

cc: @tomassedovic @tzumainn  PTAL


bogdando:
@tomassedovic works for me, PTAL.

Issue #749, opened 2017-09-21
RFE: Specific pool should be used for masters nodes running on RHEL


patlachance:
# FEATURE REQUEST INFO:
Specific pool should be used for masters nodes running on RHEL

## **Problem Description**
According to this document https://access.redhat.com/solutions/2851701, infrastructure node (OpenShift master, etcd, registry, metrics, logging, router, etc.) are free and should be attached to the **Infrastructure SKU pool** *(Red Hat OpenShift Container Platform Broker/Master Infrastructure)*  whereas nodes should be attached to one of the **Primary Node SKUs** such as:
- Red Hat OpenShift Container Platform, Standard, 2-Core
- Red Hat OpenShift Container Platform, Premium, 2-Core
- Red Hat OpenShift Container Platform, Standard (1-2 Sockets)
- Red Hat OpenShift Container Platform, Premium (1-2 Sockets)

### *Overview*
Either add a new option to the installer script(s) or detect appropriate pool to use when the playbook is executed on master nodes.

### *Alternatives*
No other I can think of

### *Security Impact*
None

### *Other End User Impact*
Reduce Red Hat subscription costs

### *Performance Impact*
None

### *Other Deployer Impact*
None

### *Developer Impact*
None

### *Documentation Impact*
Need to document the new installer option if it is implementation scenario taken

## **Acceptance criteria**
Master nodes should be attached to the "Red Hat OpenShift Container Platform Broker/Master Infrastructure" pool

## **Dependencies**
Only when installing on RHEL instances

## **References**
Somehow related to [#424], but motivation is slightly different.

PR #753, opened 2017-09-22
Attach additional RHN Pools (post-provision custom action)

Tlacenka:
#### What does this PR do?
This PR adds a custom post-provision action playbook to the `custom-actions` folder and describes this as an example in the provisioning `README`

1) Run the provisioning using RHEL images.
2) Find out which other pools are available by running `subscription-manager list --available` command on one of the hosts.
3) Run `add-rhn-pools` to add these new pools like so:
```
ansible-playbook -i <inventory> openshift-ansible-contrib/playbooks/provisioning/openstack/custom-actions/add-rhn-pools.yml --extra-vars '{"rhn_pools":["<pool ID>","<pool ID>"]}' 
```
4) List available pools on one of the hosts by running `subscription-manager list --consumed` and confirm that the new pools have been added. 
--

cc: @tomassedovic @bogdando @tzumainn PTAL

### DIFF
--- a/playbooks/provisioning/openstack/README.md
+++ b/playbooks/provisioning/openstack/README.md
@@ -511,7 +511,9 @@ If you'd like to limit the run to one particular host, you can do so as follows:
 ansible-playbook --private-key ~/.ssh/openshift -i inventory/ openshift-ansible-contrib/playbooks/provisioning/openstack/custom-actions/custom-playbook.yml -l app-node-0.openshift.example.com
 ```
 
-You can also create your own custom playbook. Here's one example that adds additional YUM repositories:
+You can also create your own custom playbook. Here are a few examples:
+
+#### Adding additional YUM repositories
 
 ```
 ---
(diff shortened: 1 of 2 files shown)


playbooks/provisioning/openstack/README.md
```diff
@@ -535,13 +537,31 @@ This example runs against app nodes. The list of options include:
...
+---
+- hosts: cluster_hosts
+  tasks:
+  - name: Attach additional RHN pool
+    become: true
+    command: "/usr/bin/subscription-manager attach --pool=<pool ID>"
+    register: attach_rhn_pool_result
+    until: attach_rhn_pool_result.rc == 0
+    retries: 10
+    delay: 1
```
ioggstream: This delay may be too low for subscription-manager related issues.

ioggstream:
Shouldn't we support multiple pools in here to be able to install OCP-on-OSP?

 https://github.com/openshift/openshift-ansible-contrib/blob/master/roles/subscription-manager/tasks/main.yml#L119

patlachance:
How about handling RHN subscriptions by node type as in #749?

Tlacenka:
@bogdando @ioggstream So if I understand it correctly, you want this to become a part of the `subscription-manager` role? No problem but it kind of defeats the purpose of it being just a custom post-provision action. Do we need this every time we run provisioning? ( @tomassedovic as the one who created this task, what do you think?)

playbooks/provisioning/openstack/README.md
Tlacenka: what delay do you propose then?

ioggstream: afaik the cdn.redhat.com DNS record  TTL is 20 seconds. Maybe 10 or 15 seconds could be fine.

ioggstream:
@Tlacenka if this is just a postinstall, we could:

  - implement the feature in the subscription_manager
  - reuse that role in this custom action, avoiding calling command

playbooks/provisioning/openstack/README.md
Tlacenka: I see. I just used the same values as in the rest of `subscription-manager` tasks. I guess those should be adjusted as well then?

tomassedovic: @ioggstream this is a delay between retries not a timeout. If the task takes 30 seconds to succeed, this won't kill it.

It's just that if it fails, we will wait 1 second and retry. And do that 10 times.

tomassedovic:
The intention of these sample playbooks is just to show people how to run some custom post-provision actions. It's mostly documentation that you happen to be able to run as a bonus.

If the deployer wants to do something these playbooks don't allow, they're expected to write their own or copy and update these.

As such, I am perfectly fine merging this.

I do get the point with with using the subscription-manager role. I see people reimplementing the same thing over and over in every red hat ansible repo. To my knowledge, other people in this repo with the CASL team are working on a more robust subscription manager role with the intention of upstreaming it either in openshift-ansible or maybe even ansible proper.

When that happens, we'll switch to it, but I don't want to block this little ad-hoc example on it.

PR #760, opened 2017-09-26
Fixing logic of conditional

cwilkers:
undefined rhsm_user failed on this line because AND precedes OR

#### What does this PR do?
Fixes logic in a registration so it doesn't fire with an undefined rhsm_user (case when using satellite and activation key instead)

Run role with rhsm_user undefined, use satellite with AK instead.

cc: @dav1x @cooktheryan


cwilkers:
Travis is being a bit of a bully today; It might be some undefined voodoo with this PR.
@dav1x @cooktheryan I'm going to close this and re-request the same change and see where it goes.

PR #768, opened 2017-09-29
all systems need the node package anyways

cooktheryan:
#### What does this PR do?
All systems need the node package anyways so why not install it early

Run reference architecture scripts

N/a

cc: @e-minguez @dav1x


tomcc31415:
This breaks deployments on AWS with deployment type=origin on centos7 hosts with the following error;
 changed": false,
 "failed": true,
 "item": "atomic-openshift-node",
 "rc": 126,
 "results": [
 "No package matching 'atomic-openshift-node' found available, installed or updated"
 ]
 }

Reverting this change allows for a complete deployment again.

cooktheryan:
Good call this is my fault. Will revert

PR #770, opened 2017-10-01
version bump for upgrade plays

cooktheryan:
#### What does this PR do?
Modify default upgrade version of playbook

manually run

N/a

cc: @e-minguez @rlopez133 @dav1x


reference-architecture/aws-ansible/playbooks/roles/cloudformation-infra/files/brownfield-byo-bastion.json.j2
```diff
@@ -1,1266 +0,0 @@
-{
```
e-minguez: Is this file deleted on purpose?

cooktheryan: good catch let me clean that up

PR #771, opened 2017-10-02
Fixing various contrib changes causing CASL breakage

oybed:
#### What does this PR do?
Fixing breakage found in CASL due to various changes in contrib. These are the changes:

heat_stack.yaml.j2:
- `k8s_type` set to new set of facts - changed to have default values for backwards compatibility  
- `image` set to new set of facts - changed to have a default value for backwards compatibility + provide more flexibility 

openstack-stack/defaults/main.yml:
- `provider_network` set to *False* instead of *None* to ensure conditionals are evaluated correctly 

The following PRs introduced these changes - please have the authors review the fixes in this PR:
https://github.com/openshift/openshift-ansible-contrib/pull/643
https://github.com/openshift/openshift-ansible-contrib/pull/637
https://github.com/openshift/openshift-ansible-contrib/pull/701

Ensure CASL provisioning works. 

N/A

cc: @tomassedovic @bogdando @Tlacenka 

@etsauer FYI

### DIFF
--- a/roles/openstack-stack/defaults/main.yml
+++ b/roles/openstack-stack/defaults/main.yml
@@ -18,4 +18,4 @@ dns_volume_size: 1
 lb_volume_size: 5
 use_bastion: False
 ui_ssh_tunnel: False
-provider_network: None
+provider_network: False
--- a/roles/openstack-stack/templates/heat_stack.yaml.j2
+++ b/roles/openstack-stack/templates/heat_stack.yaml.j2
@@ -499,7 +499,7 @@ resources:
               template: k8s_type-%index%.cluster_id
               params:
                 cluster_id: {{ stack_name }}
-                k8s_type: {{ etcd_hostname }}
+                k8s_type: {{ etcd_hostname | default('etcd') }}
           cluster_env: {{ public_dns_domain }}
           cluster_id:  {{ stack_name }}
           group:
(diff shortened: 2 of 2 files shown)


roles/openstack-stack/defaults/main.yml
```diff
@@ -18,4 +18,4 @@ dns_volume_size: 1
 lb_volume_size: 5
 use_bastion: False
 ui_ssh_tunnel: False
-provider_network: None
+provider_network: False
```
tomassedovic: Why is this necessary? Jinja's conditionals treat `None` as falsey. What is this breaking for CASL?

oybed: @tomassedovic I didn't capture the error, but running this with Ansible v2.3.1.x, the error was something like "provider_network 'None' not valid".

oybed: i.e.: error coming from the Ansible modules, not CASL specific per-say

oybed:
@tomassedovic yes, having a CI job for CASL would be great. We have discussed that for our own needs as well, so maybe we can leverage/re-use some of the CI configurations that you already have? If we can find a time to discuss, I'd love to get more details from you. Are you by a chance able to join our CASL call this Friday?

roles/openstack-stack/defaults/main.yml
tomassedovic: Hm, I'm running `2.3.1.0` and had never seen it. Might be a Jinja version or something. Whatever, I'll merge it.

tomassedovic:
@oybed yeah that might work, though our current "CI" only has a capacity for one job, so we'll have to figure out where to run it. I'll try to make it to the call (much appreciate you moving it earlier in the day).

PR #776, opened 2017-10-04
Add dynamic inventory

tomassedovic:
#### What does this PR do?

This adds an `inventory.py` script to the `sample-inventory` that lists
all the necessary servers and groups dynamically, skipping the
`static_inventory` role as well as the `hosts` creation.

It also adds an `os_cinder` lookup function which is necessary for a
seamless Cinder OpenShift registry integration without a static
inventory.

1. Update your `ansible.cfg` with the one in this pull request
2. Copy the inventory from this PRs sample-inventory (or just make sure you have the new `inventory.py` file in your inventory)
3. Set `openstack_inventory: dynamic` in `inventory/group_vars/all.yml`
4. Run the openstack provision playbook
5. Verify that the `inventory/hosts` file is **NOT** created.
6. Verify that the provisioning succeeded
7. Run openshift-ansible to install openshift
8. Verify the installation succeeded

Note: this should also work for the "set up and attach cinder registry" scenario, but that's blocked on this openshift-ansible issue:

https://github.com/openshift/openshift-ansible/issues/5657

Also, this might not work for scenarios with provider network, multiple subnets and multiple stacks in the same project. I would prefer to do them in separate pull requests.

n/a

### DIFF
--- a/lookup_plugins/os_cinder.py
+++ b/lookup_plugins/os_cinder.py
@@ -1,0 +1,30 @@
+from ansible.errors import AnsibleError, AnsibleParserError
+from ansible.plugins.lookup import LookupBase
+
+import shade
+
+
+class LookupModule(LookupBase):
+
+    def run(self, volume_names, variables=None, **kwargs):
+        cloud = shade.openstack_cloud()
+        volume_attributes = [
+            "id",
+            "name",
+            "display_name",
+            "size",
+            "description",
+        ]
+
+        def get_volume(name_or_id):
+            volume = cloud.get_volume(name_or_id)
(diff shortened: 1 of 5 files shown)


tomassedovic:
The latest commit makes provider networks work.

tomassedovic:
So FWIW, I think we can merge this. The end to end CI passed. Would appreciate reviews.

tomassedovic:
FYI I ran a separate e2e build here with the dynamic inventory enabled:

https://travis-ci.org/openshift/openshift-ansible-contrib/builds/283695294

PR #778, opened 2017-10-05
Replace the CASL references

tomassedovic:
#### What does this PR do?
Following up on the initial port of the OpenStack roles from
casl-ansible to openshift-ansible-contrib. One of the points that was
brought up in the review was to drop the references to CASL in the
code since the code has now wider reach.

Standart openstack end to end test. Nothing should break.

n/a

cc: @tzumainn @Tlacenka  @bogdando 

@oybed @etsauer I'd appreciate you guys taking a look -- do you depend on the values we're changing here?

### DIFF
--- a/playbooks/provisioning/openstack/pre_tasks.yml
+++ b/playbooks/provisioning/openstack/pre_tasks.yml
@@ -7,7 +7,7 @@
 
 - name: Set default Environment ID
   set_fact:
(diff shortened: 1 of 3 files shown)


oybed:
@tomassedovic LGTM - we don't rely on these for any of our automation. Thanks for checking in with us.

PR #779, opened 2017-10-05
[WIP]: Add support for kuryr_networking

celebdor:
This patch adds the service and the pod subnets necessary for kuryr as
well as the load balancer for the OpenShift API.

Signed-off-by: Antoni Segura Puimedon <antonisp@celebdor.com>

#### What does this PR do?
Brief explanation of the code or documentation change you've made

Include commands to run your new feature, and also post-run commands to validate that it worked. (please use code blocks to format code samples)


celebdor:
Note that this is not ready for merging. I just want to put it up here and keep working on it

roles/openstack-stack/templates/heat_stack.yaml.j2
```diff
@@ -130,6 +215,33 @@ resources:
...
+        network: { get_resource: service_net}
+        fixed_ips:
+          - subnet: { get_resource: service_subnet }
+            ip_address: {{ service_subnet_gateway }}
+        name:
+          str_replace:
+            template: openshift-ansible-cluster_id-service-subnet-router-port
+            params:
+              cluster_id: {{ stack_name }}
+  interface:
```
tomassedovic: The name here ("interface") conflicts with the "interface" resource above "service_router_port"

roles/openstack-stack/templates/heat_stack.yaml.j2
```diff
@@ -73,7 +73,92 @@ outputs:
...
+  api_lb:
+    type: OS::Neutron::LBaaS::LoadBalancer
+    properties:
+      name:
+        str_replace:
+          template: openshift-ansible-cluster_id-api-lb
+          params:
+            cluster_id: {{ stack_name }}
+      vip_address: {{ api_lbaas_ip }}
+      vip_subnet: { get_resource: service_subnet }
```
tomassedovic: This needs to be in a `use_kuryr` block, too, otherwise it won't find the `service_subnet` I think. Or can we make this work with the non-kuryr deployments as well by using the `subnet` subnet when `use_kuryr` is `false`?

tomassedovic: Also, this uses the service_subnet's IP range. What's the plan for a publicly accessible LB? Will it be a separate resource with a floating IP?

bogdando:
Thanks for opening this. I'd love to get my hands dirty with Kuryr setups by the provider playbooks!

bogdando:
Some things from the https://github.com/openshift/openshift-ansible/pull/5574 may be added for the provider docs as well.

PR #780, opened 2017-10-05
Adjust logic of host registration when: parameter

cwilkers:
Test for rhsm_user is missed due to and/or precedence.

#### What does this PR do?
Adjustment of logic to prevent referencing undefined rhsm_user

1. Ensure rhsm_user undefined
2. Ensure host is not yet registered
3. Run role against host.

cc: @dav1x @cooktheryan @e-minguez  PTAL


cwilkers:
Failed attempts were using the web based editor. Trying from my forked repo instead.

PR #789, opened 2017-10-06
Add openstack_num_etcd to sample inventory

tzumainn:
#### What does this PR do?
Adds openstack_num_etcd to sample inventory, as it seems to be the correct default

Provision using the openstack playbooks; witness the etcd node being created; deploy openshift on top of it with no errors

N/A

### DIFF
--- a/playbooks/provisioning/openstack/sample-inventory/group_vars/all.yml
+++ b/playbooks/provisioning/openstack/sample-inventory/group_vars/all.yml
@@ -36,6 +36,7 @@ openstack_default_image_name: "centos7"
 openstack_num_masters: 1
 openstack_num_infra: 1
 openstack_num_nodes: 2
+openstack_num_etcd: 1
 
 # # Used Flavors
 # # - set specific flavors for roles by uncommenting corresponding lines


tomassedovic:
@tzumainn what's the reasoning behind this? When the etcd nodes are empty, the deployment will use each master node as etcd:

https://github.com/openshift/openshift-ansible-contrib/blob/master/roles/static_inventory/templates/inventory.j2#L63

I'm not necessarily against that, but I'd like to understand the change.

tzumainn:
@tomassedovic I ran into this check with my  config while testing dynamic inventory:

https://github.com/openshift/openshift-ansible/commit/3fc499fe3a1549f994ab13b0ebb5d4bba8c2c9cf

Reading the message more carefully, I guess we don't necessarily need an independent etcd node, but it seems like we need an etcd group, which I don't think we've been defining?

tomassedovic:
@tzumainn weird. The dynamic inventory defines the etcd group, too:

https://github.com/openshift/openshift-ansible-contrib/blob/cc59269a948e632196dea2bfc37ecc9712177157/playbooks/provisioning/openstack/sample-inventory/inventory.py#L53

You can verify it by running `./inventory.py --list`.

How many master nodes do you deploy? Is it either 1 or 3? Because the openshift ansible task fails if the count is different.

Other than that, could you try pulling the latest openshift-ansible commits? Sometimes they break things for us and maybe you got unlucky?

tzumainn:
Hm, it worked this time - odd! Closing this PR, sorry for the noise!

PR #790, opened 2017-10-09
Add CentOS support to the docker-storage-setup role

tomassedovic:
#### What does this PR do?

This let's us use the role on CentOS systems, as well as RHEL. In addition, it
installs docker and makes sure it's restarted (as opposed to just "started"
which has no effect when docker is already running).

These changes will let the OSP provider to use this role instead of our own homemade solution.

Run the docker-storage-setup role on centos and verify that Docker is installed, run and is configured as per the generated docker-storage-setup file.

n/a

The role seems to be used by the vmware and rhv reference architectures, so in addition to @e-minguez  and @cooktheryan I'd appreciate a look from some of those folks to make sure we're not breaking anything.

### DIFF
--- a/roles/docker-storage-setup/tasks/main.yaml
+++ b/roles/docker-storage-setup/tasks/main.yaml
@@ -25,5 +25,21 @@
     - ansible_distribution_version | version_compare('7.4', '<')
     - ansible_distribution == "RedHat"
 
+- block:
+    - name: create the docker-storage-setup config file for CentOS
+      template:
+        src: "{{ role_path }}/templates/docker-storage-setup-dm.j2"
+        dest: /etc/sysconfig/docker-storage-setup
+        owner: root
+        group: root
+        mode: 0644
+
+  # TODO(shadower): Find out which CentOS version supports overlayfs2
+  when:
+    - ansible_distribution == "CentOS"
+
+- name: Install Docker
+  package: name=docker state=present
+
 - name: start docker
-  service: name=docker state=started enabled=true
+  service: name=docker state=restarted enabled=true


tomassedovic:
@cooktheryan et al: I was wondering whether we would have this role just generate the docker-storage-setup file and remove packaging and service setup (and leave that on additional playbooks/roles.

e-minguez:
A couple of things:

* It seems like "container-storage-setup" is the preferred method for >7.4 versions[1][2] (I think to be able to replace docker with other container technologies)
* Starting docker service is not required to create the proper configuration files, just calling `docker-storage-setup` (or `container-storage-setup`) is enough (to decouple it from docker installation as @tomassedovic said)
* `container-storage-setup` is provided by the `container-storage-setup` package[4] and should be executed before starting the `docker` service
* `docker-storage-setup` is provided by the `docker` package[4] and should be executed before starting the `docker` service[4], so if using `docker-storage-setup` it is required the `docker` package to be installed.

My 2 cents.

[1] https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux_atomic_host/7/html/managing_containers/managing_storage_with_docker_formatted_containers
[2] https://github.com/projectatomic/container-storage-setup
[3] 
```
[edu@edu-rhel7 ~]$ which container-storage-setup
/usr/bin/container-storage-setup
[edu@edu-rhel7 ~]$ rpm -qf /usr/bin/container-storage-setup 
container-storage-setup-0.6.0-1.gite67c964.el7.noarch
```

[4]
```
[edu@edu-rhel7 ~]$ which docker-storage-setup 
/usr/bin/docker-storage-setup
[edu@edu-rhel7 ~]$ rpm -qf /usr/bin/docker-storage-setup
docker-1.12.6-55.gitc4618fb.el7.x86_64

```

tomassedovic:
Thanks @e-minguez! I didn't know about container-storage-setup.

In my (limited) testing, Docker seems to run docker-storage-setup when it starts (maybe just on the first start). I didn't dig into it, but it looks like running the docker-storage-setup binary isn't strictly-speaking necessary. But if docker-storage-setup fails, Docker will start without it -- if we ran it manually, we'd get some errors back.

e-minguez:
I think `docker` service will start `docker-storage-setup` as it is a requirement:

```
[root@edu-rhel7 ~]# grep Wants /usr/lib/systemd/system/docker.service
Wants=docker-storage-setup.service
```

Running `docker-storage-setup` doesn't do anything if the docker storage has been configured and according to the official documentation, running `docker-storage-setup` should be done previous to start the `docker.service`[1]

So, from my point of view, I think we should stick to just `docker-storage-setup` tasks, including:

* Check RHEL/CentOS version
* If >= 7.4, install container-storage-setup, apply the overlay2 template and run the container-storage-setup command
* If <7.4, install docker, apply the dm template and run the docker-storage-setup command
* Leave the docker service startup to another role if we want, or restart & enable it here


[1] https://docs.openshift.com/container-platform/3.6/install_config/install/host_preparation.html#configuring-docker-storage

roles/docker-storage-setup/tasks/main.yaml
```diff
@@ -25,5 +25,21 @@
...
+- block:
+    - name: create the docker-storage-setup config file for CentOS
+      template:
+        src: "{{ role_path }}/templates/docker-storage-setup-dm.j2"
+        dest: /etc/sysconfig/docker-storage-setup
+        owner: root
+        group: root
+        mode: 0644
+
+  # TODO(shadower): Find out which CentOS version supports overlayfs2
```
bogdando: FWIW, overlay2 works with the centos 7.3 in tripleo upstream CI, with the latest kernel installed

bogdando:
@e-minguez thank you for valuable configuration bits for docker in >=7.4! 
I believe the purpose of this initial step is only reusing the openshift-ansible roles as is without additional required modifications. We may want to add the >=7.4 support in the follow up patches. @tomassedovic , thoughts? If we have to adjust something for the <7.4, would be nice to make it openshift-ansible side though, as a dependency patch.

roles/docker-storage-setup/tasks/main.yaml
e-minguez: For RHEL I'd say to be careful about supportability starting with 7.4[1]

[1] https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux_atomic_host/7/html/managing_containers/managing_storage_with_docker_formatted_containers#overlay_graph_driver

e-minguez:
Adding @dav1x and @cwilkers as they use this for their VMware and RHV ref. archs.

PR #794, opened 2017-10-11
Add Extra Docker Registry URLs (custom post-provision action)

Tlacenka:
#### What does this PR do?
This is another example playbook added to the custom-actions directory and documented in the provisioning READMe.

1. Run the provisioning.
2. Connect to an arbitrary host in your cluster and display the content of the `/etc/containers/registries.conf` file. Optionally, add a line with `registries: something` or a list of random items.
3. Run the add-docker-registry.yml playbook like so:
```
ansible-playbook -i <inventory> openshift-ansible-contrib/playbooks/provisioning/openstack/custom-actions/add-docker-registry.yml  --extra-vars '{"registries": "reg1", "insecure_registries": ["ins_reg1","ins_reg2"]}'
```
4. Check that the playbook finished successfully.
5. Connect to an arbitrary host in your cluster and display the content of the `/etc/containers/registries.conf` file again. Check that it includes the newly added registries as well as the old ones (if there were any).
6. Run `docker info` as root and check that these registries were added (under `Registries` and `Insecure Registries`).

cc: @tomassedovic @bogdando @tzumainn PTAL


playbooks/provisioning/openstack/custom-actions/add-docker-registry.yml
```diff
@@ -0,0 +1,92 @@
...
+        registries_list: [ "{{ registries }}" ]
+
+    - name: Make sure that insecure_registries is a list
+      when: 'insecure_registries is string'
+      set_fact:
+        insecure_registries_list: [ "{{ insecure_registries }}" ]
+
+    - name: Set docker file content to a unified format
+      set_fact:
+        docker_conf_registries: "{{ [] if docker_conf['registries'] is not defined else [ docker_conf['registries'] ] if docker_conf['registries'] is string else docker_conf['registries'] }}"
```
tomassedovic: Have you seen `docker_conf['registries']` not being an array? When I checked the format & the source code, it seems to always expect an array.

In which case, please remove the check so the code is easier to read and understand.

playbooks/provisioning/openstack/custom-actions/add-docker-registry.yml
```diff
@@ -0,0 +1,92 @@
...
+      when: 'insecure_registries_merge_result|skipped'
+      set_fact:
+        docker_conf: "{{ docker_conf | combine({'insecure_registries': (docker_conf_insecure_registries + insecure_registries_list) | unique }, recursive=True) }}"
+
+    - name: Load variable back to file
+      become: true
+      copy:
+        content: "{{ docker_conf | to_yaml }}"
+        dest: /etc/containers/registries.conf
+
```
tomassedovic: We must also restart the `registries` service (that comes from `atomic-registries`) for the configuration to take effect. Please add this task before restarting Docker:

```yaml
service:
    name: registries
    state: restarted
```

tomassedovic:
@Tlacenka to answer your questions in the pull request:

1. Run `docker info` -- it will print out the current registries towards the end
2. `[1, 2, 3]` is a valid YAML syntax and it's equivalent to:

```yaml
- 1
- 2
- 3
```

I recommend skimming the YAML's wikipedia page:

https://en.wikipedia.org/wiki/YAML

it's quite a complex format. Reading the full spec is only for the brave of heart :-)

PR #810, opened 2017-10-13
Updating the issue template to be more consistent in format.

mbruzek:
#### What does this PR do?
The issue template had different formatting and was hard to work with so I humbly suggest this fix.

Make sure it renders from Markdown to HTML correctly, and you agree with the changes.

No there is no issue, I just found the previous template to be hard to work with.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,6 +1,6 @@
 <!-- Thanks for filing an issue! Before hitting the button, please answer these questions.-->
 
-Is this a **[BUG REPORT](#bug-request-info)** or a **[FEATURE REQUEST](#feature-request-info)** ?:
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE.md
```diff
@@ -10,83 +10,88 @@ If this is a BUG REPORT, please:
...
+    a pull request with a proposed implementation.
 
 In both cases, be ready for followup questions, and please respond in a timely
 manner.  If we can't reproduce a bug or think a feature already exists, we
 might close your issue.  If we're wrong, PLEASE feel free to reopen it and
 explain why.
 -->
 
-# BUG REQUEST INFO:
+# BUG REQUEST INFO
```
bogdando: once you're fixing this line, could you please fix the "bug request" to become "but report"? :)

mbruzek:
@bogdando I updated the header based on your request. Good find, I enjoy being consistent.

PR #814, opened 2017-10-17
Add Flannel support

bogdando:
#### What does this PR do?

* Document flannel options. See the [updated ref arch](https://access.redhat.com/documentation/en-us/reference_architectures/2017/html-single/deploying_and_managing_red_hat_openshift_container_platform_3.6_on_red_hat_openstack_platform_10/#flannel_considerations) for more details.
* Provision a dedicated isolated neutron subnet for the flannel network and disable ports security on it.
* Add openshift-ansible as a galaxy dependency module.
* Create a post-install playbook for Flannel SDN post deployment configs
* Use the os_firewall role to configure iptables and apply rules required for Flannel SDN to function.
* Configure and enable the masked iptables service on the app nodes for Flannel SDN. Keep it as is, disabled and masked, for other nodes groups.

Uncomment flannel options in OSEv3 group vars.
Provision an env, deploy from `openshift-ansible-3.7.0-0.127.0-69-gc16d16d` or later (it must contain the flannel fixes `98db8ec` and `fdc61b8`), and run e2e, it should pass.
Verify that iptables rules added by post-install playbook are persistent across node reboots and iptables services restarts.

cc: @tomassedovic @e-minguez  PTAL

### DIFF
--- a/playbooks/provisioning/openstack/galaxy-requirements.yaml
+++ b/playbooks/provisioning/openstack/galaxy-requirements.yaml
@@ -4,3 +4,7 @@
 # From 'infra-ansible'
 - src: https://github.com/redhat-cop/infra-ansible
   version: master
+
+# From 'openshift-ansible'
+- src: https://github.com/openshift/openshift-ansible
+  version: master
--- a/playbooks/provisioning/openstack/post-install.yml
+++ b/playbooks/provisioning/openstack/post-install.yml
@@ -3,16 +3,55 @@
   gather_facts: False
   become: True
   tasks:
+    - name: Save iptables rules to a backup file
+      when: openshift_use_flannel|default(False)|bool
+      shell: iptables-save > /etc/sysconfig/iptables.orig-$(date +%Y%m%d%H%M%S)
+
+# Enable iptables service on app nodes to persist custom rules (flannel SDN)
+# FIXME(bogdando) w/a https://bugzilla.redhat.com/show_bug.cgi?id=1490820
+- hosts: app
+  gather_facts: False
+  become: True
+  vars:
+    os_firewall_allow:
+      - service: dnsmasq tcp
+        port: 53/tcp
+      - service: dnsmasq udp
+        port: 53/udp
+  tasks:
+    - when: openshift_use_flannel|default(False)|bool
+      block:
+        - include_role:
+            name: openshift-ansible/roles/os_firewall
+        - include_role:
+            name: openshift-ansible/roles/lib_os_firewall
+        - name: set allow rules for dnsmasq
+          os_firewall_manage_iptables:
+            name: "{{ item.service }}"
+            action: add
+            protocol: "{{ item.port.split('/')[1] }}"
+            port: "{{ item.port.split('/')[0] }}"
+          with_items: "{{ os_firewall_allow }}"
+
+- hosts: OSEv3
+  gather_facts: False
+  become: True
+  tasks:
     - name: Apply post-install iptables hacks for Flannel SDN (the best effort)
       when: openshift_use_flannel|default(False)|bool
       block:
-        - name: Save iptables rules to a backup file
-          shell: iptables-save > /etc/sysconfig/iptables.orig-$(date +%Y%m%d%H%M%S)
+        - name: set allow/masquerade rules for for flannel/docker
+          shell: >-
+            (iptables-save | grep -q custom-flannel-docker-1) ||
+            iptables -A DOCKER -w
+            -p all -j ACCEPT
+            -m comment --comment "custom-flannel-docker-1";
+            (iptables-save | grep -q custom-flannel-docker-2) ||
+            iptables -t nat -A POSTROUTING -w
+            -o {{flannel_interface|default('eth1')}}
+            -m comment --comment "custom-flannel-docker-2"
+            -j MASQUERADE
 
-        - name: Add iptables rules for flannel SDN and docker
-          shell: >
-            iptables -I DOCKER 1 -w -p all -j ACCEPT;
-            iptables -t nat -I POSTROUTING 1 -w -o {{flannel_interface|default('eth1')}} -j MASQUERADE
-
-        - name: Persist modified iptables rules
-          shell: iptables-save > /etc/sysconfig/iptables
+        # NOTE(bogdando) the rules will not be restored, when iptables service unit is disabled & masked
+        - name: Persist in-memory iptables rules (w/o dynamic KUBE rules)
+          shell: iptables-save | grep -v KUBE > /etc/sysconfig/iptables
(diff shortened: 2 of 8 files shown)


playbooks/provisioning/openstack/galaxy-requirements.yaml
```diff
@@ -4,3 +4,7 @@
 # From 'infra-ansible'
 - src: https://github.com/redhat-cop/infra-ansible
   version: master
+
+# From 'openshift-ansible'
+- src: https://github.com/openshift/openshift-ansible
+  version: master
```
e-minguez: This is ok for CentOS and/or Origin, but for RHEL + OCP, the openshift-ansible repo shouldn't be used, but the RPM package as there can be differences (specially if you use master branch)

playbooks/provisioning/openstack/post-install.yml
```diff
@@ -0,0 +1,57 @@
...
+- hosts: OSEv3
+  gather_facts: False
+  become: True
+  tasks:
+    - name: Apply post-install iptables hacks for Flannel SDN (the best effort)
+      when: openshift_use_flannel|default(False)|bool
+      block:
+        - name: set allow/masquerade rules for for flannel/docker
+          shell: >-
+            (iptables-save | grep -q custom-flannel-docker-1) ||
```
e-minguez: I'd say to use a specific comment in the iptables rule as "flannel-ocp-allow-docker-traffic" instead a generic one

playbooks/provisioning/openstack/post-install.yml
```diff
@@ -0,0 +1,57 @@
...
+            iptables -A DOCKER -w
+            -p all -j ACCEPT
+            -m comment --comment "custom-flannel-docker-1";
+            (iptables-save | grep -q custom-flannel-docker-2) ||
```
e-minguez: Same stuff, "flannel-ocp-masquerade"

playbooks/provisioning/openstack/post-install.yml
```diff
@@ -0,0 +1,57 @@
...
+  become: True
+  tasks:
+    - name: Save iptables rules to a backup file
+      when: openshift_use_flannel|default(False)|bool
+      shell: iptables-save > /etc/sysconfig/iptables.orig-$(date +%Y%m%d%H%M%S)
+
+# Enable iptables service on app nodes to persist custom rules (flannel SDN)
+# FIXME(bogdando) w/a https://bugzilla.redhat.com/show_bug.cgi?id=1490820
+- hosts: app
+  gather_facts: False
```
tomassedovic: `gather_facts` here should be set to `True`.

The playbook is run manually after openshift-ansible and if the facts cache gets cleared in the meantime, `os_firewall` will fail with:

```
TASK [os_firewall : Install iptables packages] ************************************************************************************************************************
Wednesday 18 October 2017  10:25:46 +0000 (0:00:00.034)       0:00:01.264 ***** 
failed: [app-node-0.openshift.example.com] (item=iptables) => {"failed": true, "item": "iptables", "msg": "Could not detect which package manager to use. Try gathering
 facts or setting the \"use\" option."}
failed: [app-node-0.openshift.example.com] (item=iptables-services) => {"failed": true, "item": "iptables-services", "msg": "Could not detect which package manager to 
use. Try gathering facts or setting the \"use\" option."}
```

playbooks/provisioning/openstack/sample-inventory/group_vars/OSEv3.yml
```diff
@@ -53,5 +53,7 @@ openshift_override_hostname_check: true
 ansible_become: true
 
 # # Flannel networking
+#osm_cluster_network_cidr: 10.128.0.0/14
 #openshift_use_openshift_sdn: false
 #openshift_use_flannel: true
```
tomassedovic: Currently, `openshift_use_flannel: true` must be present in `group_vars/all.yml` as well, otherwise the provisioning will fail.

The reason for this is that some of the prerequisite checks (and heat templates) are run on `localhost` as opposed to the openshift nodes. Which means they won't see `openshift_use_flannel` defined and the provisioning will fail.

I know it's not added in this patch, but it's something we should fix in a subsequent commit.

playbooks/provisioning/openstack/galaxy-requirements.yaml
tomassedovic: @e-minguez we will merge with openshift-ansible in the coming months. Do you think it's still worth adding the logic here in the meantime? Because once we're done, we'll get access to all openshift-ansible roles by default.

bogdando: Thanks, good point. I hope we can leave this as we gonna move this to the openshift-ansible repo soon.

playbooks/provisioning/openstack/post-install.yml
bogdando: I'll address these in a follow-up pr, thanks!

playbooks/provisioning/openstack/post-install.yml
bogdando: nice catch, will follow-up

playbooks/provisioning/openstack/sample-inventory/group_vars/OSEv3.yml
bogdando: hm, it has defaults for the heat templates

roles/openstack-stack/templates/heat_stack.yaml.j2
```diff
@@ -341,6 +341,12 @@ resources:
           protocol: tcp
           port_range_min: 9090
           port_range_max: 9090
+{% if openshift_use_flannel|default(False)|bool %}
```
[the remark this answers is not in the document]
bogdando: this is a ref arch requirement, translated the sec group rule as is

playbooks/provisioning/openstack/galaxy-requirements.yaml
e-minguez: ok to me :) (I didn't know about the merge!)

playbooks/provisioning/openstack/sample-inventory/group_vars/OSEv3.yml
tomassedovic: What I mean is: the default is false. But when you only set `openshift_use_flannel: true` in `OSEv3.yml` and then set `openstack_private_data_network_name` in `all.yml` (which according to the sample inventory and documentation should be enough), you'll get a failure here:

https://github.com/openshift/openshift-ansible-contrib/blob/master/playbooks/provisioning/openstack/net_vars_check.yaml#L9-L14

The reason is because the net_vars_check run on localhost, which is not in the OSEv3 group, so it thinks you have `openshift_use_flannel` unset and will use the `False` default.

bogdando:
@e-minguez 

> Did you tried it and worked after rebooting the hosts both DNS resolution and reachability?

The last time I did, it was fine. Although my testing layout has a single app/minion node only.

PR #817, opened 2017-10-18
Set a consistent API/UI endpoint

tomassedovic:
#### What does this PR do?
Instead of it being the first master hostname in the single-master
scenario or the first load balancer mostname in multi-master, the
sample cluster will always have the following entrypoint by default:

https://console.{{env_id}}.{{public_dns_domain}}:8443

That will simplify the documentation and be easier to look up and
understand for the newcomers trying it out.

The end-to-end CI will exercise this feature on a single master.

There needs to be a manual test on for multiple masters:

1. Copy the sample inventory
2. Set `openstack_num_masters: 3` in `inventory/group_vars/all.yml`
3. Run provisioning
4. Verify that the DNS request for `console.openshift.example.com` returns the floating IP address of the load balancer

n/a

### DIFF
--- a/playbooks/provisioning/openstack/sample-inventory/group_vars/OSEv3.yml
+++ b/playbooks/provisioning/openstack/sample-inventory/group_vars/OSEv3.yml
@@ -5,8 +5,8 @@ openshift_deployment_type: origin
 openshift_master_default_subdomain: "apps.{{ env_id }}.{{ public_dns_domain }}"
 
 openshift_master_cluster_method: native
-openshift_master_cluster_hostname: "{{ groups.lb.0|default(groups.masters.0) }}"
-openshift_master_cluster_public_hostname: "{{ groups.lb.0|default(groups.masters.0) }}"
+openshift_master_cluster_public_hostname: "console.{{ env_id }}.{{ public_dns_domain }}"
+openshift_master_cluster_hostname: "{{ openshift_master_cluster_public_hostname }}"
 
 osm_default_node_selector: 'region=primary'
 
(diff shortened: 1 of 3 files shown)


playbooks/provisioning/openstack/sample-inventory/group_vars/OSEv3.yml
```diff
@@ -5,8 +5,8 @@ openshift_deployment_type: origin
 openshift_master_default_subdomain: "apps.{{ env_id }}.{{ public_dns_domain }}"
 
 openshift_master_cluster_method: native
-openshift_master_cluster_hostname: "{{ groups.lb.0|default(groups.masters.0) }}"
-openshift_master_cluster_public_hostname: "{{ groups.lb.0|default(groups.masters.0) }}"
+openshift_master_cluster_public_hostname: "console.{{ env_id }}.{{ public_dns_domain }}"
+openshift_master_cluster_hostname: "{{ openshift_master_cluster_public_hostname }}"
```
bogdando: this requires additional testing with multi-master setup (not in CI yet)

tomassedovic:
Oh ffs, the job succeeded but travis killed it nonetheless. /me shakes fist at travis

tomassedovic:
@bogdando it's set based on `openshift_master_cluster_public_hostname` by one of these tasks:

https://github.com/openshift/openshift-ansible-contrib/blob/master/roles/dns-records/tasks/main.yml#L67-L88

playbooks/provisioning/openstack/sample-inventory/group_vars/OSEv3.yml
tomassedovic: ah good point, I'll update the PR text

Issue #453, opened 2017-06-09
GCE Multizone Deployment doesn't properly honor PV requirement when scheduling


bdurrow:
I apologize that I don't understand this problem well enough to succinctly describe it.

I spun up a cluster using master@d10ce131203d6add4baa3d156e9563095ac4703d with the patches necessary to deploying using CentOS/Origin this included the following lines in config.yaml:

```
[brad@ansible gce-cli]$ fgrep metric config.yaml
openshift_hosted_metrics_deploy: true
openshift_hosted_metrics_storage_volume_size: 20Gi
```

After the cluster is built there is a pv bound to a pvc like this:
```
Brads-MBP13:~ brad$ oc get pvc
NAME                  STATUS    VOLUME                                     CAPACITY   ACCESSMODES   AGE
metrics-cassandra-1   Bound     pvc-bfeef7c3-4cdf-11e7-8a2f-42010a80000a   20Gi       RWO           10h
```

I also made these changes after the cluster came up:
```
::::::::::::::
/tmp/gce-pd-storageclass-ssd.yaml
::::::::::::::
kind: StorageClass
apiVersion: storage.k8s.io/v1beta1
metadata:
  name: ssd
provisioner: kubernetes.io/gce-pd
parameters:
  type: pd-ssd
::::::::::::::
/tmp/gce-pd-storageclass-standard.yaml
::::::::::::::
kind: StorageClass
apiVersion: storage.k8s.io/v1beta1
metadata:
  name: standard
  annotations:
    storageclass.kubernetes.io/is-default-class: "true"
provisioner: kubernetes.io/gce-pd
parameters:
  type: pd-standard

 sudo /usr/local/bin/oc create -f /tmp/gce-pd-storageclass-standard.yaml
 sudo /usr/local/bin/oc create -f /tmp/gce-pd-storageclass-ssd.yaml  
```

Then I started troublehsooting the problem described in #446.  I changed the image part of the replication controller config and deleted the existing POD so it would get redeployed.  The scheduler assigned it to a node in the wrong zone.  Initially I thought the volume wasn't labeled properly but that doesn't appear to be the case:
```
Brads-MBP13:~ brad$ oc describe pv pvc-bfeef7c3-4cdf-11e7-8a2f-42010a80000a
Name:		pvc-bfeef7c3-4cdf-11e7-8a2f-42010a80000a
Labels:		failure-domain.beta.kubernetes.io/region=us-central1
[...]

PR #820, opened 2017-10-19
Enable docker storage check

pschiffe:
#### What does this PR do?
Enable docker storage check, fix oc annotate task on containerized deployment.

Verify that deployment succeed.

cc: @e-minguez PTAL

### DIFF
--- a/reference-architecture/gcp/ansible/playbooks/openshift-post.yaml
+++ b/reference-architecture/gcp/ansible/playbooks/openshift-post.yaml
@@ -11,7 +11,7 @@
       type: pd-ssd
   tasks:
   - name: set node selector for openshift-infra namespace
-    command: oc annotate --overwrite namespace openshift-infra openshift.io/node-selector='role=infra'
+    shell: PATH="${PATH}:/usr/local/bin:/usr/local/sbin"; oc annotate --overwrite namespace openshift-infra openshift.io/node-selector='role=infra'
 
 - name: post ocp deploy tasks for master nodes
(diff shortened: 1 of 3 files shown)


reference-architecture/gcp/ansible/playbooks/openshift-post.yaml
```diff
@@ -11,7 +11,7 @@
       type: pd-ssd
   tasks:
   - name: set node selector for openshift-infra namespace
-    command: oc annotate --overwrite namespace openshift-infra openshift.io/node-selector='role=infra'
+    shell: PATH="${PATH}:/usr/local/bin:/usr/local/sbin"; oc annotate --overwrite namespace openshift-infra openshift.io/node-selector='role=infra'
```
e-minguez: This is ok, but the "standard way" is to set individual components node selectors as:
* openshift_metrics_cassandra_nodeselector
* openshift_metrics_hawkular_nodeselector
* openshift_metrics_heapster_nodeselector

pschiffe: Sure, I'm doing that here [1]. The namespace selector is useful when you need to do some manual tweaks later. See the discussion about this here [2]. Thanks for the review.

[1] https://github.com/openshift/openshift-ansible-contrib/blob/master/reference-architecture/gcp/ansible/playbooks/openshift-installer-common-vars.yaml#L26
[2] https://github.com/openshift/openshift-ansible-contrib/issues/453

PR #822, opened 2017-10-20
Add soft-teardown option, which helps to save some time during redeployment

pschiffe:
#### What does this PR do?
Add soft-teardown option, which helps to save some time during redeployment

cc: @e-minguez PTAL


pschiffe:
Basically only network and certificates will stay with soft-teardown, so it shouldn't cost anything.

PR #832, opened 2017-10-30
Add in openshift storage setup to ocp reference arch

HackToHell:
When the playbook for GCP installation runs, the node storage setup is not done. This causes the playbook to fail.

The playbook has been tested against the 3.6 commit tag and is able to setup a cluster.

cc: @tomassedovic @bogdando

### DIFF
--- a/reference-architecture/gcp/ansible/playbooks/openshift-post.yaml
+++ b/reference-architecture/gcp/ansible/playbooks/openshift-post.yaml
@@ -1,5 +1,6 @@
 ---
 - include: ../../../../playbooks/empty-dir-quota.yaml
+- include: ../../../../playbooks/openshift-storage.yaml
 
 - name: post ocp deploy tasks for single master node
   hosts: single_master


pschiffe:
In GCP ref. arch. we are doing storage setup with cloud-init here [1]. What error did you see?

[1] https://github.com/openshift/openshift-ansible-contrib/blob/master/reference-architecture/gcp/deployment-manager/core.jinja#L149

HackToHell:
@pschiffe The /dev/sdc volume was not mounted and was causing the origin-node service to fail, there were no entries in `/etc/fstab` either. Looking through the node now, I see `/dev/disk/by-id/google-openshift` symlinked to ../../sdc properly. Will investigate this further day after tomorrow.

HackToHell:
I am not able to replicate the issue, tried it two times, not sure why it happened in the first place. In the newer cluster the fstab file contains the proper references. Will reopen if I run into an issue with proper logs etc.

PR #836, opened 2017-10-30
Document python-passlib dependency

asgeirf:
Metrics deployer in 3.6 requires python-passlib to install.

#### What does this PR do?
Documents the need to install python-passlib on the control machine.

Without this dependency, `ose-on-aws.py` fails with
```
TASK [openshift_metrics : Check that python-passlib is available on the control host] ****************************************************************************************************************
fatal: [ose-master01.interaxo-dev.com]: FAILED! => {
    "assertion": "'not installed' not in passlib_result.stdout",
    "changed": false,
    "evaluated_to": false,
    "failed": true,
    "msg": "python-passlib rpm must be installed on control host"
}
```

### DIFF
--- a/reference-architecture/aws-ansible/README.md
+++ b/reference-architecture/aws-ansible/README.md
@@ -27,7 +27,8 @@ $ yum -y install python2-boto \
                  python-click \
                  python-httplib2 \
                  python-passlib \
-                 httpd-tools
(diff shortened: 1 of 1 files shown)


cooktheryan:
@dav1x maybe PR this as well to your setup scripts

PR #839, opened 2017-11-02
Docker ss atomicfix

cwilkers:
#### What does this PR do?
- Handle case where docker has already been started and exists (by removing docker-storage sysconfig)
- Handle case where role runs on Atomic Host, (and cannot package install)

Test docker setup on 7.3, 7.4, CentOS, and RHEL Atomic Host

Found in testing on Atomic

cc: @dav1x @e-minguez @pschiffe

### DIFF
--- a/roles/docker-storage-setup/tasks/main.yaml
+++ b/roles/docker-storage-setup/tasks/main.yaml
@@ -3,7 +3,7 @@
   service: name=docker state=stopped
 
 - block:
-    - name: create the docker-storage config file
+    - name: create the docker-storage-setup overlayfs config file
       template:
         src: "{{ role_path }}/templates/docker-storage-setup-overlayfs.j2"
         dest: /etc/sysconfig/docker-storage-setup
@@ -15,7 +15,7 @@
     - ansible_distribution == "RedHat"
 
 - block:
-    - name: create the docker-storage-setup config file
+    - name: create the docker-storage-setup devicemapper config file
       template:
         src: "{{ role_path }}/templates/docker-storage-setup-dm.j2"
         dest: /etc/sysconfig/docker-storage-setup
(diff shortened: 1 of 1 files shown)


cwilkers:
[(3) Adjust docker-storage-setup role for Atomic [epic-rhv]](https://trello.com/c/B2NDOBbf/1102-3-adjust-docker-storage-setup-role-for-atomic-epic-rhv)

e-minguez:
See this https://github.com/openshift/openshift-ansible/blob/master/roles/contiv_facts/tasks/main.yml#L1-L15 for check if it is atomic

cwilkers:
@e-minguez Do you think I need to switch over to the `stat` test for the `ostree`? I think for the purpose of this role, simply excepting the installation of `docker` if the package manager is `unknown` should be fine, and will not have any unintended side effects.

roles/docker-storage-setup/tasks/main.yaml
```diff
@@ -41,6 +46,7 @@
 
 - name: Install Docker
   package: name=docker state=present
+  when: ansible_pkg_mgr != "unknown"
```
cooktheryan: https://github.com/openshift/openshift-ansible-contrib/blob/master/roles/prerequisites/tasks/main.yaml

Check this. When using the library it will generate the facts to automatically know if a host is atomic or not

cwilkers: Ah cool. I'd assumed you had to go into the BYO playbook before it would set that var. Will update...

cwilkers:
I'm adding a bit of extra storage related functionality to the role, Atomic hosts come with a `docker-pool` lv already taken out of the root (`atomicos`) lv for the `devicemapper` config. To use `overlay2` and give both docker and the /var/ partition more space, we need to override this and grow the root lv.

Extra review once I'm done play testing. @e-minguez @cooktheryan

e-minguez:
@cwilkers check travis issues and I'll merge it.
Thanks!

cwilkers:
@e-minguez It looks like Travis failed its environment build before it got to checking the code. Local detox runs return success. (with exception of python 3.5 which I haven't installed.)

e-minguez:
Could you try to perform a "fake" commit to trigger travis build again? Thx.

cwilkers:
@e-minguez @dav1x Looks like Travis is happier now.

PR #845, opened 2017-11-09
[WIP] Allow different public and private DNS hostnames

bogdando:
#### What does this PR do?
Allow different public and private DNS hostnames.
Do not manage hostnames with openstack provider any more.
Instead, allow the cloud-init to do its job.
Ignore the hostnames for provisioned VMs. Rely on openshift-ansible
inventory variables to configure openshift clusters.

When a private FQDN differs from the public FQDN,
define 'private_fqdn' metadata for provisioned with Heat Nova servers.
Place 'openshift_(public)_ip/hostname' vars into a static and
dynamic inventory depending on the public/private IP and FQDNs setup.

* e2e with defaults should pass
* e2e with arbitrary hostnames of provisioned VMs should pass.

* Provision a cluster with `-e openshift_openstack_private_hostname_suffix: openshift-priv` and `-e openshift_openstack_private_dns_domain: example.local`.
  * The stack name, nova servers names, ansible inventory hostnames should keep using the public name prefix (defaults to cluster id, 'openshift') and belong to the public DNS domain
  * nslookup by short name executed on nodes should resolve with the public FQDN "autocompleted"
  (that depends on the hosts' search domain setup in /etc/resolv.conf, so this one is allowed to fail)
  * nslookup by public FQDNs should resolve (with a floating IP provided) via external queries against the public DNS server
  * nslookup by private FQDNs should NOT resolve via external queries against the public DNS server
  * nslookup by private FQDNs executed on nodes should resolve via internal DNS server
  * nslookup by public FQDNs executed on nodes should resolve via internal DNS server, yet with a private IP, not the floating IP

* Deploy a cluster with `-e openshift_openstack_private_hostname_suffix: openshift-priv` and `-e openshift_openstack_private_dns_domain: example.local`. E2e should pass with default and arbitrary hostnames of provisioned VMs.

* Provision using the same custom vars as above, but with an external DNS server deployed by [osp 10 ref arch](https://github.com/openshift/openshift-ansible-contrib/tree/master/reference-architecture/osp-dns) and multi-master setup. Expectations are the same as above.

Maintain the change compatible with
https://github.com/openshift/openshift-ansible/pull/6039

cc: @tomassedovic @oybed PTAL


bogdando:
The master-api service cannot start as it lookups master-0 via its public FQDN, yet via the private network interface. Perhaps, the hosts name should not change to the updated private DNS and suffix?
@tomassedovic need your input please.

bogdando:
Okay, if I keep the hostnames as is (public based FQDNS), the test env gives me the following results for DNS queries against private & public FQDNs:

```
[openshift@infra-node-0 ~]$ getent ahostsv4 infra-node-0.openshift.cluster.lc
192.168.99.6    STREAM infra-node-0.openshift.cluster.lc
192.168.99.6    DGRAM  
192.168.99.6    RAW    
172.17.0.1      STREAM 
172.17.0.1      DGRAM  
172.17.0.1      RAW    
[openshift@infra-node-0 ~]$ getent ahostsv4 infra-node-0.openshift-priv.example.local
192.168.99.6    STREAM infra-node-0.openshift-priv.example.local
192.168.99.6    DGRAM  
192.168.99.6    RAW 
```
This looks much better now, I hope that's the expected result, right?

bogdando:
It works for me now! PTAL folks

bogdando:
@tomassedovic If that's good enough as is, I'd keep it. Although  we could try to adjust `openshift_(public)_hostname` as well, f.e. like in this comment https://github.com/openshift/openshift-ansible/pull/5883#issuecomment-343508093 described for AWS case. Perhaps we could lift off the hack with cloud-init prevented from updating hosts name as a free bonus!

roles/dns-records/tasks/main.yml
```diff
@@ -28,30 +28,49 @@
...
+        key_name: "{{ nsupdate_private_key_name|default('private-' + openshift_openstack_full_private_dns_domain) }}"
         key_secret: "{{ nsupdate_key_secret_private }}"
         key_algorithm: "{{ nsupdate_key_algorithm_private | lower }}"
         entries: "{{ private_records }}"
 
+- name: "Generate the private for public Add section for DNS"
+  set_fact:
+    private_named_records:
+      - view: "private"
+        zone: "{{ openshift_openstack_full_public_dns_domain }}"
```
tomassedovic: Since this is using the public zone, the DNS servers never receive the entries with the private prefixes.

I think this task should say "private" everywhere here.

playbooks/provisioning/openstack/sample-inventory/group_vars/all.yml
```diff
@@ -3,6 +3,20 @@ env_id: "openshift"
...
+# https://github.com/openshift/openshift-ansible/pull/6039 namespaces
+# remove it once resubmitted by the new location in openshift-ansible
+openshift_openstack_clusterid: "{{ env_id }}"
+openshift_openstack_public_dns_domain: "{{ public_dns_domain }}"
+
+# Allows to override the suffix provided in openshift_openstack_clusterid
+# and the private DNS domain.
+openshift_openstack_public_hostname_suffix: "{{ openshift_openstack_clusterid }}"
+openshift_openstack_private_hostname_suffix: "{{ openshift_openstack_clusterid }}"
+openshift_openstack_private_dns_domain: "{{ openshift_openstack_public_dns_domain }}"
```
tomassedovic: Can we set a default value so this doesn't break existing users?

roles/openstack-stack/templates/heat_stack.yaml.j2
```diff
@@ -521,6 +521,9 @@ resources:
               params:
                 cluster_id: {{ stack_name }}
                 k8s_type: {{ etcd_hostname | default('etcd') }}
+{% if openshift_openstack_private_dns_domain != openshift_openstack_public_dns_domain %}
```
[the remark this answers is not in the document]
tomassedovic: Also, can we just skip the if altogether? And assume that if private is not set, it will default to public?

That should make the templates simpler and not harm anything.

tomassedovic:
Also, I wonder whether setting the openshift hostname and IP options like we do here means we could do the deployment without requiring the internal DNS whatsoever.

I'll go & try to test that. If we can, that'll be fantastic news for the move to openshift-ansible, because they don't want to ship DNS there.

bogdando:
It's time to resubmit on top of https://github.com/openshift/openshift-ansible/pull/6039

roles/dns-records/tasks/main.yml
bogdando: FWIW, that is the main point for multiple names resolvable via internap view. Public names should be as well resolved, yet with priv IPs. It doesn't work another way. PTAL the test case in the commit msg, it describes the behavior that matches this code. Firstly, let's adjust the test cases if I understood the thing wrong

playbooks/provisioning/openstack/sample-inventory/group_vars/all.yml
bogdando: it has it at the line 3

playbooks/provisioning/openstack/sample-inventory/inventory.py
```diff
@@ -72,12 +72,18 @@
             'ansible_host': ssh_ip_address
         }
 
+        vars['openshift_public_hostname'] = server.name
+        if 'private_fqdn' in server.metadata:
+            vars['openshift_hostname'] = server.metadata.private_fqdn
```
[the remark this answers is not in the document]
bogdando: The deployment works for me with arbitrary hostnames. I reomved any dependencies on the host names

roles/openstack-stack/templates/heat_stack.yaml.j2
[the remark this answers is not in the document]
bogdando: I'll try to rework this, good idea

bogdando:
> because they don't want to ship DNS there

;-(

bogdando:
@tomassedovic btw, I hope we could merge it here, once I submit those fixes for `_full_ != ...`, and hope to see it either addressed in the openshift-ansible as well, or not.

bogdando:
DNS management is deprecated for the provider. And the provider had been moved into openshift-ansible w/o DNS setup bits.

bogdando:
Reimplemented as https://github.com/openshift/openshift-ansible/pull/6349

PR #851, opened 2017-11-10
Fallback inventory_dir to the env PWD var

bogdando:
#### What does this PR do?
In ansible 2.4 localhost's inventory_dir is undefined.
Fix this with the env's $PWD lookup used as a fallback.

Fix CI job by passing inventory path as an extra var.

provision and deploy (defaults all) with ansible 2.4.x. e2e shall pass.

cc: @tomassedovic PTAL

### DIFF
--- a/playbooks/provisioning/openstack/provision-openstack.yml
+++ b/playbooks/provisioning/openstack/provision-openstack.yml
@@ -14,7 +14,7 @@
         - cinder_hosted_registry_size_gb is defined
     - role: static_inventory
       when: openstack_inventory|default('static') == 'static'
-      inventory_path: "{{ openstack_inventory_path|default(inventory_dir) }}"
+      inventory_path: "{{ openstack_inventory_path|default(lookup('env','PWD') + '/inventory') }}"
       private_ssh_key: "{{ openstack_private_ssh_key|default('') }}"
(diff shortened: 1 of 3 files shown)


playbooks/provisioning/openstack/provision-openstack.yml
```diff
@@ -14,7 +14,7 @@
         - cinder_hosted_registry_size_gb is defined
     - role: static_inventory
       when: openstack_inventory|default('static') == 'static'
-      inventory_path: "{{ openstack_inventory_path|default(inventory_dir) }}"
+      inventory_path: "{{ openstack_inventory_path|default(lookup('env','PWD') + '/inventory') }}"
```
tomassedovic: /me shakes fist at Ansible.

The default will now only work when the directory is in `$PWD/inventory`. Previously, it worked for any location. But I can't figure out how to do that under the new rules so let's do this.

tomassedovic:
@bogdando what do you think about the end to end CI? Should we set it to use Ansible 2.4 in this PR or later?

Anyway, this needs a the openstack CI run before merging.

bogdando:
Let's poke versions in a follow-up @tomassedovic

PR #860, opened 2017-11-21
Fix static inventory

bogdando:
#### What does this PR do?
Do not add empty hosts into static inventory and ssh config.

Provision openstack VMs with openshift-ansible, then generate a static inventory as a separate playbook called from this repo. There should be no empty/'---' hosts entries in the static `inventory/hosts` file. `ansible -m ping all` should be working with the static SSH config as well.

cc: @tomassedovic PTAL


bogdando:
We do not maintain the static inventory in the openshift-ansible, it's fallen behind in the attic, and this repository keeps hosting it for those who wants to use it.

bogdando:
Static inventory and SSH config for the provider is not maintained any more

PR #861, opened 2017-11-24
Fix VMWare deployment when auth type is set to none

takuvata:
#### What does this PR do?
Allows to successfully deploy vmware ref arch when auth_type variable is set to 'none'. Without this fix the conditional handling the scenario falls under conditional checking if auth_type == ldap rendering it to be just a dead code.

Set variable auth_type=none in ocp_on_vmware.ini, deploy - it should work.

Don't think so


e-minguez:
@dav1x PTAL?

dav1x:
Ill need to review this. This was before I refactored the ocp-on-vmware.py.

PR #863, opened 2017-11-27
Azure add host

jhorn-redhat:
## What does this PR do?
This PR fixes the generation of add_host.sh from bastion.sh.  Without this update the add_host.sh contains variable expansion from bastion.sh instead of the supplied arguments given when calling add_host.sh.
This can be tested by deploying the bastion host and verifying the add_host.sh script doesn't contain variable expansion from bastion.sh.


cooktheryan:
@e-minguez @glennswest PTAL

e-minguez:
LGTM. @glennswest since you merged the script with the bastion.sh, do you mind to check it? Thx!

e-minguez:
@glennswest do you mind to take a look? Thx

PR #882, opened 2017-12-18
Updated dns-records role to allow for greater flexibility

oybed:
#### What does this PR do?
In preparation for discontinuing the management of the DNS server as part of these tools, this PR updates the `dns-records` role to better align with dns server zones to allow for better flexibility when utilizing external DNS server(s). This is done by using the `fqdn` and stripping the matching `zone` (which may be the `full_dns_domain` if not provided as part of the inventory). This allows the records to use a sub-domain of the zone.

Provision an OpenShift cluster and observe no errors for DNS records adding tasks. To test "new functionality", use external DNS servers, declare the `external_nsupdate_keys` in the inventory and make sure to set the `zone` as part of the public and private section, e.g.: 

```
external_nsupdate_keys:
  public:
    key_secret: '<value-here>'
    key_algorithm: '<value-here>'
    server: '<value-here>'
    key_name: '<value-here>'
    zone: '<value-here>'
  private:
    key_secret: '<value-here>'
    key_algorithm: '<value-here>'
    server: '<value-here>'
    key_name: '<value-here>'
    zone: '<value-here>'
```

N/A

cc: @tomassedovic @bogdando @etsauer

### DIFF
--- a/roles/dns-records/tasks/main.yml
+++ b/roles/dns-records/tasks/main.yml
@@ -28,10 +28,10 @@
   set_fact:
     private_records: "{{ private_records | default([]) + [ { 'type': 'A', 'hostname': '*.' + (openshift_master_default_subdomain | replace('.' + nsupdate_zone_private, '')), 'ip': hostvars[item]['private_v4'] } ] }}"
   when:
-  - openshift_master_default_subdomain is defined
-  - openshift_master_default_subdomain|trim != ''
+    - openshift_master_default_subdomain is defined
+    - openshift_master_default_subdomain|trim != ''
   with_items:
-  - "{{ groups['infra_hosts'] }}"
+    - "{{ groups['infra_hosts'] }}"
 
 - name: "Add public master cluster hostname records to the private A records (single master)"
   set_fact:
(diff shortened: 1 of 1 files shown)


bogdando:
Thank you for the fix @oybed ! Unfortunately, there is a [new location](https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_openstack/tasks/populate-dns.yml) takes priority over changes being proposed into this repo. Could you please propose a fix to the new location instead/as well?

bogdando:
CI started https://travis-ci.org/openshift/openshift-ansible-contrib/jobs/324067852

bogdando:
CI passed, so I'm ok to merge this. Just please keep in mind that we'll need a cherry-pick

oybed:
@bogdando yeah, we (CASL) haven't moved to the new `openshift-ansible` implementation yet as it seems quite a bit different as-is and may require quite some work on our side to adopt. For now we'd like to get this up to what it needs to be to support our v3.7 approach and then we'll look to move to the `openshift-ansible` approach.

PR #747, opened 2017-09-19
Allow the specification of server group policies when provisioning openstack

tzumainn:
#### What does this PR do?
Allow the specification of master and infra server group policies when provisioning openstack

#### How should this be manually tested?

Uncomment the following parameters in all.yml:

    # Specify server group policies for master and infra nodes. Nova must be configured to
    # enable these policies.
    openstack_master_server_group_policies: [affinity]
    openstack_infra_server_group_policies: [affinity]

You can specify 'affinity' or 'anti-affinity'. Run OpenStack provisioning; it should complete with no errors.

#### Is there a relevant Issue open for this?
N/A

#### Who would you like to review this?
cc: @tomassedovic @bogdando @Tlacenka PTAL

PR #886, opened 2017-12-19
Fixing compatible issues between repos

oybed:
#### What does this PR do?
Fixing issues due to PR https://github.com/openshift/openshift-ansible-contrib/pull/747 . The introductions of the `*_server_group_policies` causes issues with existing inventories within the CASL repo + should be an optional configuration parameter. 

Run with one of the CASL repo inventories

N/A

cc: @tomassedovic @tzumainn @bogdando @etsauer

### DIFF
--- a/roles/openstack-stack/templates/heat_stack.yaml.j2
+++ b/roles/openstack-stack/templates/heat_stack.yaml.j2
@@ -562,14 +562,14 @@ resources:
       - interface
 {% endif %}
 
-{% if master_server_group_policies|length > 0 %}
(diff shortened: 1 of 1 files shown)


bogdando:
Note that this is an obsolete role, the most recent one is located [here](https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_openstack/templates/heat_stack.yaml.j2) . Although it's missing some if the features, like DNS management, so LGTM as we have no choice but modify this one

bogdando:
Never mind, the new role contains defaults and the cherry-pick is not needed there

PR #905, opened 2018-01-13
Using safe to prevent ansible errors with some Ansible environments.

mbruzek:
#### What does this PR do?
The openshift-ansible repository does not include a DNS server. While trying to use the osp-dns the playbook did not work in my environment. 

Details: ansible 2.4.2, python 2.7.5 rhel 7.4. 

@markllama worked with me and we discovered the output_value from `openstack stack output show -f value -c output_value` was not safe YAML it contained things like an empty list ([]) which did not work on my environment. 
"{u'master': {u'name': u'ns-master', u'address': u'172.21.0.134'}, u'slaves': []}"

Try to create a dns server with the openstack client version 3.11.0
`ansible-playbook -vvv --key-file ~/.ssh/id_rsa -e @dns_env.yml openshift-ansible-contrib/reference-architecture/osp-dns/deploy-dns.yaml`
You will see it fail on the "set_fact" step with the error:
```
TASK [set_fact] ***********************************************************************************************************************************************                                    
task path: /home/cloud-user/openshift-ansible-contrib/reference-architecture/osp-dns/deploy-dns.yaml:44                                                                                            
Friday 12 January 2018  17:43:57 -0500 (0:00:02.339)       0:02:32.932 ********                                                                                                                    
fatal: [localhost]: FAILED! => {                                                                                                                                                                   
    "msg": "the field 'args' has an invalid value ([]), and could not be converted to an dict.The error was: Expecting property name enclosed in double quotes: line 1 column 2 (char 1)\n\nThe err
or appears to have been in '/home/cloud-user/openshift-ansible-contrib/reference-architecture/osp-dns/deploy-dns.yaml': line 44, column 5, but may\nbe elsewhere in the file depending on the exact
 syntax problem.\n\nThe offending line appears to be:\n\n\n  - set_fact:\n    ^ here\n\nexception type: <type 'exceptions.ValueError'>\nexception: Expecting property name enclosed in double quote
s: line 1 column 2 (char 1)"                                                                                                                                                                       
}  
```
After applying the fix I was able to run the command without any errors: 
`ansible-playbook -vvv -e @dns_env.yml openshift-ansible-contrib/reference-architecture/osp-dns/deploy-dns.yaml`

No I did not create an issue, I am submitting this PR as the solution.

cc: @tomassedovic PTAL


e-minguez:
Those scripts are no longer updated and are not supported. Closing this. Thanks

Issue #913, opened 2018-01-26
[VMWare] add-node.py doesn't pull in SSH key if it is absent


vponomaryov:
# BUG REPORT INFO

### Environment

- **Cloud provider**:  VMWare

- **OS details** (`printf "$(uname -srm)\n$(cat /etc/os-release)\n"`):  RHEL 7.4

- **Version of Ansible** (`ansible --version`):  2.4.2.0

- **Version of Jinja** (`pip freeze | grep -i jinja`):  Jinja2==2.7.2

- **Version of Shade** (`pip freeze | grep -i shade`):  n/a

**Command used to invoke ansible**:  
./add-node.py

# BUG DESCRIPTION
We are able to add nodes to the existing OCP cluster calling "add-node.py" python module.
But we fail to do it with SSH connection errors in case we haven't run "ocp-on-vmware.py" from the same machine/dir or just "git stashed" local changes for other needs. The reason for it is that only "ocp-on-vmware.py" pulls in SSH key required by "ansible", but not "add-node.py".

PR #914, opened 2018-01-26
[VMWare] Make sure ssh key is pulled in running add-node.py

vponomaryov:
For the moment, from two python modules, which run ansible playbooks,
only one pulls in SSH key configured in ansible.cfg
It is 'ocp-on-vmware.py', but it should also be done in 'add-node.py'.
So, add the same logic in second module to avoid SSH connection errors
running 'add-node.py' module from other dirs/machines than
'ocp-on-vmware.py' had run on.

https://github.com/openshift/openshift-ansible-contrib/issues/913

cc: @tomassedovic PTAL @dav1x @cooktheryan


dav1x:
nice work! lgtm

PR #939, opened 2018-02-15
Changes needed to deploy this osp-dns in our environment.

mbruzek:
#### What does this PR do?
I am using the osp-dns to deploy a DNS server in the OpenStack based OpenShift cluster. While working in this code, there were a few errors. The first error was dnssec validation failed, and the second error was the json object that was returned from the stack output contained special characters and the object could not be used in Ansible on the following steps. 

Please review these changes and let me know if any updates are needed.

Follow the [osp-dns README](https://github.com/mbruzek/openshift-ansible-contrib/blob/dns-fixes/reference-architecture/osp-dns/README.adoc) 

I did not create a bug for this issue, just the PR.

cc: @tomassedovic @markllama PTAL

### DIFF
--- a/reference-architecture/osp-dns/deploy-dns.yaml
+++ b/reference-architecture/osp-dns/deploy-dns.yaml
@@ -42,7 +42,8 @@
     register: stack_output_raw
 
   - set_fact:
-      stack_output: "{{ stack_output_raw.stdout|from_json }}"
+      # Need "safe" to create an Ansible safe dictionary object.
+      stack_output: "{{ stack_output_raw.stdout|safe }}"
 
   - name: Add the master to the inventory
     add_host:
--- a/reference-architecture/osp-dns/ansible/templates/named.conf.j2
+++ b/reference-architecture/osp-dns/ansible/templates/named.conf.j2
@@ -16,22 +16,22 @@ options {
 	memstatistics-file "/var/named/data/named_mem_stats.txt";
 	allow-query     { any ; };
 
-	/* 
+	/*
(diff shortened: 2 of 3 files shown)


tzumainn:
Hi! Is it possible to make these changes optional, perhaps based on a parameter set in the inventory?

mbruzek:
@tzumainn thanks for the review! I made the values controlled by optional variables as you requested. We need these changes for a successful deploy. Please take another look.

reference-architecture/osp-dns/deploy-dns.yaml
```diff
@@ -42,7 +42,8 @@
     register: stack_output_raw
 
   - set_fact:
-      stack_output: "{{ stack_output_raw.stdout|from_json }}"
+      # Need "safe" to create an Ansible safe dictionary object.
+      stack_output: "{{ stack_output_raw.stdout|safe }}"
```
tzumainn: One question: is it possible to only 'safe' the particular output piece you need, when you need it? I'm just worried that this change could have unintended consequences when other parts of the playbook read this data.

mbruzek: It may be possible to pull apart that object but as it is now it will NOT complete successfully on the latest version of ansible 2.4.2 that openshift-ansible requires. I believe this is some old code that may have worked in the past but does not work now. Please feel free to run it yourself to verify.

tzumainn: Ah, okay - looks good then!

Issue #949, opened 2018-02-26
gcp: enhancement to include the storage tiering


rgordill:
# FEATURE REQUEST INFO

Configuration of storage tier-ing using config.yaml. Currently, pd-ssd is hardcoded. 

## Problem Description

Storage tiering configuration

## Proposed Change

new parameter in config.yaml, disk_type to select the storage (pd-ssd, pd-standard, etc) to google cloud provider.

## Implementation

New PR on the go.

## Acceptance criteria

ocp-on-gcp.sh --infra should provision the right storage

## Testing

Configure config.yaml with new parameter.

Execute ocp-on-gcp.sh --infra.

PR #950, opened 2018-02-26
Property for disk type (tier) added

rgordill:
#### What does this PR do?
Add disk tiering as a parameter for playbooks

just an ocp-on-gcp.sh --infra should do the work

https://github.com/openshift/openshift-ansible-contrib/issues/949

cc: @pschiffe PTAL


rgordill:
Expired, no longer needed.

PR #954, opened 2018-02-28
Fix import script project name & remove empty files

e-minguez:
#### What does this PR do?
Fix import script project name & remove empty files

Use project_export and project_import scripts

Nope

cc: @cooktheryan PTAL

### DIFF
--- a/reference-architecture/day2ops/scripts/project_export.sh
+++ b/reference-architecture/day2ops/scripts/project_export.sh
@@ -377,7 +377,7 @@ done
 
 PROJECT=${1}
 
-mkdir ${PROJECT}
+mkdir -p ${PROJECT}
 
 ns
 rolebindings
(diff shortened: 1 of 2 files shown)


reference-architecture/day2ops/scripts/project_export.sh
```diff
@@ -407,4 +407,7 @@ deployments
 replicasets
 poddisruptionbudget
 
+echo "Removing empty files"
+find ${PROJECT} -type f -empty -delete
```
pschiffe: If project variable is provided by user ($1) it should be definitely quoted, as the path may contain white spaces. This is definitely important when deleting something. If I can recommend, use this tool to check your bash scripts - https://www.shellcheck.net/ It's even in Fedora repos ($ sudo dnf install ShellCheck ).

Issue #960, opened 2018-03-05
"Ensure that the openshift_clusterid is set ..." in add-node on aws


livsky87:
# BUG REPORT INFO

### Environment

- **Cloud provider**:  aws

- **OS details** (`printf "$(uname -srm)\n$(cat /etc/os-release)\n"`):  
Linux 3.10.0-693.5.2.el7.x86_64 x86_64
NAME="CentOS Linux"
VERSION="7 (Core)"
ID="centos"
ID_LIKE="rhel fedora"
VERSION_ID="7"
PRETTY_NAME="CentOS Linux 7 (Core)"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:centos:centos:7"
HOME_URL="https://www.centos.org/"
BUG_REPORT_URL="https://bugs.centos.org/"
CENTOS_MANTISBT_PROJECT="CentOS-7"
CENTOS_MANTISBT_PROJECT_VERSION="7"
REDHAT_SUPPORT_PRODUCT="centos"
REDHAT_SUPPORT_PRODUCT_VERSION="7"

- **Version of Ansible** (`ansible --version`):  2.3.3.0

- **Version of Jinja** (`pip freeze | grep -i jinja`):  Jinja2==2.7.2

- **Version of Shade** (`pip freeze | grep -i shade`):  none

- **The openstack-ansible-contrib version (commit)** (`git rev-parse --short HEAD`):  fa05d56

**Copy of inventory files and custom variables used (please omit your secrets!)**:  add-node.py
(in ./reference-architecture/aws-ansible/)

**Command used to invoke ansible**:  
./add-node.py (with some options)

**Output of ansible run**:  
FAILED! => {"changed": false, "failed": true, "msg": "Ensure that the openshift_clusterid is set and that all infrastructure has the required tags.\nFor dynamic provisioning when using multiple clusters in different zones, tag each node with Key=kubernetes.io/cluster/xxxx,Value=clusterid where xxxx and clusterid are unique per cluster. In versions prior to 3.6, this was Key=KubernetesCluster,Value=clusterid.\nhttps://github.com/openshift/openshift-docs/blob/master/install_config/persistent_storage/dynamically_provisioning_pvs.adoc#available-dynamically-provisioned-plug-ins\n"}

**Anything else do we need to know**:

livsky87:
[...]

PR #961, opened 2018-03-05
Resolve add-node error on aws(#960)

livsky87:
Error "Ensure clusterid is set along with the cloudprovider" occur
in add-node playbook runtime.
It is because node-setup.yaml is called without "openshift_clusterid"
variable setting.
With openshift_clusterid variable, we can avoid this issue.

#### What does this PR do?
About An issue (https://github.com/openshift/openshift-ansible-contrib/issues/960)

I used python script ./add-node.py (with some options)

https://github.com/openshift/openshift-ansible-contrib/issues/960

cc:  @dav1x


livsky87:
Hi, @dav1x :)
I'm very happy to contribute openshift-ansible-contrib.
Would you review this?

PR #972, opened 2018-03-15
Check for the existence of atomic files before copy.

bobross419:
When these directories aren't present an error is generated:
 ```
   Exporting OCP related files to ./backup
    cp: cannot stat ‘/etc/sysconfig/atomic-*’: No such file or directory
 
```

Added logic to check for the existence of /etc/sysconfig/atomic-* before invoking the cp.  I only manually tested on an existing server that does not have the /etc/sysconfig/atomic- files.  I don't have access an atomic based install.


e-minguez:
Thanks for finding the issue! I've created a different PR [here](https://github.com/openshift/openshift-ansible-contrib/pull/973) including origin support. Let me know if that fixes it so we can merge that PR.
Thanks again!

e-minguez:
Also, I've created a new PR to create the backup using an ansible playbook instead running a script, see here https://github.com/openshift/openshift-ansible-contrib/pull/974

bobross419:
Thanks @e-minguez !  973 solves for me and I'm looking forward to 974. Closing this PR.

PR #973, opened 2018-03-15
Properly backup sysconfig files

e-minguez:
#### What does this PR do?
Copy OCP and Origin files in /etc/sysconfig if found

Run the script

Related to https://github.com/openshift/openshift-ansible-contrib/pull/972

cc: @bobross419 @cooktheryan  PTAL


bobross419:
I tested this manually on the same server as I used in #972, diff of the output directories match.  LGTM! Thanks @e-minguez 

```
# oc version
oc v1.5.0+031cbe4
kubernetes v1.5.2+43a9be4
```

PR #1006, opened 2018-05-10
allow unlimited number of app_nodes

mmckinst:
#### What does this PR do?

Right now there's a hardcoded limit of 3 app nodes. Openshift can support more
than three without issue. This PR will allow that.

I have changed the logic so its more of a sanity check that someone is deploying
with at least one app node.

Deploy on AWS with 6 app nodes.

No.

cc: @tomassedovic PTAL


cooktheryan:
@mmckinst one issue that I ran into was that the cloudformation json would become too large to actually deploy

mmckinst:
Now that you mention that I remember running in to the same issue. I think the workaround was to just add more app nodes after the initial install.

PR #1011, opened 2018-05-15
update required repos to list 3.9 instead of 3.6

mmckinst:
#### What does this PR do?

Updates the required repos list to 3.9 instead of 3.6

NA

No

cc: @cooktheryan PTAL

### DIFF
--- a/roles/rhsm-repos/defaults/main.yaml
+++ b/roles/rhsm-repos/defaults/main.yaml
@@ -1,2 +1,8 @@
 ---
-openshift_required_repos: ['rhel-7-server-rpms', 'rhel-7-server-extras-rpms', 'rhel-7-server-ose-3.6-rpms', 'rhel-7-fast-datapath-rpms']
+# repos from https://docs.openshift.com/container-platform/3.9/install_config/install/host_preparation.html#host-registration
+openshift_required_repos:
+  - 'rhel-7-server-rpms'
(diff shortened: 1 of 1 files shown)


roles/rhsm-repos/defaults/main.yaml
```diff
@@ -1,2 +1,2 @@
 ---
-openshift_required_repos: ['rhel-7-server-rpms', 'rhel-7-server-extras-rpms', 'rhel-7-server-ose-3.6-rpms', 'rhel-7-fast-datapath-rpms']
+openshift_required_repos: ['rhel-7-server-rpms', 'rhel-7-server-extras-rpms', 'rhel-7-server-ose-3.9-rpms', 'rhel-7-fast-datapath-rpms']
```
kanekoh: I wonder that new required repository 'rhel-7-server-ansible-2.4-rpms' is needed.

mmckinst: On OCP 3.9 ansible comes from the `rhel-7-server-extra-rpms` repo.

mmckinst: Interesting. ansible 2.4 is also in the `rhel-7-server-extra-rpms` repo so adding `rhel-7-server-ansible-2.4-rpms` seems redundant. Nevertheless I'll update my PR to include that repo per the docs.

Issue #1043, opened 2018-06-05
Export project failed


philipho11:
Environment: openshift 3.9
Failed at getting secrets

https://github.com/openshift/openshift-ansible-contrib/blob/bfbb2fc2de4ee932e00609c70ea982c82ae6e6d4/reference-architecture/day2ops/scripts/project_export.sh#L74

###########
# WARNING #
###########
This script is distributed WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND
Beware ImageStreams objects are not importables due to the way they work
See https://github.com/openshift/openshift-ansible-contrib/issues/967
for more information
Exporting namespace to wp-demo/ns.json
Exporting rolebindings to wp-demo/rolebindings.json
Exporting serviceaccounts to wp-demo/serviceaccounts.json
Exporting secrets to wp-demo/secrets.json
error: syntax error, unexpected QQSTRING_START, expecting IDENT

        .metadata.annotations."kubernetes.io/service-account.uid"
                               ^
1 compile error

Issue #1039, opened 2018-06-01
Error from server (Forbidden): User "admin" cannot list egressnetworkpolicies.network.openshift.io in the namespace "sample-py": User "admin" cannot list egressnetworkpolicies.network.openshift.io in project "sample-py" (get egressnetworkpolicies.network.openshift.io)


brikeyes:
bkeyes@AEM-9Q8N662 MINGW64 ~/exports/friday/contrib/openshift-ansible-contrib/reference-architecture/day2ops/scripts (master)
$ ./project_export.sh sample-py
###########
# WARNING #
###########
This script is distributed WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND
Beware ImageStreams objects are not importables due to the way they work
See https://github.com/openshift/openshift-ansible-contrib/issues/967
for more information
Exporting namespace to sample-py/ns.json
Exporting rolebindings to sample-py/rolebindings.json
Exporting serviceaccounts to sample-py/serviceaccounts.json
Exporting secrets to sample-py/secrets.json
Exporting deploymentconfigs to sample-py/dc_*.json
Patching DC...
Patching DC...
Exporting buildconfigs to sample-py/bcs.json
Exporting builds to sample-py/builds.json
Exporting imagestreams to sample-py/iss.json
Exporting imagestreamtags to sample-py/imagestreamtags.json
Exporting replicationcontrollers to sample-py/rcs.json
Exporting services to sample-py/svc_*.json
Exporting pods to sample-py/pods.json
Exporting podpreset to sample-py/podpreset.json
Exporting configmaps to sample-py/cms.json
Exporting egressnetworkpolicies to sample-py/egressnetworkpolicies.json
Error from server (Forbidden): User "admin" cannot list egressnetworkpolicies.network.openshift.io in the namespace "sample-py": User "admin" cannot list egressnetworkpolicies.network.openshift.io in project "sample-py" (get egressnetworkpolicies.network.openshift.io)

Is this a **BUG REPORT** or a **FEATURE REQUEST** ? (choose one):

# BUG REPORT INFO

### Environment

- **Cloud provider**:  

- **OS details** (`printf "$(uname -srm)\n$(cat /etc/os-release)\n"`):  

- **Version of Ansible** (`ansible --version`):  

- **Version of Jinja** (`pip freeze | grep -i jinja`):  

- **Version of Shade** (`pip freeze | grep -i shade`):  

[...]

PR #1068, opened 2018-08-20
Refactored export script

gerald1248:
* lists containing multiple objects are valid (not stacked) JSON
* like the original, they can be restored via `oc create -f`
* backup continues when a resource type is not recognised
* backup continues when permission to access an object is refused (see issues #1039, #1043)
* deletion of empty files is no longer necessary
* the two exceptions (dc and svc), which create a separate file for each object, are unchanged

#### What does this PR do?
The current project_export.sh produces JSON files that can't be parsed by all processors and so may require work before they can be restored. (`jq` is fine but `python -m json.tool` complains about "Extra data".) This version produces List objects instead.

The commit also ensures the export handles resource types not known to the master (the user may be running an older OpenShift version) and types which the user/service account can't access (e.g. Secrets) gracefully. With this change, the script supports older OpenShift versions and non-admin users.

Log in to any OpenShift cluster and create a backup of an existing project (e.g. myproject):

```
$ ./project_export.sh myproject
```

https://github.com/openshift/openshift-ansible-contrib/issues/1039
https://github.com/openshift/openshift-ansible-contrib/issues/1043
https://github.com/openshift/openshift-ansible-contrib/issues/1072

cc: @e-minguez

### DIFF
--- a/reference-architecture/day2ops/scripts/project_export.sh
+++ b/reference-architecture/day2ops/scripts/project_export.sh
@@ -43,7 +43,7 @@ exportlist(){
 
   echo "Exporting '${KIND}' resources to ${PROJECT}/${BASENAME}.json"
 
-  BUFFER=`oc get ${KIND} --export -o json -n ${PROJECT} || true`
+  BUFFER=$(oc get ${KIND} --export -o json -n ${PROJECT} || true)
 
   # return if resource type unknown or access denied
   if [ -z "${BUFFER}" ]; then
@@ -52,7 +52,7 @@ exportlist(){
   fi
 
   # return if list empty
-  if [ `echo ${BUFFER} | jq '.items | length > 0'` == "false" ]; then
+  if [ $(echo ${BUFFER} | jq '.items | length > 0') == "false" ]; then
     echo "Skipped: list empty"
     return
   fi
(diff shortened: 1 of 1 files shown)


reference-architecture/day2ops/scripts/project_export.sh
```diff
@@ -31,48 +31,60 @@ usage(){
...
+    return
+  fi
+
+  KIND=$1
+  BASENAME=$2
+  DELETEPARAM=$3
+
+  echo "Exporting '${KIND}' resources to ${PROJECT}/${BASENAME}.json"
+
+  BUFFER=`oc get ${KIND} --export -o json -n ${PROJECT} || true`
```
e-minguez: Do you mind to use $() instead ``? mywiki.wooledge.org/BashFAQ/082

reference-architecture/day2ops/scripts/project_export.sh
```diff
@@ -31,48 +31,60 @@ usage(){
...
+
+  # return if resource type unknown or access denied
+  if [ -z "${BUFFER}" ]; then
+    echo "Skipped: no data"
+    return
+  fi
+
+  # return if list empty
+  if [ `echo ${BUFFER} | jq '.items | length > 0'` == "false" ]; then
```
e-minguez: Same here

reference-architecture/day2ops/scripts/project_export.sh
```diff
@@ -31,48 +31,60 @@ usage(){
...
+  fi
+
+  echo ${BUFFER} | jq ${DELETEPARAM} > ${PROJECT}/${BASENAME}.json
 }
 
+ns(){
+  exportlist \
+    ns \
+    ns \
+    'del(.items[].status,.items[].metadata.uid,.items[].metadata.selfLink,.items[].metadata.resourceVersion,.items[].metadata.creationTimestamp,.items[].metadata.generation)'
```
e-minguez: Ideally this list should be multiple lines to avoid scrolling

reference-architecture/day2ops/scripts/project_export.sh
```diff
@@ -31,48 +31,60 @@ usage(){
...
-  oc get --export -o=json rolebindings -n ${PROJECT} | jq '.items[] |
-  del(.metadata.uid,
-      .metadata.selfLink,
-      .metadata.resourceVersion,
-      .metadata.creationTimestamp
-      )' > ${PROJECT}/rolebindings.json
+  exportlist \
+    rolebindings \
+    rolebindings \
+    'del(.items[].metadata.uid,.items[].metadata.selfLink,.items[].metadata.resourceVersion,.items[].metadata.creationTimestamp)'
```
e-minguez: Same for this list

reference-architecture/day2ops/scripts/project_export.sh
```diff
@@ -31,48 +31,60 @@ usage(){
...
-  oc get --export -o=json serviceaccounts -n ${PROJECT} | jq '.items[] |
-    del(.metadata.uid,
-        .metadata.selfLink,
-        .metadata.resourceVersion,
-        .metadata.creationTimestamp
-        )' > ${PROJECT}/serviceaccounts.json
+  exportlist \
+    serviceaccounts \
+    serviceaccounts \
+    'del(.items[].metadata.uid,.items[].metadata.selfLink,.items[].metadata.resourceVersion,.items[].metadata.creationTimestamp)'
```
e-minguez: And this (and any other list to avoid commenting too much)

reference-architecture/day2ops/scripts/project_export.sh
gerald1248: Sorry, fixed in 02cfd39.

reference-architecture/day2ops/scripts/project_export.sh
gerald1248: Sorry, fixed in 02cfd39.

reference-architecture/day2ops/scripts/project_export.sh
gerald1248: Fixed in 02cfd39.

reference-architecture/day2ops/scripts/project_export.sh
gerald1248: Fixed in 02cfd39.

reference-architecture/day2ops/scripts/project_export.sh
gerald1248: Fixed in 02cfd39.

gerald1248:
@e-minguez I've fixed the bug you discovered. One of the jq expressions was missing the leading ".spec".

e-minguez:
I've tested and it seems to be working fine. Thanks!

